PR #1, opened 2015-09-29
Chores/code quality support #104349670

chrisdolendo:


kaptron:
:+1: Great!

Gemfile
```diff
@@ -13,6 +12,8 @@ gem 'uglifier', '>= 1.3.0'
 gem 'coffee-rails', '~> 4.0.0'
 # See https://github.com/rails/execjs#readme for more supported runtimes
 # gem 'therubyracer',  platforms: :ruby
+gem 'rubocop', require: false
```
dstavis: What does 'require: false' do?

chrisdolendo: It means install the gem, but do not call require when you start Bundler

chrisdolendo:
@dstavis @menslow @kaptron just need one more thumbs up from any of you :p

menslow:
Does is make sense to put both of these in a "development" group?

gem 'rubocop', require: false
gem 'rails_best_practices'

menslow:
As well as overcommit.

Gemfile
```diff
@@ -1,10 +1,9 @@
 source 'https://rubygems.org'
 
-
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '4.1.10'
-# Use sqlite3 as the database for Active Record
-gem 'sqlite3'
+ruby '2.2.2'
```
kaptron: just to be a stickler, latest ruby is 2.2.3 and rails is 4.2.4, might as well use the latest as a starting point if we are going to specify version.

PR #4, opened 2015-10-06
Chores/jsc preset #104770240

CindyConway:
Nothing fancy, just changed the .jscsrc file so the preset is airbnb instead of google.


exygy-dev:
:+1:

chrisdolendo:
:+1:

PR #5, opened 2015-10-06
added overcommit intsall command to README

CindyConway:
I had to run `overcommit --install` so that the git hooks would run as expected, so I added that to the README.

I also modified the list of dependancies to make them show up in a list.


chrisdolendo:
:+1: right on!

PR #6, opened 2015-10-07
[#1405352] - Added custom pre-commit script to prevent committing directly to protected branches (master/demo/staging/production)

kaptron:


kaptron:
So the "are you sure? y/n" pattern didn't really work with overcommit because the message isn't displayed until _after_, so you never actually see the question until it's too late. Because of that I just made it so you aren't allowed to commit to any of the protected branches at all, and it just gives you an error if you try. If you ever did want to commit directly for whatever reason, it seems fair that you'd have to go out of your way to do so.

chrisdolendo:
:+1: right on!

PR #16, opened 2015-11-05
Chore/bring icons from pt #107451858

menslow:
Hi Folks,
I migrated over the existing SVG icons from the pattern library into the app. Basically, the collection just lives at the bottom of the page and can be called anywhere in the app by referencing the id. Example:

```
svg
  use xlink:href="#i-like"
```


chrisdolendo:
:+1:

PR #17, opened 2015-11-05
Feature/single listing page part 1 #106968870

CindyConway:
Created single listing page. Nav bar was coded in the application layout file, moved it to an Angular template.

We added a link to MyFavorites in the nav bar. It currently goes to non existent route: favorites


kaptron:
:+1: 
PR looks good to me! @chrisdolendo and I were working on adding some tests to the ListingController and some refactoring which will apply to this feature as well, so I think it's ok if this PR doesn't include those tests yet.

chrisdolendo:
:+1: sick job @menslow and @CindyConway !!!!

PR #18, opened 2015-11-06
Updated icons and styles from Pattern Library

menslow:
Simple PR to bring some new CSS and Icons into the main app.


chrisdolendo:
@menslow looks like another feature was tied in with this -- the favorites page. so i would remove that commit if possible :dancer:

PR #19, opened 2015-11-07
Updated SVG icons from Pattern Library, and refreshed site CSS

menslow:
(Corrected from original pull request):

Pulled over additional SVG icons and CSS styles from pattern library:

#i-like-fill
#i-map


kaptron:
:+1: Looks good.

PR #21, opened 2015-11-10
Feature/my favorites page #106899442

menslow:
My favorites page is as complete as it can be until we are able to query the Salesforce API.

All the hooks are in place for us to pass an object or array (depending on the format Salesforce wants) to Salesforce via the getListingsByIds() function.

This function copies the listings over to Service.Listings and we display the listing on the favorites page the same way we do on the main listings page.

NOTE: You will see all listings on the favorites page, because we don't have that Salesforce result of your favorites yet. After the API endpoint is available, you will only see your favorites in this view.


exygy-dev:
Right on! Looks awesome. :+1: 

My only question is if there should be a title on the page at the top to indicate what page I'm on like:
`h1 ' My Favorites`

but I guess that's not our problem to worry about since it's not part of any wireframes or anything.

exygy-dev:
Oh forgot to mention, don't forget to include story numbers in your commits so they show up on the pivotal story!

PR #28, opened 2015-11-13
Chore/Refactor Navbar - [#107960374]

kaptron:
Refactored navbar to not have to be included on every page, moved its logic to its own controller (out of SharedService). 

However now all the route names are prefaced with "dahlia." like "dahlia.welcome". We could use "app" there or any default name, but "dahlia" seemed in line with how we're prefixing other angular components. See here (towards the bottom of the page) for the pattern I used to set up the header -- @menslow would be good to get your take on it: http://www.funnyant.com/angularjs-ui-router/


menslow:
I likey!

PR #33, opened 2015-11-19
[#107533338] - Eligibility filters get applied to browse listings page

kaptron:
Currently with fake data; also added the ability to go back and edit your filters on the form, which are saved with `$cookies` the same way we store our favorites.


menslow:
Looks great Dave!

chrisdolendo:
looks good!

chrisdolendo:
looks like this needs re-basing since we moved away from cookies to ngStorage in another PR (auto populate income in elig estimator)

PR #38, opened 2015-11-30
Feature/initial salesforce sync #106898740

kaptron:


spec/requests/api/v1/listings_spec.rb
```diff
@@ -0,0 +1,33 @@
+# spec/requests/api/v1/listings_spec.rb
+require 'spec_helper'
+require 'support/vcr_setup'
+
+describe 'Listings API' do
```
chrisdolendo: yay for tests!

spec/support/vcr_setup.rb
```diff
@@ -0,0 +1,20 @@
+VCR.configure do |config|
```
chrisdolendo: So dope that you used VCR!

spec/spec_helper.rb
```diff
@@ -21,6 +21,9 @@
 require 'codeclimate-test-reporter'
 CodeClimate::TestReporter.start
 
+require 'webmock/rspec'
+WebMock.disable_net_connect!(allow_localhost: true)
```
chrisdolendo: @kaptron Interesting - What networks were interfering with the tests to need only localhost to connect?

chrisdolendo:
:+1:

spec/spec_helper.rb
kaptron: @chrisdolendo I got that line from Webmock instructions, it is sort of confusing how Webmock + VCR work together but my understanding is that, if you wrap a request w/ VCR then it knows "ok I'm allowed to go outside the network if a cassette needs to be recorded" otherwise that line is basically saying "throw an exception if you try to connect outside the network, but allow localhost" e.g. if one of your tests is hitting `localhost:3000/xyz` or something, then that is allowed.

PR #40, opened 2015-12-02
Features/populate income in eligibility #108832666

kaptron:


app/assets/javascripts/pages/templates/eligibility-estimator.html.slim
```diff
@@ -37,9 +37,14 @@ section.row
         .medium-12.columns.bg-mist.padding-top.padding-bottom
           h3.t-sans.gamma
             | Next add up your income from wages, benefits and other sources.
-          label.p-relative ng-class="{ error: inputInvalid('income_total') || inputInvalid('income_timeframe') }"
-            span.eligibility-form_header Household Income
-            input name="income_total" ng-model="filters.income_total" placeholder="Total all of your income sources" type="text" required=""
+          .row.collapse
+            label.p-relative ng-class="{ error: inputInvalid('income_total') || inputInvalid('income_timeframe') }"
+              span.eligibility-form_header Household Income
+            .columns ng-class="{ 'small-9': hasCalculatedIncome(), 'small-12': !hasCalculatedIncome() }"
```
chrisdolendo: Dope use of `ng-class`!

spec/javascripts/services/listing_service_spec.coffee
```diff
@@ -7,16 +7,16 @@ do ->
     httpBackend = undefined
     fakeListings = getJSONFixture('/listings.json')
     fakeListing = getJSONFixture('/listings/0.json')
-    $cookies = undefined
+    $localStorage = undefined
```
chrisdolendo: @kaptron cool, is there a reason why you favored $localStorage vs $cookies?

app/assets/javascripts/pages/EligibilityEstimatorController.js.coffee
```diff
@@ -2,20 +2,24 @@
...
-  formDefaults =
-    'household_size': ''
-    'income_timeframe': ''
-    'income_total': ''
+EligibilityEstimatorController = ($scope, $state, ListingService, IncomeCalculatorService) ->
+  $scope.filter_defaults = ListingService.eligibility_filter_defaults
+  $scope.filters = angular.copy(ListingService.eligibility_filters)
+
+  # check if we've used the IncomeCalculator to calculate income
+  if IncomeCalculatorService.calculateTotalYearlyIncome() > 0
```
chrisdolendo: @kaptron this looks interesting, I'm usually used to seeing everything encompassed under $scope, how does this work in terms of "refreshing"? Is it always constantly called since it starts with a condition?

chrisdolendo:
Pretty dope. Just got a few questions but this is :+1:

spec/javascripts/services/listing_service_spec.coffee
kaptron: @chrisdolendo Good Q, yeah it is super clean, you don't need any getters and setters ("No Getter 'n' Setter Bullshit" as it says on [their github](https://github.com/gsklee/ngStorage)) which makes the 2-way binding super duper easy, and since that's what I used on income calculator I figured it is cleaner to refactor to use the same "locally stored data" solution for favorites too rather than sometimes $localStorage, sometimes $cookies.

app/assets/javascripts/pages/EligibilityEstimatorController.js.coffee
kaptron: @chrisdolendo any lines there even setting `$scope.foo = bar` gets evaluated when the controller first runs, which in our case using the ui-router means when you reach the page that uses that controller. You could think of it like an initializer. So the `if` statement here is saying, "when we get to the eligibility estimator page, check if this condition is true, if so initialize the values on the page for income_total and income_timeframe"

kaptron:
thanks for the good comments and questions @chrisdolendo!

spec/javascripts/services/listing_service_spec.coffee
chrisdolendo: :dancer:

app/assets/javascripts/pages/EligibilityEstimatorController.js.coffee
chrisdolendo: right on!

PR #42, opened 2015-12-03
[#108885514] - Implemented Salesforce speed improvements

kaptron:
These were investigated and worked on last week but I seemed to have not made a PR for them yet. Here it is.


kaptron:
I added `rack-mini-profiler` to debug the slowness which is a pretty handy tool, but it does add a little debug info in your browser window (only on dev of course) which may be annoying to some, however you can always turn it off for the rest of your browser session with: `http://localhost:3000/?pp=disable`

PR #50, opened 2015-12-10
Features/share listing by copy to clipboard #109326066

kaptron:


kaptron:
This should work and be ready for approval -- note that the "input" field for the link is readonly, and there's a click action on it when you click on it to select the link text.

kaptron:
realizing this doesn't seem to work in iOS, and in doing research... it may be hard to do "copy to clipboard" consistently across all browsers, particularly mobile. Some solutions are Flash-only (which obviously doesn't work in iOS anyway), will have to look into further...

PR #55, opened 2015-12-12
Features/salesforce eligibility match #107537070

kaptron:


kaptron:
**bug:** @dstavis noticed that when you clear your filters it still seems to give you the match/not-match status

PR #61, opened 2015-12-14
[#110105662] Removed "Resources" link from mobile nav

menslow:
"Resources" have been removed from the mobile nav.


chrisdolendo:
:+1:

PR #67, opened 2015-12-15
[#110108790] Added scroll to top function on NavController.

menslow:
Added utility function to NavController so that whenever the view state changes, the document will scroll to the top of the screen.

Note: I tried just using autoscroll on our ui-views, but that would only scroll to the top of the view itself.


chrisdolendo:
:+1:

PR #68, opened 2015-12-15
[#110105886] Content updates to lottery module and unit features

menslow:
Updated content on Lottery Module and Unit Features.


chrisdolendo:
:+1:

PR #73, opened 2015-12-16
Feature: Listings are now grouped into 3 buckets - #109856972

kaptron:
**still to-do:** add descriptors on the listings themselves for the different states
-- also: move the sorting of "Does Match" into the JS


kaptron:
added the descriptors, also decided that refactoring of sorting "Does Match" is also be captured in a later Pivotal story for bucketing Matches / Not Matches in the browse grid.

PR #80, opened 2015-12-19
Features/Angular Pretty URLs

kaptron:
Was something Jay Nath requested, and it seemed easy enough to give it a shot. (only took 5 mins)

Played with setting up pretty URLs in angular ui-router, seems to work fine. Just had to add a catch-all case in Rails router to get it to play nice.

### DIFF
--- a/app/assets/javascripts/AngularConfig.js.coffee
+++ b/app/assets/javascripts/AngularConfig.js.coffee
@@ -28,7 +28,7 @@ angular.module('dahlia.controllers',[])
 
 # This routing directive tells Angular about the default route for our  The term "otherwise" here
 # might seem somewhat awkward, but it will make more sense as we add more routes to our application
-@dahlia.config ['$stateProvider', '$urlRouterProvider', ($stateProvider, $urlRouterProvider) ->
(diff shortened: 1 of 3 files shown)


kaptron:
TBD: test on Internet Explorer?  it has a fallback to be turned off in IE (v.9 or less) but we haven't actually tested it yet.

app/assets/javascripts/AngularConfig.js.coffee
```diff
@@ -135,6 +135,10 @@ angular.module('dahlia.controllers',[])
           templateUrl: 'income-calculator/templates/pages/summary.html'
     })
   $urlRouterProvider.otherwise('/') # default to welcome screen
+
+  # have to check if browser supports html5mode (http://stackoverflow.com/a/22771095)
+  if !!(window.history && history.pushState)
```
chrisdolendo: @kaptron what is the significance of window.history and history.pushState?

chrisdolendo:
:+1: looks good just had one quick question

chrisdolendo:
This was accepted by client--- merging in.

PR #93, opened 2016-01-13
[#111572018] Added ng-currency directive in order to format Income fi…

menslow:
* Added ng-currency directive to handle currency in the input field better.
* Currency formatting updated on blur
* Correct number gets passed to the listings results.


kaptron:
Looks great. :+1:

PR #97, opened 2016-01-27
Features/buckets for matched #110288060

menslow:
* Added Matched and Not Matched Buckets in Views Listings 
* Labels for Matched/Not-Matched only appear on Open Listings
* Matched/Not Matched buckets only appear when Eligibility filters are active


kaptron:
:+1: looks good to me, I also rebased from master (the bugfix from this morning) and pushed that to this branch. 

as a sidenote, it seems like the match data on sandbox (where we have fake open listings) is a little odd, like one has a min income that is greater than its max income, maybe we can get them to give us better test data for that. In any case I did get it to match and show me the matched bucket and that looks good.

menslow:
Sweet. Thanks for checking it out Dave. I'll wait until the story gets
accepted to merge.

PR #108, opened 2016-01-29
[#111989747] Updated Additional Rules section to include live Salesfo…

menslow:
Now pulling the following fields from Salesforce:

* listing.Credit_Rating
* listing.Eviction_History
* listing.Criminal_History

Notes:

* When field is not present, accordion module does not display on page.
* When all three of these fields are not present, the entire Additional Rules section does not display on the page.


dstavis:
That's one good PR right thar.

PR #117, opened 2016-02-04
[112924435] - Updated more resources link to additional resources pag…

chrisdolendo:
…e on listings.html

I didn't write a test on this because it was a template change :p I didn't write an additional test to an unrelated feature coz I wrote a bunch for eligibility estimator already :p i'll write more this week!


menslow:
This looks good!

PR #119, opened 2016-02-04
[#11309790] - Removes favorited listing if it no longer exists in the backend

kaptron:
And added a test!


kaptron:
accidentally combined two different bugfix tickets into this same branch, oh well.

menslow:
This looks good Dave! I verified functionality on the review app.

PR #122, opened 2016-02-09
[#113043431] Added unit test for modalInstanceController

menslow:
Added unit test for ModalInstanceController!


kaptron:
@menslow looks good to me.

PR #127, opened 2016-02-10
Features/display line breaks #111980253

chrisdolendo:


kaptron:
@chrisdolendo looks good to me since I already checked this out with you, however one thing I noticed is the inclusion of `ngSanitize` but I don't think that's being used anywhere? Perhaps you had started to use it but then removed it from the code? (or else I'm missing where it's used)

chrisdolendo:
@kaptron good question -- I used `$sanitize` in the `ListingController` and it's by default used by `ngBindHtml`. See: https://docs.angularjs.org/api/ng/directive/ngBindHtml

kaptron:
@chrisdolendo gotcha! Yeah I didn't see any reference to `$sanitize` is what I meant, however I see this from the docs that you pointed out:
```To utilize this functionality, ensure that $sanitize is available, for example, by including ngSanitize in your module's dependencies (not in core Angular).```

Looks good to me. :+1:

PR #128, opened 2016-02-10
[#112103451] Converted menu link to button instead of anchor tag

menslow:
Ready to merge.

Menu has been converted to <button> and added CSS rules to style menus inline.

Income calculator has been converted to <anchor> because of state change, which makes it more voiceover friendly.


chrisdolendo:
looks good to me :+1:

PR #135, opened 2016-02-16
Chores/fix codeclimate rubocop #113866521

kaptron:
...if we create the PR and then CodeClimate is happy, then all is well.


kaptron:
looks like that worked!

PR #139, opened 2016-02-19
Features/db backed preferences #113479989

kaptron:
**DO NOT MERGE** until salesforce `ListingsDetail` endpoint has been pushed to prod

### DIFF
--- a/app/assets/javascripts/listings/ListingService.js.coffee
+++ b/app/assets/javascripts/listings/ListingService.js.coffee
@@ -14,6 +14,7 @@ ListingService = ($http, $localStorage, $modal) ->
   # these get loaded after the listing is loaded
   Service.AMI = []
   Service.maxIncomeLevels = []
+  Service.lotteryPreferences = []
 
   $localStorage.favorites ?= []
   Service.favorites = $localStorage.favorites
@@ -214,6 +215,15 @@ ListingService = ($http, $localStorage, $modal) ->
       if data && data.ami
         angular.copy(data.ami, Service.AMI)
         angular.copy(Service.maxIncomeLevelsFor(Service.listing, Service.AMI), Service.maxIncomeLevels)
+    ).error( (data, status, headers, config) ->
+      # console.log data
+    )
+
+  Service.getLotteryPreferences = ->
+    angular.copy([], Service.lotteryPreferences)
+    $http.get('/api/v1/lottery-preferences.json').success((data, status, headers, config) ->
+      if data && data.lottery_preferences
+        angular.copy(data.lottery_preferences, Service.lotteryPreferences)
     ).error( (data, status, headers, config) ->
       # console.log data
     )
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -218,6 +219,26 @@ ListingService = ($http, $localStorage, $modal) ->
...
+    ).error( (data, status, headers, config) ->
+      # console.log data
+    )
+
+  Service.getListingUnits = ->
+    $http.get("/api/v1/listings/#{Service.listing.Id}/units").success((data, status, headers, config) ->
+      if data && data.units
+        console.log(data.units, Service.listing.Units)
+        angular.copy(data.units, Service.listing.Units)
+        # angular.copy(data.ami, Service.AMI)
```
chrisdolendo: Can we remove the commented out code and console.logs to keep it clean?

spec/requests/api/v1/listings_spec.rb
```diff
@@ -52,4 +52,34 @@
     # check to make sure the Eligibility_Match param is present
     expect(json['listings'].first['Does_Match']).not_to be_nil
   end
+
+  it 'gets AMI results' do
```
chrisdolendo: AWESOME SAUCE!

chrisdolendo: AWESOME SAUCE!

spec/javascripts/services/listing_service_spec.coffee
```diff
@@ -195,5 +197,29 @@ do ->
         return
       return
 
+    describe 'Service.getLotteryPreferences', ->
+      afterEach ->
+        httpBackend.verifyNoOutstandingExpectation()
+        httpBackend.verifyNoOutstandingRequest()
+        return
+      it 'assigns Service.lotteryPreferences with the results', ->
+        stubAngularAjaxRequest httpBackend, requestURL, fakeLotteryPreferences
```
chrisdolendo: :+1:

kaptron:
@chrisdolendo good call. We have those `# console.log data` in pretty much all of our error handlers, I can clean those up. And eventually maybe we should actually do something on the UX to handle data errors? :wink:

public/json/listings/units.json
```diff
@@ -0,0 +1,1186 @@
+{
```
chrisdolendo: @kaptron did you create this manually? i'm wondering if we can use `save_jasmine_fixtures` in the future so that the fixtures are always automatically regenerated and saved and we don't have to manually make sure to update this in case there are changes to the fields etc. that would be dope.

chrisdolendo:
@kaptron that's a really great point about UX error handling! @dstavis is this something we can consider for the next iteration

public/json/listings/units.json
kaptron: Yeah, I think all of our fake JSON have been created manually (unless we've been using `save_jasmine_fixtures` already)? But I agree we should do that, would be great!

PR #156, opened 2016-03-01
[#114738967] - added sitemap.xml

CindyConway:
Just added a sitemap.xml file temporarily until an automated solution can be created. (see my slack)


kaptron:
:+1: looks good @CindyConway and agreed, we can table the automated version of this for a later date. Having something initial that works is much more important, especially since these URLs should be fixed for a while. Will be interesting to think about how we can figure out generating Sitemap URLs for the listings, based on whatever Salesforce listing IDs are available... (other than manually)

PR #172, opened 2016-03-19
Features/#115005539 design browse

slowbot:
Updating styles for browse listing page and various states

Includes:

Status headers
Pricing Cards
Status flags
Styles for global content pages
Single listing icon headers


app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -47,11 +47,11 @@ ListingController = ($scope, $state, $sce, $sanitize, $filter, Carousel, SharedS
   $scope.isFavorited = (listing_id) ->
     ListingService.isFavorited(listing_id)
 
-  $scope.formattedAddress = (listing) ->
-    ListingService.formattedAddress(listing)
+  $scope.formattedAddress = (listing, display) ->
```
chrisdolendo: I feel like we should turn the address formatting into a filter one day. it works now but I think a filter is more applicable with the way we present the address

chrisdolendo:
looks good to me! just made one note but nothing you need to take action on right now. :+1:

PR #183, opened 2016-05-06
[#118971335] - Added initial page of Review section

kaptron:
Added initial page of Review section to hold the "How did you hear about this listing?" question.

Added a directive for the repeated checkbox form item.


menslow:
@kaptron PR looks good. I just made one adjustment that Ro requested.

PR #185, opened 2016-05-09
Features/move optional questions #118943089

kaptron:
removed b5-optional-info from You section and moved optional gender/ethnicity/race questions to the first page of review, above the referral question

(note: this feature relied on some of the previous unmerged ones, so this branch is merged with those branches as well)


menslow:
@kaptron One note... I think the dropdown for Ethnicity and Race should start with "Select One"

PR #187, opened 2016-05-12
Features/remove page counts #118970091

menslow:
…rm application

* Removed the page count directive in short form as it's no longer being used.


kaptron:
@menslow looks good -- I also removed a couple of functions from `ShortFormNavigationService` that went along with the directive

PR #188, opened 2016-05-12
Features/internationalization first page #118610977

kaptron:
Initial support for internationalization, which applies to the first page of short form. 

Some notes and decisions made, to be further discussed: 
- Language setting is set at the root level of the Angular UI-router, meaning that the initial `/` accepts a `lang` parameter e.g. `/es` or `/es/listings/...` and defaults to `en` so that if you don't have any lang param, e.g. `/listings` will behave the same as `/en/listings. 
- Language setting does not store in cookies/local so that if you come back to the site it doesn't remember your setting (I believe figuring out how this should work was TBD, and not included in this story on purpose). However since it is in the URL, page refreshing will leave you on the same language, as well as navigating through the site, since it will retain the state of your current `lang` param in the URL. 
- I used YAML to parse the language text files which is much cleaner than JSON, which requires putting everything in quotes, brackets and commas. However we should actually have a larger discussion of how these files might get created/edited as there ultimately should be another service for that (a spreadsheet?  or something like https://poeditor.com/) which could then output JSON/etc rather than editing those files directly
- Similarly I started setting up a structure to the translation file so that some global words e.g. "Application" are not namespaced but things within Short form are in the `SHORT_FORM` namespace... there is probably a lot more thought/structure we can put into this to plan for having the site be fully translated and how to best manage that.


menslow:
@kaptron You crushed it! This looks perfect.

PR #189, opened 2016-05-12
Features/add more languages #119289821

kaptron:
DRYed up the language selection by moving options to the controller along with the other options; added more languages to the list per the story.


menslow:
This PR looks good @kaptron

PR #190, opened 2016-05-12
Features/change tagalog to filipino #119392869

kaptron:
Changed all references of "Tagalog" to "Filipino" including the welcome-filipino template and route


menslow:
Looks good @kaptron

PR #193, opened 2016-05-16
Features/update short form nav #116065197

kaptron:
Ended up playing with a few different ways of implementing this, ended up saving in the application (i.e. localStorage) which sections you've completed, which is triggered when you hit the first page of the next section (only implemented for the couple of sections that we currently have in place). For example when you get to "Household intro" then it sets `ShortFormApplicationService.completeSection('You')` to indicate that you've completed the "You" section. 

The logic for enabling/disabling each nav section is based on the following:
(copied this from my comments in shortFormNavDirective)
```
    # Section nav is disabled if:
    # 1. It is not the currently active section AND
    # 2. User can't access (e.g. have not made it that far)
    #  -- OR --
    # 3a. The form is invalid AND
    # 3b. It's not a previous section
```


menslow:
Logic seems sound on how you implemented this @kaptron. I feel we can move forward with this approach and adjust if we see any issues.

PR #195, opened 2016-05-18
Features/listing must close #117730975

menslow:
* Updated process sidebar for desktop and property hero for mobile to display Application Due Date time.
* Adjusted the listingIsOpen function to factor in time along with date.


kaptron:
:+1: Looks great.

PR #196, opened 2016-05-18
Features/status section #108643126

kaptron:
Set up first page of status section with "special programs" checkboxes. Other pages of the section won't come into play until we go past BMR (e.g. Multi-family). There is one visual CSS bug related to the checkboxes being nested inside a `<form>` element which changes the styling in an undesired way, will check w/ @slowbot on that one.


kaptron:
visual bug has also been fixed, no need for @slowbot to get involved!

PR #198, opened 2016-05-19
Features/initial translation capability #118801227

kaptron:
So far only includes first couple pages of short form, but I've figured out the workflow for loading/importing translations and using POEditor.


chrisdolendo:
Cool beans! :+1:

PR #199, opened 2016-05-19
Features/income section #119118085

menslow:
* Added new Income section with simple household income total and per month/year radios.


kaptron:
👍 @menslow looks good, I just updated the `h2` style of the header to match the other pages.

chrisdolendo:
+1

PR #201, opened 2016-05-24
Features/browser close data #116095539

chrisdolendo:


kaptron:
@chrisdolendo 👍 Nice! It's cool that it's as simple as removing the reference to `localStorage` to have it all persist in the browser session and then disappear when you close it. Ran it locally and worked well for me.

chrisdolendo:
thanks @kaptron! will merge when the feature gets accepted

PR #206, opened 2016-05-28
Features/addresses for preferences #116626475

kaptron:
This PR has two main components:
- Setting up the "Work in SF?" and "Same address?" Yes/No radio buttons for You + Household Members, and the logic for what happens if you select Yes/No
- Setting up Address validation using 3rd party service (EasyPost) which will overwrite your entered address with the cleaned up version OR reject it if it's determined to be "undeliverable". **Note**: according to my chat w/ Ashley, we are just doing this for "home address" e.g. "Your address" and not for the "I have a separate mailing address" fields. 

Added some controller tests, still requires better tests for `AddressValidationService` as well as ruby tests for the equivalent API call w/ VCR, etc.


app/assets/javascripts/short-form/ShortFormHelperService.js.coffee
```diff
@@ -1,6 +1,13 @@
 ShortFormHelperService = ($translate, $filter) ->
   Service = {}
 
+  Service.alternate_contact_options = [
```
chrisdolendo: SMART!

chrisdolendo:
@kaptron this looks great! right on for the tests. I guess my only question is if we need to do the EasyPost validation on the Household Members who have separate addresses (since live/work preferences count the household members as well). That's more of a roshen question.. wish I could tag her here lol. I'm guessing if we do, we can capture it in Icebox

PR #207, opened 2016-06-01
Features/proof of preference upload #120245335

chrisdolendo:


kaptron:
@chrisdolendo 👍
I also tested the upload on my phone and it works / looks correct!

PR #208, opened 2016-06-01
Features/updated back button #116556761

kaptron:
Added backButton functions in ShortFormNavService


chrisdolendo:
:+1: the dopest of dope

PR #211, opened 2016-06-04
Bugs/deleting household member #120720411

kaptron:
Cleared out all preference options and proof settings for householdMembers upon their deletion

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
+++ b/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
@@ -10,6 +10,8 @@ ShortFormApplicationService = ($localStorage) ->
     alternateContact: {
       primary_language: "English"
     }
+    householdMembers: []
+    preferences: {}
     completedSections: {
       Intro: false,
       You: false,
@@ -17,7 +19,6 @@ ShortFormApplicationService = ($localStorage) ->
       Status: false,
       Income: false
     }
-    householdMembers: []
   }
   Service.current_id = 0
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
```diff
@@ -100,6 +101,19 @@ ShortFormApplicationService = ($localStorage) ->
   Service.cancelHouseholdMember = ->
     householdMembers = Service.householdMembers.filter (m) ->
       (m != Service.householdMember && m.id != Service.householdMember.id)
+    # clear out preferences where this householdMember was selected
```
chrisdolendo: @kaptron I always wondered if namespacing would work or if we should nest the data structure.

something like 

`application:  { preferences: { preferenceName: [householdmembers] }` 

or something. it takes some work though and not sure if it's worth the work if these are the only preferences we'll be working with.

chrisdolendo:
👍 just had that one question

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
kaptron: @chrisdolendo yeah I totally agree, nested would be better, doing the string check `indexOf('_household_member')` to see if it's a "household member"-type of key is a little bit hacky vs. just marking explicitly that it is `{ preferenceName: { householdMember: "Jane" }}`. I also agree it's not a big deal for what we have. 

You'll also see I slightly refactored the preferences to move them into the `application` namespace instead of `applicant` since they apply to the whole application/household (and not the applicant's own personal data like address, etc.) So you may want to merge this in with the work you're currently doing on #120716389 (or else we can just merge after that story is done).

chrisdolendo: @kaptron coolio. Yeah I am gonna merge this for sure, thanks so much for the heads up!

PR #219, opened 2016-06-09
Features/browser inactivity timeout #119469055

kaptron:
- Uses `ngIdle` plugin to handle most of the functionality.
- Also as part of this story, added a check in angular router to see if you're leaving Short Form, which means you are past the short form intro (e.g. in one of the form sections, "You", "Household", etc.) and you are trying to navigate away to Listings, etc. If you confirm that you want to navigate away, your short form data will be cleared out. 
- Will not trigger the `isLeavingShortForm` if you are navigating to save and finish later aka `dahlia.create-account` which is being set up by @menslow in another story


kaptron:
ALSO: it is set up to timeout after 10 seconds instead of 1 minute, just for testing purposes. Will change after it has been tested.

chrisdolendo:
👍 code makes sense to me!

PR #220, opened 2016-06-09
Features/account create an account markup #119417309

menslow:
* Added markup for create-account view
* Added behavior for returning to application from where you left off
* Added basic controller spec for new AccountController


kaptron:
👍  @menslow looks good to me

PR #230, opened 2016-06-16
Chores/accessibility income calc #121331423

menslow:
Resolved the following accessibility issues for Income Calculator:

* Submit button added to form
* for attribute added to labels for select and text inputs


kaptron:
@menslow 👍  just had to make one update to make the `inputInvalid` method on the IncomeCalculatorController work with the accessibility updates to the input `name` attributes.

PR #234, opened 2016-06-18
Features/voucher skips income check #121625907

kaptron:


app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -266,6 +266,8 @@ ShortFormApplicationController = (
   $scope.validateHouseholdEligibility = (match, callbackUrl) ->
     $scope.clearHouseholdErrorMessage()
     form = $scope.form.applicationForm
+    if match == 'incomeMatch' && $scope.applicant.household_vouchers_subsidies == 'Yes'
+      $state.go(callbackUrl)
     ShortFormApplicationService.checkHouseholdEligiblity($scope.listing)
```
kaptron: ^^ this is really the only change needed for this feature, however because of some changes on the Salesforce side I also ended up regenerating some of the VCR cassettes/fixtures and did some other minor refactoring / typo-fixing so that also went into this PR

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -266,6 +266,8 @@ ShortFormApplicationController = (
   $scope.validateHouseholdEligibility = (match, callbackUrl) ->
     $scope.clearHouseholdErrorMessage()
     form = $scope.form.applicationForm
+    if match == 'incomeMatch' && $scope.applicant.household_vouchers_subsidies == 'Yes'
```
chrisdolendo: shouldn't this be `if match == 'incomeMatch' || $scope.applicant.household_vouchers_subsidies == 'Yes'` as opposed to `&&`?

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -266,6 +266,7 @@ ShortFormApplicationController = (
   $scope.validateHouseholdEligibility = (match, callbackUrl) ->
     $scope.clearHouseholdErrorMessage()
     form = $scope.form.applicationForm
+    # skip the check if we're doing an incomeMatch and the applicant has vouchers
     if match == 'incomeMatch' && $scope.applicant.household_vouchers_subsidies == 'Yes'
```
kaptron: @chrisdolendo to your previous comment, no... I added a clarifying comment here. The logic is "if we're doing an incomeMatch check, *and* the person said Yes to having vouchers, then skip the check". If it was a `||` then it would always skip ahead any time you were trying to validate household income regardless of their Yes/No answer to vouchers.

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -266,6 +266,9 @@ ShortFormApplicationController = (
   $scope.validateHouseholdEligibility = (match, callbackUrl) ->
     $scope.clearHouseholdErrorMessage()
     form = $scope.form.applicationForm
+    # skip the check if we're doing an incomeMatch and the applicant has vouchers
```
kaptron: @chrisdolendo to your previous comment, no... I added a clarifying comment here. The logic is "if we're doing an incomeMatch check, and the person said Yes to having vouchers, then skip the check". If it was a || then it would always skip ahead any time you were trying to validate household income regardless of their Yes/No answer to vouchers.

chrisdolendo: @kaptron oh lol for some reason I was reading it as if the income passed validation, which is so silly coz i wrote this function. looks good to me!

chrisdolendo:
:+1:

PR #239, opened 2016-06-21
Features/don't automatically check preferences #121541653

kaptron:


chrisdolendo:
looks dope! but slightly painful to see tests go hahaha jk :+1:

kaptron:
@chrisdolendo hah I know, it was sad to delete such nice tests. I poured out some Philz ☕  as I sent them off to git heaven. 👼

PR #240, opened 2016-06-21
Features/grab geocode for charlies api #120725321

kaptron:


app/assets/javascripts/short-form/GeocodingService.js.coffee
```diff
@@ -4,13 +4,14 @@
 
 GeocodingService = ($http) ->
   Service = {}
-  Service.geocoding_data = {}
 
   Service.geocode = (params) ->
     $http.post('/api/v1/geocode-address.json', params).success((data, status, headers, config) ->
-      angular.copy(data.geocoding_data, Service.geocoding_data)
+      # append geocoding_data to address
+      params.address.geocoding_data = data.geocoding_data
```
chrisdolendo: @kaptron This is cool -- so you don't need to explicitly say which model you're assigning the geocoding data - whether it's the  applicant's home address or the household member's address (e.g. ShortFormService.applicant.home_address.geocoding_data) because `params.address` is just a pointer to the object?

chrisdolendo:
this looks cool to me! just had one question. :+1: nice that we are a step closer to doing the neighborhood boundary check

app/assets/javascripts/short-form/GeocodingService.js.coffee
kaptron: correct!

PR #253, opened 2016-06-30
Features/shortform general lott notice #121066365

chrisdolendo:


app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
```diff
@@ -111,6 +111,11 @@ ShortFormNavigationService = ($state) ->
           'household-members'
         else
           'household-intro'
+      when 'general-lottery-notice'
+        'status-programs'
+      when 'status-vouchers'
```
chrisdolendo: I simplified the back state for this. ideally id be able to call $scope.checkIfPreferencesApply (couldn't coz couldn't inter-inject dependency) and determine route from there, but routing to status-programs is a catch-all. open to suggestions!

chrisdolendo:
yay! story was accepted -- waiting for a review.

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
kaptron: @chrisdolendo I went with your idea and refactored `preferencesApplyForHousehold` to be in `ShortFormApplicationService`... however then I realized this function is in `ShortFormNavigationService`. But, it's totally fine for services to include each other, and in fact it probably makes sense for `ShortFormNavigationService` to have access to the AppService as well (like for example it's sort of unnecessary to pass `$scope.application` as a param to `Service.backPageState` when it could just get that directly from the AppService, we could refactor that). Likewise I realize I've put in some "navigation-y" features in AppService because they also needed access to AppService functions, like `authorizedToProceed` uses `userCanAccessSection` -- however if NavService can tap into these functions then problem solved!
Long story short -- this works now!

kaptron:
👍  for this feature. See my comments, ShortFormNavService and AppService could probably use some further refactoring (unrelated to this feature, but something that was surfaced).

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
chrisdolendo: @kaptron I like what you did -- it was not possible for me to inject Nav to Ctrl and Ctrl to Nav, but you did inject services and that made it much easier.

And yes we can make it a chore to refactor it further but this works! merging.

PR #254, opened 2016-06-30
Features/application submit sync #108846424

chrisdolendo:


app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
```diff
@@ -321,6 +322,187 @@ ShortFormApplicationService = ($localStorage, $translate, $http, Upload, uuid) -
...
+  Service.submitApplication = (listingId)->
+    applicationParams = Service._formatApplication(listingId)
+    console.log(applicationParams)
+    $http.post('/api/v1/short-form/submit-application', applicationParams).success((data, status, headers, config) ->
+      data
+    ).error( (data, status, headers, config) ->
+      return
+    )
+
+  Service._formatApplication = (listingId) ->
```
chrisdolendo: I'll probably want to pull this out into it's own helper/service in the future, especially when we receive app info -- but kept it here for now so as not to setup a new test file.

kaptron: @chrisdolendo agreed!

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -33,6 +33,11 @@ def delete_proof
     end
   end
 
+  def submit_application
+    response = SalesforceService.submit_application(params)
```
kaptron: @chrisdolendo this really should send something like `application_params` which performs some checks e.g. `require(:application)`, `permit(...)` rather than sending through the entire `params` object without any checks. For example this object includes things like `controller` and `action` which end up getting sent along as well. It will be slightly annoying in the sense that we will want to double check all of the permitted field names inside of the Rails controller much like we're already doing in JS, but it's somewhat necessary.

chrisdolendo: good idea @kaptron  -- i'll update right now.

app/services/salesforce_service.rb
```diff
@@ -82,8 +82,8 @@ def self.submit_application(params)
     api_post(endpoint, params)
   end
 
-  def self.api_get(endpoint, params = nil, parse_response = true)
-    response = oauth_client.get(endpoint, params)
+  def self.api_call(method = :get, endpoint, params, parse_response)
```
chrisdolendo: nice refactor @kaptron!

PR #255, opened 2016-07-01
Features/visual design accounts #121065403

menslow:
Add design updates and static markup has been added to My Accounts section

* New routes added for my-account, my-application, my-favorites, eligibility-settings, and account-settings
* Static mark added for all my account pages
* Non-functional My Account dropdown menu in place
* Mobile nav items for My Account added


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
```
kaptron: @menslow looks like you did a `bundle update` which updated all the gems, generally not recommended since it can be hard to track down issues if something is suddenly broken -- so far using the site it seems like it's ok though. If this broke something then we could roll this back.

kaptron:
@menslow looks like you did a `bundle update` which updated all the gems, generally not recommended since it can be hard to track down issues if something is suddenly broken -- so far using the site it seems like it's ok though. If this broke something then we could roll this back.

chrisdolendo:
:+1:

kaptron:
@menslow @slowbot (+mari) -- these pages look great! I did notice some awkward interaction with the "My Account" navigation menu + dropdown though. You have to click "My Account" in order to open the dropdown, however it also takes you to the "My Account" page when you click on it. E.g. if I'm trying to get to My Favorites I click once to open the dropdown, which also navigates to My Account, and then the menu is open and I can now click My Favorites. Also, there's no way to close the menu by clicking on "My Account" again, you have to make a selection to close it. 

Thoughts? I can see the benefit of going straight to My Account by clicking on the menu item (vs. having a separate menu item in the dropdown like "Dashboard", which would be a different solution) however it feels weird to have the click take you to the page and open the dropdown at the same time.

PR #256, opened 2016-07-01
Features/account email verification #123677291

kaptron:


app/assets/javascripts/directives/equalizer.js.coffee
```diff
@@ -0,0 +1,11 @@
+angular.module('dahlia.directives')
```
chrisdolendo: is this to help the accounts page be responsive?

chrisdolendo:
:+1: big feature! just had that one question about `equalizer`.

kaptron:
@chrisdolendo this _is_ a big feature, however it looks slightly bigger than it is (on its own) because it includes #255 (so that being logged in could take you to somewhere, to the newly designed static pages + new menu). @menslow probably knows more about the equalizer directive since he added it, however equalizers are generally for height matching, my assumption is that it's for the 4 squares on "My Account" -- equalizer will make them match the same height so they look nice and lined up.

chrisdolendo:
@kaptron I see it was branched off another PR. cool!

PR #260, opened 2016-07-07
Chores/new salesforce lottery endpoints #125019981

kaptron:


chrisdolendo:
:+1: application submit is merged so we can merge this as well when you are ready :)

PR #261, opened 2016-07-07
Features/shortform sexual orientation #124859691

chrisdolendo:


menslow:
Looks good @chrisdolendo!

PR #264, opened 2016-07-09
[#123151561] - Updated portal logo

chrisdolendo:


menslow:
Rock on! @chrisdolendo

PR #263, opened 2016-07-09
Features/confirmation email on submission #120638465

kaptron:

PR #275, opened 2016-07-15
Features/upload proof of preference #122179037

kaptron:
is branched off #263


app/services/salesforce_service/base.rb
```diff
@@ -1,31 +1,24 @@
...
 
     def self.oauth_client
-      client = Restforce.new(
+      Restforce.new(
         authentication_retries: 0,
         oauth_token: oauth_token,
         instance_url: ENV['SALESFORCE_INSTANCE_URL'],
         mashify: false,
       )
-      client.middleware.use :custom_headers, headers if headers.present?
```
chrisdolendo: Why did we need to get rid of this?

config/initializers/salesforce_services.rb
```diff
@@ -0,0 +1,6 @@
+### shortcut to alias SalesforceServices
+# NOTE: if you update any of these services (new methods, etc),
+#       you will need to restart spring in order for the alias to "refresh"
+AccountService = SalesforceService::AccountService
```
chrisdolendo: cool i actually might need to update my pr since i declare this in the devise overrides registrations ctrl

chrisdolendo:
:+1: looks GREAT!

app/services/salesforce_service/base.rb
kaptron: Yup. This `custom_headers` thing I came up with didn't seem to properly send the headers through to salesforce 😞  so I got rid of it -- the `api_post_with_headers` function replaces it.

config/initializers/salesforce_services.rb
kaptron: It's fine to declare at the top of the controller too, doesn't hurt anything... in fact the benefit there is it _will_ reload the service definition every time you make a request, so while in dev, if you're updating the service you don't have to restart your server to have the alias pick up the new changes. :shrug:

PR #276, opened 2016-07-17
Features/account creation salesforce sync #121701649

chrisdolendo:


kaptron:
cool! 👍 like ashley mentioned in the pivotal story, I'm also not seeing where DOB comes through on the salesforce end. However everything looks good and works re: connecting contactId from salesforce. I just added one minor visual update to `.form-group` for Your Name where it shows the error state if those fields are missing.

PR #279, opened 2016-07-19
Features/prepopulate name dob email #126175209

kaptron:
I thought about refactoring some shared form functions between ShortFormApplicationController and AccountController, namely `handleErrorState` and `inputInvalid` ... however I spent about 10 mins going down that path and realized it was going to be a little more pain than it's worth. So for now they just share a couple of similar functions.


chrisdolendo:
:+1: looks good to me!

PR #285, opened 2016-07-21
Features/prepopulate from account #126178839

kaptron:
branched from #279

### DIFF
--- a/app/assets/javascripts/AngularConfig.js.coffee
+++ b/app/assets/javascripts/AngularConfig.js.coffee
@@ -113,6 +113,9 @@ angular.module('dahlia.controllers',['ngSanitize', 'angular-carousel', 'ngFileUp
         'container@':
           templateUrl: 'account/templates/create-account.html'
           controller: 'AccountController'
+      onEnter: ['AccountService', (AccountService) ->
+        AccountService.unlockFields()
+      ]
     })
     .state('dahlia.short-form-application.create-account', {
       # duplicated from above but to differentiate state for "Save and finish later"
@@ -122,6 +125,11 @@ angular.module('dahlia.controllers',['ngSanitize', 'angular-carousel', 'ngFileUp
         'container@':
           templateUrl: 'account/templates/create-account.html'
           controller: 'AccountController'
+      onEnter: ['AccountService', (AccountService) ->
+        AccountService.copyApplicantFields()
+        AccountService.lockCompletedFields()
+      ]
+
     })
     .state('dahlia.sign-in', {
       url: '/sign-in'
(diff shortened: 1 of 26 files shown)


app/assets/javascripts/AngularConfig.js.coffee
```diff
@@ -660,4 +671,5 @@ angular.module('dahlia.controllers',['ngSanitize', 'angular-carousel', 'ngFileUp
       apiUrl: '/api/v1'
       storage: 'sessionStorage'
       confirmationSuccessUrl: conf.confirmationSuccessUrl
+      validateOnPageLoad: false
```
chrisdolendo: what does this do?

app/assets/javascripts/AngularConfig.js.coffee
```diff
@@ -581,7 +592,7 @@ angular.module('dahlia.controllers',['ngSanitize', 'angular-carousel', 'ngFileUp
 @dahlia.run [
   '$rootScope', '$state', '$window', '$translate', 'ShortFormApplicationService', 'AccountService', 'ShortFormNavigationService',
   ($rootScope, $state, $window, $translate, ShortFormApplicationService, AccountService, ShortFormNavigationService) ->
-    # check if user is logged in
+    # check if user is logged in on page load
     AccountService.validateUser()
```
kaptron: @chrisdolendo see here -- so the `validateOnPageLoad` is an option for ngTokenAuth, it will automatically check if there is a logged in user session on page load. However we do that manually ourselves on pageLoad via `AccountService.validateUser()` so that we can also do some custom stuff like call `ShortFormApplicationService.importUserData(Service.loggedInUser)` which will pull in the logged in user's data into the `Service.application` object, or anything else we might want to do.

app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -30,35 +31,75 @@ AccountService = ($state, $auth) ->
...
+    Service.userAuth.DOB = formattedDOB
+
+  # reverse of the above function
+  Service._reformatDOB = ->
+    return false if _.isEmpty(Service.loggedInUser)
+    split = Service.loggedInUser.DOB.split('-')
+    Service.loggedInUser.dob_year = parseInt(split[0])
+    Service.loggedInUser.dob_month = parseInt(split[1])
+    Service.loggedInUser.dob_day = parseInt(split[2])
+
```
kaptron: @chrisdolendo ^^ speaking of formatting/re-formatting, massaging/re-massaging... I wonder if all of those functions could get pulled out into their own `DataMassagingService` (or whatever) because it's useful both in the `AccountService` and `ShortFormApplicationService` and I sense the opportunity for some DRY-ness.

PR #292, opened 2016-07-27
Features/verify email modal #126810619

kaptron:
branched from #287


app/assets/javascripts/directives/alertBox.js.coffee
```diff
@@ -44,8 +44,8 @@ angular.module('dahlia.directives')
         $translate.instant("ERROR.ALT_CONTACT_TYPE")
       else if scope.householdEligibilityErrorMessage
         $translate.instant("ERROR.NOT_ELIGIBLE") + " " + scope.householdEligibilityErrorMessage
-      else if scope.accountError && (scope.accountError.message == 'Email already in use')
-        $translate.instant("ERROR.EMAIL_ALREADY_IN_USE")
+      else if scope.accountError && scope.accountError.message
```
chrisdolendo: nice refactor

app/controllers/overrides/registrations_controller.rb
```diff
@@ -24,7 +24,10 @@ def sync_with_salesforce
           salesforce_contact_id: salesforce_contact['contactId'],
         )
       else
-        @resource.destroy
+        # TODO: Fix this error handling.
```
chrisdolendo: should we capture in a chore and put to icebox?

kaptron: done.

chrisdolendo:
@kaptron does this feature also work when a user saves and finishes later and signs in with an unconfirmed account?

kaptron:
@chrisdolendo yeah, I mean... _trying_ to save and finish later and sign in unconfirmed won't actually let you sign in, so you won't really be "saving" your application anyway. I don't think there's anything different we have to account for here (it's the same sign in page + controller).

chrisdolendo:
@kaptron cool yeah -- I was just wondering if the modal will popup as well in that workflow, looks like it does! :+1:

PR #294, opened 2016-07-28
[hotfix] - fix for save and finish later for logged in users; prevent…

kaptron:
… from trying to submit the short form page and showing the "are you sure you want to leave?" popup


chrisdolendo:
:+1: good catch!

PR #295, opened 2016-07-28
Features/my applications #120248083

kaptron:


app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -92,6 +93,12 @@ AccountService = ($state, $auth, $modal, $http, ShortFormApplicationService) ->
       return
     )
 
+  Service.getMyApplications = ->
```
chrisdolendo: is there a reason why you chose `myApplications` as opposed to `userApplications` or `currentUserApplications`?

app/controllers/api/v1/account_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def my_applications
+    applications = current_user_applications
+    listing_ids = applications.collect { |a| a['listingID'] }.uniq
+    listings = ListingService.listings(listing_ids.join(','))
+    applications.each do |app|
+      app['listing'] = listings.find { |l| l['listingID'] == app['listingID'] }
+    end
+    render json: { applications: applications }
+  end
```
chrisdolendo: so for this action, we're getting all the applications of the current user, 
then we're getting the listing IDs of those applications,
then we're getting the listing objects of those listings and mapping them into the each corresponding application, so that each application has all the listing details.

very cool! wondering if we can make this a private method like `map_listing_data_to_applications`.

spec/factories/user.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
```
chrisdolendo: :clap:

app/assets/javascripts/account/AccountService.js.coffee
kaptron: it's a good question... I have some things like the private method in `account_controller.rb` called `current_user_applications` however everything else I kind of used the "my applications" wording. I guess I liked the semantic nature of it, "get my applications" describes what it's doing (from the user's perspective) and lines up with the name of the page. But I agree we should make sure we have consistent naming conventions as we add more user features. Like do we call other features "my {xxx}" and "get my {xxx}"...

spec/support/devise_macros.rb
```diff
@@ -0,0 +1,10 @@
+# Spec helpers for Devise Token auth
+module DeviseMacros
```
chrisdolendo: :+1: surprised this didn't come with devise token auth!

app/assets/javascripts/account/templates/my-applications.html.slim
```diff
@@ -8,44 +8,4 @@
...
-              .feed-item-action
-                p.feed-item-status
-                  span.feed-item-label.is-submitted Status: Submitted
-                a.button.small href="#"  View Application
-            footer.feed-item-footer
-              .feed-item-links
-                a.feed-item-link.lined href="#"  See Listing
-              .feed-item-meta
-                p.feed-item-date Submitted: January 12th 2016
+        my-application ng-repeat="application in myApplications" application="application"
```
chrisdolendo: power of directives!

app/controllers/api/v1/account_controller.rb
kaptron: @chrisdolendo good call! yeah I didn't even think about the fact that we'd need to separately grab the listing data when first starting this feature, and I started doing it a couple different ways before settling on this one, but agree it could be refactored. Done! To please the private method gods and goddesses! 🙏  💃

spec/support/devise_macros.rb
kaptron: @chrisdolendo yeah both Devise and Token Auth each give instructions of how to set up tests (and I got part of this from some comments in a thread on Token Auth) but they don't provide anything specific for some of this stuff since you could be using different test suites, etc... just some examples. Also most "login" examples are for controller tests (not request tests) so it took me a while to get this working and get it right 😌

app/assets/javascripts/account/AccountService.js.coffee
chrisdolendo: hmm.. I see your point. I just always prefer 3rd person naming as opposed to 1st person -- that way we know which player it is. Right now we have one player which is the user, but what if in the future we create an agent portal or a admin portal or something, in which case the 3rd person naming might be helpful. 

But if we ever add more players, we can easily rename things as they are needed. So I'm cool with it staying this way, I guess so long as all exclusively logged-in user functions follow the same naming.

chrisdolendo:
:+1: very cool, just had a few Qs!

spec/support/devise_macros.rb
chrisdolendo: :clap:

PR #311, opened 2016-08-10
Features/forgot password #119417319

chrisdolendo:


config/environments/development.rb
```diff
@@ -35,6 +35,19 @@
...
+      port:           '587',
+      authentication: :plain,
+      user_name:      ENV['SENDGRID_USERNAME'],
+      password:       ENV['SENDGRID_PASSWORD'],
+      domain:         'heroku.com',
+      enable_starttls_auto: true,
+    }
+  end
+
   # Raises error for missing translations
```
kaptron: @chrisdolendo did you add this just to test real emails rather than use the server log (which understandably can be kind of hard to read)? I should also show you the `EmailerPreview` class which makes it a lot easier to test these emails without having to actually send them off

kaptron:
@chrisdolendo you can also see in `emailer.rb` for `confirmation_instructions` how I set it up to hit Salesforce to pull in first+lastName, we could use the same thing here.

app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -51,6 +51,26 @@ AccountService = ($state, $auth, $modal, $http, $translate, ShortFormApplication
...
+      # handle error response
+    return
+
+  Service.updatePassword = ->
+    params =
+      password: Service.userAuth.password
+      password_confirmation: Service.userAuth.password_confirmation
+
+    $auth.updatePassword(params).then((resp) ->
+      $state.go('dahlia.my-applications')
```
kaptron: I think this should be `dahlia.my-account`? I didn't see any info in the story instructions as to where it should redirect, but that's the usual place to redirect after login -- we only redirect to `my-applications` after you submit an application (so you can see it in there).

kaptron:
tried it out, everything with the password reset flow worked great! one thing that I noticed is that the password + password confirmation doesn't seem to have frontend validation, e.g. for "passwords don't match". Should be (relatively) easy to pull this over from the create-account form.

config/environments/development.rb
chrisdolendo: @kaptron Yeah I knew that they were sending via the server log, but I needed to click on the link in the email to do a full cycle test locally to make sure of correct routing and resetting of the password. If EmailerPreviewer I can do this I can remove this. :+1:

app/assets/javascripts/account/AccountService.js.coffee
chrisdolendo: @kaptron yeah in the story it says to be directed to `my-applications`. See:

```
'update password' button is built but non-functional (becomes functional when 'my applications' page is built since that is where it will take the user) -
```

Although I tend to agree with you. I would prefer `my-account`.

Re userAuth namespacing: sounds good -- I will the reset password stuff to use `userAuth.user` :dancer:

chrisdolendo:
@kaptron updated the repo with the following changes:
1. fixed the password confirmation bug -- turns out I needed to include the form in `$scope.accountForm = -> as well.`
2. removed send grid creds
3. updated password reset email with first and last name
4. created error messages for non-existing/non-confirmed emails on forgot password form

did a q/a on the review app and everything works great! if you are cool, will merge to master.

PR #301, opened 2016-08-05
Features/continue editing draft application #126268001

kaptron:
Not quite ready to merge yet. Also includes a couple of the other open PRs #299 #300

PR #312, opened 2016-08-10
Features/account change name dob #127691091

kaptron:
Yet another one that continues off the other work done in #301, #307. 

Still WIP -- Name + DOB will update but it just gives off a browser alert message and not the proper notification.


app/assets/javascripts/account/AccountController.js.coffee
```diff
@@ -89,12 +90,11 @@ AccountController = ($scope, $state, $document, $translate, AccountService, Shor
       $scope.handleErrorState()
 
   $scope.updateEmail = ->
-    # form = $scope.form.accountEmail
     $scope.form.current = $scope.form.accountEmail
-    AccountService.updateAccount()
+    # AccountService.updateAccount()
+    false
```
chrisdolendo: @kaptron is there a reason why you returned false here -- was that just something temp for the feature? Can i revert it back? Might be good to go through some of this with you before I deliver the story

chrisdolendo:
This one is about ready to go, but need to go over some weird behavior with you when trying to access account settings or my application right after confirming account by clicking on the email link. I have an idea why but need your opinion.

PR #316, opened 2016-08-16
Features/update email #127691179

kaptron:


chrisdolendo:
we did this together so 👍

PR #317, opened 2016-08-16
Features/confirm email update #127691269

kaptron:
WIP for now, just creating PR


app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -117,12 +122,31 @@ AccountService = ($state, $auth, $modal, $http, $translate, ShortFormApplication
...
+    if infoType == 'email'
+      params =
+        user:
+          email: Service.userAuth.user.email
+      $http.put('/api/v1/auth', params).success((data) ->
+        Service.accountSuccess.messages.email = $translate.instant("ACCOUNT_SETTINGS.VERIFY_EMAIL")
+      ).error((response) ->
+        msg = response.errors.full_messages[0]
+        if msg == 'Email has already been taken'
+          Service.accountError.messages.email = $translate.instant("ERROR.EMAIL_ALREADY_IN_USE")
```
chrisdolendo: nice re-use of existing translations

app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -13,8 +13,11 @@ AccountService = ($state, $auth, $modal, $http, $translate, ShortFormApplication
   Service.myApplications = []
   Service.createdAccount = {}
   Service.rememberedShortFormState = null
-  Service.accountError = {message: null}
-  Service.accountSuccess = {message: null}
+  Service.accountError =
+    message: null
```
chrisdolendo: YAY refactoring the alert-box and error/success messages! Notice that you have message and messages -- and some using message and messages. I kind of see the benefit of having both, but wondering if it's worth just using messages. I'm good either way.

app/mailers/emailer.rb
```diff
@@ -17,27 +17,33 @@ def submission_confirmation(params)
...
   def confirmation_instructions(record, token, opts = {})
-    contact = AccountService.get(record.salesforce_contact_id)
-    @name = if contact.present?
-              "#{contact['firstName']} #{contact['lastName']}"
-            else
-              record.email
-            end
-    super
+    load_salesforce_contact(record)
+    if record.pending_reconfirmation?
```
chrisdolendo: nice re-use of emailer method. Is `pending_confirmation` a devise method that is determined by `email_changed?`.

app/assets/javascripts/account/AccountService.js.coffee
kaptron: yeah! didn't totally refactor but this is definitely a start. 

and I agree, I figured "baby steps" and not reworking all the account messages at once, and let it use `message` for the previous features (sign in, create account, forgot password) without breaking anything. However yes that would probably be even better to consolidate all of that.

app/mailers/emailer.rb
kaptron: yeah, `pending_reconfirmation` is basically checking if there's an `unconfirmed_email` present. 

However by default devise uses the same email for confirmation and **re**confirmation so this is what allows us to use 2 different emails.

app/models/user.rb
```diff
@@ -7,7 +7,47 @@ class User < ActiveRecord::Base
...
+    confirmable.confirm if confirmable.persisted?
+    confirmable
+  end
+
+  def sync_reconfirmation_with_salesforce
+    if pending_reconfirmation? && persisted?
+      # TODO: we should not have to grab the existing applicant first.
+      #  This is currently necessary because salesforce is rejecting an
+      #  update to just the email field, saying we need to pass first/last name as well.
+      #  A bug has been filed for this.
```
kaptron: ^^ turns out this is not a bug, and just a necessary thing based on how Salesforce API works. I can remove this comment. **shrug**

app/models/user.rb
```diff
@@ -7,7 +7,47 @@ class User < ActiveRecord::Base
...
   def error_details(field)
     errors.details[field].collect { |i| i[:error] }
   end
+
+  def email_changed?
+    initiate_email_reconfirmation.present?
+  end
+
+  # override from Devise so that we can add hook into Salesforce update
+  def self.confirm_by_token(confirmation_token)
```
chrisdolendo: so with these two methods, when the user confirms their email, the model syncs with salesforce (line 29) - yeah?

This will be usable for the upcoming password reset confirm email story as well right?

app/views/devise/mailer/reconfirmation_instructions.html.slim
```diff
@@ -0,0 +1,11 @@
+p
+  = "#{t(:hello).capitalize} #{@name},"
+p
+  = t '.confirm_link_msg'
```
chrisdolendo: I will follow this pattern when doing future emails (for translations)

app/models/user.rb
chrisdolendo: oh wells. Salesforce....

app/models/user.rb
kaptron: first part of your question -- yes. 

second part, for password changing we don't need to do anything with salesforce though, right?

chrisdolendo:
:+1: Very cool. Thanks so much for clarifying and sad I wasn't able to pair with you on this. But there will be more opportunities. Thanks!

app/models/user.rb
chrisdolendo: ah yes you're right. nothing with salesforce, but i believe the story asks for the user to confirm their password change via email see: https://www.pivotaltracker.com/story/show/127691397

PR #320, opened 2016-08-16
[#127691325] - Account Name/DOB update - send email success

chrisdolendo:


app/mailers/emailer.rb
```diff
@@ -18,21 +29,13 @@ def submission_confirmation(params)
 
   def confirmation_instructions(record, token, opts = {})
     contact = AccountService.get(record.salesforce_contact_id)
-    @name = if contact.present?
-              "#{contact['firstName']} #{contact['lastName']}"
-            else
-              record.email
-            end
+    @name = name(contact, record)
```
kaptron: aha, we both refactored this function in each of our PRs 😄 

we can merge it afterwards when they're combined.

kaptron:
one comment that is really more MOHCD related, which is consistency on these emails, some say "Hello ...", some say "Dear...", etc. 

the other thing is related to using the Rails translations file for the emails which, even though it's totally separate from the Angular translation file, it's probably good to use. It looks like we're currently using it for the Devise override emails but not for some of these custom ones. So we should probably refactor these ones (submission_confirmation, and this one).

app/mailers/emailer.rb
chrisdolendo: ha! nice. sounds good.

chrisdolendo:
@kaptron yes it would be nice to have a rails translations file too.. I will capture as a chore.

As for the "Hello" vs "Dear" -- do you want me to just stick to Dear and I can update all the one that says "Hello" in this PR?

kaptron:
@chrisdolendo you'll see we actually already _are_ using the translations file but just for the devise override emails and not the more custom ones. 

And actually I set up some of those to have a consistent signature, like:

```
p
  = "#{t(:thank_you).capitalize},"
  br
  = t 'devise.mailer.signature'
```

But yeah let's just capture a chore for this, it can be both of these things combined into one chore, since they're related: 

- Use translations file for all emails
- Consistent intro and signature for all emails

chrisdolendo:
@kaptron oh I see -- gotcha! will capture a chore for both.

PR #321, opened 2016-08-17
[#127770411] - Fix for auth token becoming invalid after confirmation…

kaptron:
… also "sign out" user if there was an error signing out

### DIFF
--- a/app/controllers/overrides/confirmations_controller.rb
+++ b/app/controllers/overrides/confirmations_controller.rb
@@ -10,8 +10,11 @@ module Overrides
         # create client id
         add_resource_token
 
-        yield if block_given?
+        yield @resource if block_given?
 
+        # we want to disable the after_action,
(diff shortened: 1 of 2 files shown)


kaptron:
@chrisdolendo @menslow this is a fix for a bug that I had originally reported, so I will merge this in. however feel free to check out the PR/commit. It was a long painful process of debugging that ended up with just a couple lines of code.

app/controllers/overrides/confirmations_controller.rb
```diff
@@ -10,8 +10,11 @@ def show
         # create client id
         add_resource_token
 
-        yield if block_given?
+        yield @resource if block_given?
```
kaptron: not 100% sure what this does but this is how it's done in the newer version of the gem. didn't actually seem to affect anything related to what we're using.

chrisdolendo:
Huge kudos @kaptron!

PR #323, opened 2016-08-17
include README install instructions to create database and ignore warning

mluedke2:
A few small changes to the `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,11 @@ We will create a Boxen for this project, but for now to get started:
 4. Run `overcommit --install` to install git hooks into the repo
 5. Run `bower install`
 6. Run `grunt`. Ensure that the *latest* pattern library is in the same folder as this repo.
-Running grunt will migrate the css over from the patter library.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,8 +30,11 @@ We will create a Boxen for this project, but for now to get started:
 4. Run `overcommit --install` to install git hooks into the repo
 5. Run `bower install`
 6. Run `grunt`. Ensure that the *latest* pattern library is in the same folder as this repo.
-Running grunt will migrate the css over from the patter library.
+Running grunt will migrate the css over from the pattern library. You can ignore the warning that says `Unable to match 1 pattern.`.
 6. Run `rake bower:install`
```
menslow: Looks like there are two 6. bullet points. Given that it's markdown, I think you can convert them all the 1. and the numbering should work.

menslow:
We have a Github/Pivotal Tracker integration which tracks all your commits for a particular story, so in the future prefix your commit messages with: [#128535881] and Tracker will connect it to the story.

Other than that, things look good.

PR #326, opened 2016-08-18
Features/create account after submit #127841979

kaptron:


.rubocop.yml
```diff
@@ -54,7 +54,7 @@ Style/ConditionalAssignment:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 80
+  Max: 90
```
chrisdolendo: haha

kaptron: 😁  Yeah there was a line it was complaining about (one that I was overriding/copying from the gem) and I was like "JEEZ RUBOCOP IT LOOKS FINE", style guidelines are really up to us so if it's becoming more of a nuisance than a help, we can feel free to tweak them to our liking.

PR #328, opened 2016-08-18
[#126186999] Removed living with HIV question from short form. Remove…

menslow:
* Removed living with HIV question from short form
* Removed Living with HIV label from translation file
* No related Controller, Service or Jasmine code needed to cleanup.


kaptron:
there were just a few `hiv` field related things worth cleaning up, particularly the one that checks `surveyComplete` which will no longer be valid for the `hiv` question. I also updated the "sexual orientation" question to match the updated styling of the other questions on that page.

PR #332, opened 2016-08-22
Features/update contact info short form #123974501

chrisdolendo:


kaptron:
looked at this with you yesterday, 👍

PR #344, opened 2016-08-26
[#127513221] - Updated error message for Address Validation; set up s…

kaptron:
…ervice to return more helpful errors in the event that we want to use them

(this is for the PO Box error)

### DIFF
--- a/app/services/address_validation_service.rb
+++ b/app/services/address_validation_service.rb
@@ -18,4 +18,11 @@ class AddressValidationService
     !@validation.verifications.delivery.success ||
       @validation.street1.include?('PO BOX')
   end
+
+  def error
+    return nil unless @validation.present? &&
+                      @validation.verifications.delivery.errors.present?
+    return 'PO BOX' if @validation.street1.include?('PO BOX')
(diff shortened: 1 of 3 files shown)


app/services/address_validation_service.rb
```diff
@@ -18,4 +18,11 @@ def invalid?
     !@validation.verifications.delivery.success ||
       @validation.street1.include?('PO BOX')
   end
+
+  def error
+    return nil unless @validation.present? &&
+                      @validation.verifications.delivery.errors.present?
+    return 'PO BOX' if @validation.street1.include?('PO BOX')
```
chrisdolendo: You return 'PO BOX' if it includes that, but is there a custom message for indicating this or is the ERROR.ADDRESS_VALIDATION always the default error message?

kaptron: Yeah, I thought the acceptance for the story was to give different error messages for both, so I set up this error message. However I realized it was really just saying "make sure we check for PO Box and also update the text of the error message." So I figure just leave in this functionality in case we ever do want to use it to give more descriptive errors.

PR #355, opened 2016-08-29
Features/logged in idle #127514293

chrisdolendo:


chrisdolendo:
I've started to think about what is the best way to test this -- should we start thinking about E2E tests when we are done with the release?

kaptron:
@chrisdolendo re: E2E testing, yes probably! Yeah it would be pretty great to set up some tests that were actually simulating clicks and browsing through the site (e.g. w/ fake salesforce data) -- it wouldn't account for salesforce errors but would at least ensure that some frontend features are working as they're supposed to. And we'll have to figure out some clever ways of testing the idling stuff (like without it having to wait 2 minutes to test it out) 😄

app/assets/javascripts/application.js
```diff
@@ -41,4 +41,7 @@
 //= require ng-token-auth
 //= require angular-ui-validate
 //= require angular-uuids
+//= require config/angularModules.js
+//= require config/angularRoutes.js
+//= require config/angularInitialize.js
```
kaptron: 🙌

kaptron:
👍

PR #357, opened 2016-08-29
[#127689143] - updated instructional text and proof dropdown options …

kaptron:
…for live/work preferences


chrisdolendo:
very nice  👍

PR #358, opened 2016-08-30
[##128547253] - Updated lottery search UX to make field always editable

chrisdolendo:


chrisdolendo:
Let's not merge this one until Jesse is good with the UX -- aka has accepted the story.

kaptron:
seems to work well w/ keyboard input, however it seems like we've lost the "X" that comes up after you've searched which seems useful to let you know you can clear the value, rather than having to delete the number and search again.

PR #360, opened 2016-08-30
[hotfix] - choose draft alert messages were accidentally swapped

kaptron:
just a fix because jesse noticed that the alert title and message were swapped.


chrisdolendo:
:+1:

PR #364, opened 2016-08-31
[#127513973] - updated eligibility alert boxes to not get accidentall…

kaptron:
…y triggered

### DIFF
--- a/app/assets/javascripts/short-form/templates/c1-household-intro.html.slim
+++ b/app/assets/javascripts/short-form/templates/c1-household-intro.html.slim
@@ -3,7 +3,7 @@
     h2.app-card_question
       | {{'C1_HOUSEHOLD_INTRO.TITLE' | translate}}
 
-  alert-box(short-form="true" custom-message="householdEligibilityErrorMessage")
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/short-form/templates/c1-household-intro.html.slim
```diff
@@ -3,7 +3,7 @@
     h2.app-card_question
       | {{'C1_HOUSEHOLD_INTRO.TITLE' | translate}}
 
-  alert-box(short-form="true" custom-message="householdEligibilityErrorMessage")
+  alert-box(invert="true" custom-message="householdEligibilityErrorMessage" hide-alert="hideAlert")
```
chrisdolendo: @kaptron what is the significance of `invert` for this?

chrisdolendo:
aha! merging without :+1:-ing!

app/assets/javascripts/short-form/templates/c1-household-intro.html.slim
kaptron: @chrisdolendo invert gives it the styling that we want it to use where the background is dark red. (otherwise it is light red background with dark text). 

`"short-form=true"` was accomplishing that but also creating a bug where the alertbox was triggering twice

PR #366, opened 2016-09-01
Features/store nrhp qualification #128517263

kaptron:
this actually includes both related pivotal stories: #128517263, #128517425


chrisdolendo:
i paired on this with you so :+1:

PR #367, opened 2016-09-01
[#127767731] - Display middle name in review page

chrisdolendo:


app/assets/javascripts/short-form/templates/partials/_review-application-details.html.slim
```diff
@@ -3,7 +3,7 @@ review-summary-section header="{{'T.YOU' | translate}}" to="dahlia.short-form-ap
 
 .app-inner
   review-summary-item label="{{'LABEL.NAME' | translate}}"
-    | {{applicant.firstName}} {{applicant.lastName}}
+    | {{applicant.firstName}} {{applicant.middleName}} {{applicant.lastName}}
```
kaptron: @chrisdolendo I thought this would leave an extra space if there was no middlename but checked it out and it looks right, since it's HTML it treats multiple spaces as just one space. tricky!

PR #369, opened 2016-09-02
Chores/upgrade rails 4.2.7 #126731123

kaptron:
pretty straightforward, also upgraded `angular-rails-templates` because it used to not support sprockets > 3.0 -- if the heroku app works without any issues then we should be all good.


Gemfile
```diff
@@ -26,16 +24,15 @@ gem 'bower-rails'
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', '~> 0.4.0', group: :doc
 # Adds HTML templates into Angular's $templateCache using asset pipeline.
-gem 'angular-rails-templates'
-# NOTE: angular-rails-templates not yet compatible with sprockets > 3.0
+# angular-rails-templates > 1.0.0 now compatible with sprockets > 3.0
 # https://github.com/pitr/angular-rails-templates/issues/93
-gem 'sprockets', '~> 2.12.4'
```
chrisdolendo: curious - why remove sprockets?

kaptron: @chrisdolendo it's not technically "removing" sprockets -- sprockets is a dependency of rails so you don't normally need to specify it in the gemfile. But we had listed it in the gemfile in order to specifically tell it to use `~> 2.12.4`, however with the upgrade to `angular-rails-templates` we can just allow it to be included by default and in the `Gemfile.lock` it's using sprockets 3.7.0 (which is now compatible)

chrisdolendo: coolio!

chrisdolendo:
:+1:

PR #372, opened 2016-09-02
Features/show-apply-online-using-salesforce-field-#128035689

kaptron:
- hide "Apply Online" button unless listing.Accepting_Online_Applications == true
- also did minor refactor of `ListingService.listingIsOpen` to be passed a `listing` instead of just a date because it made more sense. I thought I was going to have to use this function but realized it was a much simpler change, literally just a couple of lines in the view to determine if we show the button or not.


kaptron:
actually the refactor made sense! I added a function to kick people out of the short form routes if they are trying to load a Listing that is not currently accepting online applications.

PR #377, opened 2016-09-07
Chores/update missing short form tests #126468679

chrisdolendo:


spec/requests/api/v1/listings_spec.rb
```diff
@@ -130,24 +130,24 @@
     expect(json['ami'].length).to eq(9)
   end
 
-  it 'gets Lottery Preferences' do
-    VCR.use_cassette('listings/lottery-preferences') do
-      get '/api/v1/listings/lottery-preferences.json'
-    end
+  # it 'gets Lottery Preferences' do
```
chrisdolendo: @kaptron this doesn't return anything anymore by salesforce and is blank on the listing page -- should I report this as a bug on Salesforce Pivotal or is this just stale code? I'm guessing the former

spec/requests/api/v1/short_form_spec.rb
```diff
@@ -38,10 +40,71 @@
...
+      url = '/api/v1/short-form/listing-application/a0Wf0000003j03WEAQ.json'
+      VCR.use_cassette('shortform/show_application') do
+        get url, {}, @auth_headers
+      end
+      expect(response).to be_success
+    end
+  end
+
+  # describe 'submit_application' do
+  # RSPEC IS CONVERTING THE BOOLEAN TO STRINGS, SALESFORCE REJECTING PARAMS
```
chrisdolendo: Running across this issue when I try to run this test: https://github.com/rspec/rspec-rails/issues/610

converts the boolean from the params to strings. WAI.

kaptron: hm it does make sense that the params are converted to strings, anything you post via a request will be strings. I can take a look at it and see what you mean.

spec/requests/api/v1/listings_spec.rb
kaptron: @chrisdolendo correct, this was an old endpoint that was never really properly implemented -- in fact that's one of the big changes coming down the pipe from Luke is figuring out how to query for Listing Lottery Preferences, so he said this salesforce endpoint was going to be deprecated.

kaptron:
@chrisdolendo I couldn't figure out the VCR bool -> string issue exactly so I just setup the JSON without any booleans, created the cassette with that, and at least the test will pass. I added a note about it (basically if we want to re-create that cassette later) but at least now the test can run and pass.

PR #380, opened 2016-09-08
[#127320539] - changed header text to say "BETA" and updated google f…

kaptron:
…eedback form link


chrisdolendo:
NO GOOD! jk :+1:

PR #381, opened 2016-09-08
Features/visual design for lottery preferences #129869597

kaptron:


chrisdolendo:
thanks for the comment to remove hardcoded features! 👍

PR #383, opened 2016-09-11
[#129914019] - Warn user when clicking back button from confirmation …

chrisdolendo:
…screen

This is the simplest implementation i could do


kaptron:
@chrisdolendo cool, looks like a good way to do it. 👍

PR #384, opened 2016-09-12
[#128994783] - fixed checkbox behavior by adding more logic to the ng…

kaptron:
…-change functions in the b2-contact form


chrisdolendo:
:+1:

PR #386, opened 2016-09-13
Features/preference bucket updates #127192635

kaptron:


app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -39,6 +39,7 @@ ListingController = (
   $scope.lotterySearchNumber = ''
   $scope.smallDisplayClass = "small-display-none"
   $scope.lotteryRankingSubmitted = false
+  $scope.loadingLotteryResults = false
```
chrisdolendo: I'm wondering if we should do the ng-disabled stuff for forms/buttons to be just a general `processingSubmit` type thing as opposed to creating unique one for each type of form. thoughts?

chrisdolendo:
looks good just had a Q! 👍

app/assets/javascripts/listings/ListingController.js.coffee
kaptron: that's a good point, and I'd agree... I don't think we're using different disabled/loading variables in very many places so it should be an easy refactor. but let's capture that separately, I've found that some bugs that they're finding in QA have come from some refactoring that caused some unexpected glitches that then need to get fixed (I'm thinking of one of the things I did w/ Account userAuth clearing messages)

PR #392, opened 2016-09-15
Chores/implement gtm ids #128993567

menslow:
Added first group of GTM tracking IDs.


menslow:
Thanks @chrisdolendo!

PR #395, opened 2016-09-15
[#130247761] - added special case for capturing going to sign-in from…

kaptron:
… create-account after you have completed your application


chrisdolendo:
already commented on other branch with my Qs so :+1:

PR #396, opened 2016-09-15
Bugs/account settings not being saved #130252597

kaptron:
includes #395, merge after that one is merged.


app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -77,13 +77,17 @@
     ##########################
     # < Account/Login pages >
     ##########################
+    # TODO: refactor "my account" pages to be under the same namespace/controller
```
chrisdolendo: :+1:  good catch

app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -152,6 +152,9 @@ AccountService = ($state, $auth, $modal, $http, $translate, ShortFormApplication
         contact: Service.userDataForSalesforce()
       $http.put('/api/v1/account/update', params).success((data) ->
         Service.accountSuccess.messages.nameDOB = $translate.instant("ACCOUNT_SETTINGS.ACCOUNT_CHANGES_SAVED")
+        data.contact.signedIn = true
```
chrisdolendo: re `signedIn` why indicate this manually?

app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -96,6 +100,8 @@
           templateUrl: 'account/templates/create-account.html'
           controller: 'AccountController'
       onEnter: ['AccountService', (AccountService) ->
+        AccountService.clearAccountMessages()
```
chrisdolendo: I wonder for the places where it says

```
AccountService.clearAccountMessages()
AccountService.resetUserAuth() 
````

if it makes sense to do at stateChangeStart, and use the toState.name to map through these routes, and if it matches, we call these so there is less redundancy.

File: spec/javascripts/helpers/ajax_helpers.js [hunk 4]
```diff
@@ -37,6 +37,7 @@ var stubAjaxRequest = function(requestURL, fixturePath) {
 var stubAngularAjaxRequest = function(httpBackend, requestURL, responseData) {
   httpBackend.when('GET', requestURL).respond(responseData);
   httpBackend.when('POST', requestURL).respond(responseData);
+  httpBackend.when('PUT', requestURL).respond(responseData);
```
chrisdolendo: nice

chrisdolendo:
:+1: looks good just had 2 Qs

app/assets/javascripts/config/angularRoutes.js.coffee
kaptron: @chrisdolendo yeah, unfortunately... we had already tried to refactor this into a stateChange callback if you remember (which seemed so smart), and I realized when doing this bugfix that our refactor had broken things. The order of operations is: `stateChangeStart`, then `resolve/onEnter`, then `stateChangeSuccess` ... the problem was that we can't `clearUserAuth` on `stateChangeSuccess` (otherwise it will wipe out other things that happen in `onEnter/resolve` functions in the routes) and we can't do it `stateChangeStart` otherwise it was too early -- can't remember the exact use cases but trust me I tried all different combinations for a while, and ended up putting things back into the routes rather than go down the 🐰  hole. 😞

app/assets/javascripts/account/AccountService.js.coffee
kaptron: @chrisdolendo good question -- the point was to copy over the contact into the loggedInUser while also preserving `signedIn = true` however I'm realizing this can be done better w/ a merge instead of a copy. Will update that, good catch!

PR #397, opened 2016-09-16
[#129640405] Added global rule for input to hide ms-clear

menslow:
Removed the ms-clear "x" in order to eliminate an IE browser bug that disables form submit after user clicks the "x" to clear the form field.


menslow:
You have to be viewing with IE 11 on Win 7.

Default IE browser behavior is to display an "X" once a form input has data in it. You can click the "X" to clear the data from the form. Unfortunately, due to a bug in IE, this behavior disables the ability to submit the form by clicking [Enter].

Through CSS you can hide the "X" and therefore have the input work like every other browser. That what I've done here.

PR #398, opened 2016-09-16
Features/print text truncate #130531829

chrisdolendo:


chrisdolendo:
@kaptron see also styling on PL

kaptron:
👍

PR #401, opened 2016-09-19
[#128596841] - copy applicant address to household member if you sele…

kaptron:
…ct "same address as applicant"


chrisdolendo:
thought this was already a thing haha :+1:

PR #404, opened 2016-09-20
[#126801511] - Applicant can review submitted application after submi…

chrisdolendo:
…scion

Link on the confirmation page, accessible whether or not they are signed in


chrisdolendo:
@kaptron updated the code to what we discussed with Jesse -- have a look see!

chrisdolendo:
@kaptron, this PR says you requested changes but wondering if you can approve so I can squash and merge? just don't wanna diss your request even if you gave the go on slack haha :dancer:

kaptron:
@chrisdolendo I only looked at the code but I'd like to actually click through it in the app as well, will do this afternoon!

chrisdolendo:
@kaptron ok sounds good!

kaptron:
@chrisdolendo approved! 👍  I just added a couple of tweaks to handle the state when you go through that flow as a logged in user, in that case I just have it take you to the `short-form-review` page since it's basically like taking them to My Applications which is their other option from the confirmation page.

PR #408, opened 2016-09-21
[#128699025] - streamlined email content by placing the standard gree…

kaptron:
…ting and signature in the layout so that it can be shared by all emails


chrisdolendo:
dope! :+1:

PR #412, opened 2016-09-22
[#129899809] Updated CSS to fix Short Form Navigation spacing

menslow:
Moved Jesse's short form nav spacing CSS updates over to Web App


kaptron:
👍

PR #414, opened 2016-09-22
Features/print listings #128425521

menslow:
Added "hide-for-print" class to button elements that are not needed for print views.


menslow:
"Reject" the story and we'll include it in this one.

chrisdolendo:
:+1:

PR #417, opened 2016-09-23
Features/combined live work preference #130413459

kaptron:


coffeelint.json
```diff
@@ -22,7 +22,7 @@
   },
   "cyclomatic_complexity": {
     "value": 65,
-    "level": "warn"
+    "level": "ignore"
```
kaptron: ^^ if the warning is not useful and we keep ignoring it anyway...

chrisdolendo:
dope!

PR #422, opened 2016-09-26
[129965669] - Updated leasing agent address on listing and submission…

chrisdolendo:
… confirmation


chrisdolendo:
@kaptron are you able to create review apps ok? looks like it's broken, and i recreated the PR.

PR #425, opened 2016-09-27
Bugs/firefox print #131009057

kaptron:


kaptron:
one line CSS bugfix for firefox -- will merge to QA

PR #426, opened 2016-09-27
[#130757945] - disable contact form inputs if you indicate you do not…

kaptron:
… have email/phone/address


app/assets/javascripts/short-form/templates/b2-contact.html.slim
```diff
@@ -4,16 +4,16 @@
       | {{'B2_CONTACT.TITLE' | translate:applicantFirstName()}}
 
   .app-inner.inset.border-bottom
-    .form-group.has-option ng-class="{ error: $scope.requiredContactInformationMissing() }"
+    .form-group.has-option ng-class="{ error: inputInvalid('phone') }"
```
chrisdolendo: how were you able to trigger inputInvalid for phone without calling the function to see if there was missing required contact info?

chrisdolendo:
works great -- just had that question.

app/assets/javascripts/short-form/templates/b2-contact.html.slim
kaptron: 🙌 <-- magic fingers. Really though, I realized that everything came together nicely by just using the `inputInvalid` and just letting the alertBox show up under the right conditions (e.g. the form is submitted and invalid _and_ `requiredContactInformationMissing`). The reason the `ng-class=error` works is because we are saying "this field is required IF required contact info is missing" -- therefore it's the `ng-required` that kicks in to trigger the `inputInvalid` by saying "due to the missing info conditions, this form is now required"

chrisdolendo: right on! I see the ng-required function. Cool! I like the idea of just running that function and marking the input invalid so we only need to run a one function for all the error classes.

PR #428, opened 2016-09-27
[#130253435] - added a catch so that you don't get the confirmation m…

kaptron:
…odal unless you enter the correct pw on sign in


chrisdolendo:
:+1:

PR #429, opened 2016-09-28
[#129965669] - Removed extra line break on submission confirmation email

chrisdolendo:


chrisdolendo:
this is just a quick request from Ashley to related to updating leasing agent address to fix the line breaks on the confirmation email

PR #431, opened 2016-09-28
[#130416355] - Update privacy policy

chrisdolendo:


kaptron:
looks like it was a little tedious but can't think of any other way around it (without rethinking how we allow them to enter large blocks of text... I think doing this in a very structured way like this makes sense though). Thanks for doing this!

PR #434, opened 2016-09-29
[#120686801] - using new listing.Office_Hours field 

kaptron:
instead of the old open/closed fields...

WIP - was discussing w/ @slowbot how to best give salesforce users the ability to format the office hours without having to use anything beyond `p` and `strong` tags, right now they have to wrap each combo of days/hours in a `p.content-group_days.inline`


kaptron:
no longer WIP - this is complete.

PR #435, opened 2016-09-29
Features/cta buttons in modal #127929025

kaptron:


kaptron:
included one Gemfile config change I ran into when upgrading macOS, which then led to some weird semaphore errors w/ jasmine + phantomjs -- everything seems fine now. As a result I did actually figure out a trick to make our semaphore builds go faster, because they were always trying to redownload/install phantomjs every time (the download happened to timeout, which was the error). Adding the line `change-phantomjs-version 2.1.1` to our semaphore project setup makes the build env already have phantomjs 2.1.1 available to use without needing to download it over again.

PR #440, opened 2016-09-30
[#130251635] - updated eligibility estimator logic for childrenUnder6

kaptron:
only show children values that are < the total in the household; also fixed one issue with displaying notMatched listings when you have 0 matches


chrisdolendo:
:+1:

PR #460, opened 2016-10-12
[#131402097] - radio-block-item needs to use on-change instead of ng-…

kaptron:
…click, otherwise was getting weird behavior for keyboard navigation


chrisdolendo:
:+1:

PR #466, opened 2016-10-14
Features/install gtm tracking on create account #131309683

kaptron:
still WIP + requires further discussion with the team for proper GTM setup. Also for now the `angulartics.debug` module is enabled, which logs any `$analytics` calls to the console.


chrisdolendo:
is this still WIP?

kaptron:
@chrisdolendo the build is "done" in that it should be working and tracking the Create Account form, however we still need to go over it with Lukas still to finalize how we want the GTM calls to be made before considering this complete.

kaptron:
no longer WIP!

kaptron:
re: `angulartics-addons/angulartics-gtm.js.coffee` -- I'm realizing the proper way to do this would probably be to fork the original project and add this to `bower.json` rather than stick this inside of the application. Will redo that tomorrow.

PR #467, opened 2016-10-15
Features/updates demographics #127692787

menslow:
Updates to Short form Demographics Page:

- Removed "optional" labels from individual fields
- Removed "Decline to state" from dropdowns
- Made "please specify" not required for sexual orientation.


chrisdolendo:
looks good -- do we need to have salesforce update these pick lists? decline to state was an option.

menslow:
@chrisdolendo Our change wont break anything because we just removed one of the options. However, Ashley is checking into whether or not we should remove the Salesforce one to keep things in synch.

chrisdolendo:
yup i don't expect it to break, but yes it would be good to keep things in sync

PR #469, opened 2016-10-17
[#127692035] - Line break on confirmation page heading

chrisdolendo:


kaptron:
👍

kaptron:
@chrisdolendo looks right visually, however I think the `p` should be a `br`? I don't think we should nest `p` inside `h2`

chrisdolendo:
@kaptron ha, yeah i was choosing between p and br. just updated.

PR #475, opened 2016-10-18
[#126712861] - Shortform - live/work preference dropdown shown if one…

chrisdolendo:
… eligible, with household members

### DIFF
--- a/app/assets/javascripts/short-form/directives/checkbox-preference-item.html.slim
+++ b/app/assets/javascripts/short-form/directives/checkbox-preference-item.html.slim
@@ -1,6 +1,6 @@
 div
   p.checkbox-block
-    input id="preferences-{{checkbox_pref_type}}" name="preferences" data-event="{{pref_data_event}}" ng-model="application.preferences[checkbox_pref_type]" type="checkbox" aria-labelledby="{{labelledby}}" ng-click='reset_preference_data(checkbox_pref_type)'
+    input id="preferences-{{checkbox_pref_type}}" name="preferences" data-event="{{pref_data_event}}" ng-model="application.preferences[checkbox_pref_type]" type="checkbox" aria-labelledby="{{labelledby}}" ng-click="reset_preference_data(checkbox_pref_type);refresh_member_dropdown()"
     label.checkbox-block_label for="preferences-{{checkbox_pref_type}}"
       strong.form-label.caps
         | {{title}}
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/short-form/directives/checkbox-preference-item.html.slim
```diff
@@ -1,6 +1,6 @@
 div
   p.checkbox-block
-    input id="preferences-{{checkbox_pref_type}}" name="preferences" data-event="{{pref_data_event}}" ng-model="application.preferences[checkbox_pref_type]" type="checkbox" aria-labelledby="{{labelledby}}" ng-click='reset_preference_data(checkbox_pref_type)'
+    input id="preferences-{{checkbox_pref_type}}" name="preferences" data-event="{{pref_data_event}}" ng-model="application.preferences[checkbox_pref_type]" type="checkbox" aria-labelledby="{{labelledby}}" ng-click="reset_preference_data(checkbox_pref_type);refresh_member_dropdown()"
     label.checkbox-block_label for="preferences-{{checkbox_pref_type}}"
```
kaptron: yes I'm remembering this directive getting sort of beastly especially after the introduction of the "combined Live OR Work" preference. I'm wondering if this `ng-click` should be refactored to `ng-click="onPreferenceChange(checkbox_pref_type)"` and inside _that_ method it calls `reset_preference_data` and `refresh_member_dropdown`. 

Likewise the `refresh_member_dropdown` could be called within `select_liveWork_type` which is really the onSelect for that input (could be renamed if it needs to be clearer, I also realize we sometimes mix ruby + js naming conventions together, I am definitely at fault for a lot of it 😬 )

chrisdolendo: yeah i was gonna go and put it within `select_liveWork` but it's really doing something else and didn't want that dependency.

I do like `onPreferenceSelect` that calls both though.. that's more semantic-- let me refactor.

chrisdolendo:
@kaptron can you re-check? i think this is good for the story before we do a further refactor tomorrow!

kaptron:
👍

kaptron:
that was a hard-earned thumbs up 😁

chrisdolendo:
@kaptron no biggie it was super fun! can't wait to further refactor.

PR #483, opened 2016-10-25
Bugs/short form keyboard navigation #131204189

kaptron:


app/assets/javascripts/shared/NavController.js.coffee
```diff
@@ -2,12 +2,28 @@
...
 ############################################################################################
 
-NavController = ($document, $rootScope, $scope, $state, AccountService) ->
+NavController = ($document, $rootScope, $scope, $state, $timeout, AccountService) ->
   $scope.loggedIn = AccountService.loggedIn
 
   # Utility function to scroll to top of page when state changes.
   $rootScope.$on '$stateChangeSuccess', ->
     $document.scrollTop(0)
+    $timeout ->
```
chrisdolendo: why $timeout?

chrisdolendo:
:+1: just had that one Q!

app/assets/javascripts/shared/NavController.js.coffee
kaptron: something about the timing of the content actually being rendered... without the timeout then it doesn't seem to be able to locate certain elements on the page (i.e. they don't exist yet). I changed this event to `$on '$viewContentLoaded'` which seems more appropriate but it still didn't behave 100% correct w/out the timeout.

PR #487, opened 2016-10-26
Bugs/date on file upload #132619695

chrisdolendo:
can't replicate consistently locally.. did some stuff where it didn't happen again. testing on review app


kaptron:
👍  nice catch on removing the unused code from ShortFormApplicationService (which got moved to FileUploadService), I removed some more unnecessary references to `Upload`.

PR #493, opened 2016-10-27
[#132685371] - refactored alertBox to not use angular mm.foundation d…

kaptron:
…irective and directly include the close button; set up the button with aria-label and respond to keyboard input


chrisdolendo:
:+1:

PR #496, opened 2016-10-28
[#130337033] - updated d2_live_work instruction text

kaptron:
also updated the corresponding string in POEditor so that it will match up when we merge.


chrisdolendo:
:+1:

PR #520, opened 2016-11-03
Bug - upload file validation (#519)

chrisdolendo:


chrisdolendo:
@kaptron this is what I pushed to staging

PR #521, opened 2016-11-03
[#133637863] Remove blur function after focusing top element

creativecoder:
Maintain focus for the top element of the page so that
- Keyboard users have a visual indication of focus
- Validations are not triggered for form fields


kaptron:
@creativecoder -- according to @menslow's suggestion we should just remove this focus/blur section (the whole thing inside the `$timeout`) within the `$on $viewContentLoaded` callback as the fix, basically rolling back this feature.

creativecoder:
@kaptron Commented out top focus functionality, as requested. Please ignore this until we resolve more pressing issues.

PR #523, opened 2016-11-04
Bugs/review invalid entry #132135085

menslow:
* Added aria-invalid attribute to all required form fields of web app in order to prevent screen readers from assuming the fields is invalid by default.

### DIFF
--- a/app/assets/javascripts/account/AccountController.js.coffee
+++ b/app/assets/javascripts/account/AccountController.js.coffee
@@ -58,6 +58,8 @@ AccountController = (
     field = form[fieldName]
     if form && field
       field.$invalid && (field.$touched || form.$submitted)
+    else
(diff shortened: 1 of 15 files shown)


kaptron:
👍  - did not do any screen reader testing but markup looks good to me

chrisdolendo:
bumping this as this PR is pretty old.. ready to merge?

kaptron:
@chrisdolendo yeah it had been stale for a little while, but I finally just fixed + updated it recently. next step is @slowbot is planning to check it out w/ screen reader users and check it out. 

However thanks for flagging this, I think one approach might be to consider this feature "complete" -- and if the user tests result in any other findings, then that might just inform us of new user stories rather than keeping this one alive forever. Agree?

PR #528, opened 2016-11-08
[#133934987] - bugfix for re-initialization of liveWork combo checkbox

kaptron:


chrisdolendo:
:+1: should we have an integration test for this?

kaptron:
@chrisdolendo like add an E2E test? yeah would be cool to do so...

chrisdolendo:
yes I was thinking for bug stories, we can use our discernment on whether to do an E2E test.. i feel like this is one that would be good to have. let me know if you'd like to pair

app/assets/javascripts/short-form/directives/preference-proof-uploader.html.slim
```diff
@@ -11,7 +11,7 @@
     | {{'D2_LIVE_WORK_PREFERENCE.LETTER_FROM_EMPLOYER_INSTRUCTIONS' | translate}}
 
 .form-group.text-center ng-if="show_preference_uploader()"
-  button.button.ghost-white.expand ngf-select="uploadProof($file, preference, application.preferences[preference_proof_option])" ngf-accept="'image/jpg, image/jpeg, image/png, application/pdf'" name="file_{{preference_proof_file}}" ngf-max-size="2MB" ng-required="application.preferences[preference]" aria-describedby="{{preference_proof_file}}_error" ng-model="application.preferences[pref_type_proof_file]"
+  button.button.ghost-white.expand ngf-select="uploadProof($file, preference, application.preferences[preference_proof_option])" ngf-accept="'image/jpg, image/jpeg, image/png, application/pdf'" name="file_{{preference_proof_file}}" ngf-max-size="2MB" ng-required="application.preferences[preference]" aria-describedby="{{preference_proof_file}}_error" ng-model="application.preferences[preference_proof_file]"
```
kaptron: @chrisdolendo note #1: remember we had to re-add `ng-model` here? well `pref_type_proof_file` didn't seem to be anything any more, so my hunch is that correcting this might fix our "uploaded file is sometimes blank" issue though still unclear. But this at least has the correct var name now.

app/assets/javascripts/short-form/templates/d1-preferences-programs.html.slim
```diff
@@ -10,7 +10,7 @@
       .checkbox-group role="group"
         preference-without-proof title="{{'D1_PREFERENCES_PROGRAMS.CERT_OF_PREFERENCE' | translate}}" preference="certOfPreference" labelledby="special_programs_label" dataevent="gtm-preference-cop" translated-description="D1_PREFERENCES_PROGRAMS.CERT_OF_PREFERENCE_DESC"
 
-        preference-without-proof title="{{'D1_PREFERENCES_PROGRAMS.DISPLACED' | translate}}" preference="displaced" labelledby="special_programs_label" dataevent="gtm-preference-dthp" translated-description="D1_PREFERENCES_PROGRAMS.DISPLACED_TENANTS_DESC" ng-if="!listingIs('280 Brighton')"
+        preference-without-proof title="{{'D1_PREFERENCES_PROGRAMS.DISPLACED' | translate}}" preference="displaced" labelledby="special_programs_label" dataevent="gtm-preference-dthp" translated-description="D1_PREFERENCES_PROGRAMS.DISPLACED_TENANTS_DESC" ng-show="!listingIs('280 Brighton')"
```
kaptron: @chrisdolendo note #2: another place where the `ng-if` was F'ing things up (I think we had seen it before w/ these preference directives). Discovered on a test listing that _did_ have DTHP and the ng-if was giving us some `undefined`'s

app/assets/javascripts/short-form/directives/preference-proof-uploader.html.slim
chrisdolendo: oh right on. fingers crossed. thanks for pointing this out to me

app/assets/javascripts/short-form/templates/d1-preferences-programs.html.slim
chrisdolendo: gotcha. we'll use `ng-show` for now moving forward

kaptron: yeah it's only because of something to do with all the variables and things going on inside that directive. if it was `ng-if` for a little piece of content it wouldn't matter.

PR #545, opened 2016-11-17
[#132350717] - upgraded ng-file-upload package to fix the issues with…

kaptron:
…… (#544)

* [#132350717] - upgraded ng-file-upload package to fix the issues with proof upload (filename or date not showing, not triggering validation properly)

* [#132350717] - added better handling of case where you switch between live/work/liveWork

* [#132350717] - removed "button" element from uploader

* [#132350717] - turning off the modal instance bugfix since it was causing issues with the sign in confirmation modal

* [#132350717] - updated "button" element in uploader

* [#132350717] - refactored scope.checkForLiveOrWorkValues into its own function

* [#132350717] - added one last catch for odd file bug

* [#132350717] - added error catch for GeocodingService getting an error (500)


chrisdolendo:
:+1:

PR #548, opened 2016-11-21
[#130778859] - Short Form - Dynamically display preference availability

chrisdolendo:


app/assets/javascripts/short-form/templates/d2-live-work-preference.html.slim
```diff
@@ -9,7 +9,8 @@
       label.sr-only id="additional_preferences_label" translate="LABEL.ADDITIONAL_PREFERENCES"
       .checkbox-group role="group"
 
-        preference-with-proof(title="{{'D2_LIVE_WORK_PREFERENCE.LIVE_SF_PREFERENCE.TITLE' | translate}}" preference="liveInSf" labelledby="additional_preferences_label" dataevent="gtm-preference-livesf" translated-description="D2_LIVE_WORK_PREFERENCE.LIVE_SF_PREFERENCE.DESCRIPTION" ng-show="showPreference('liveInSf')")
+        preference-with-proof(title="{{'D2_LIVE_WORK_PREFERENCE.LIVE_SF_PREFERENCE.TITLE' | translate}}" preference="liveInSf" labelledby="additional_preferences_label" dataevent="gtm-preference-livesf" translated-description="D2_LIVE_WORK_PREFERENCE.LIVE_SF_PREFERENCE.DESCnce
+        RIPTION" ng-show="showPreference('liveInSf')")
```
kaptron: @chrisdolendo looks like something got messed up on this line

app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -347,6 +365,9 @@ ListingService = ($http, $localStorage, $modal, $q, $state) ->
   Service.listingIs = (name) ->
     Service.LISTING_MAP[Service.listing.Id] == name
 
+  Service._returnPreferenceName = (preference) ->
```
kaptron: @chrisdolendo does really this need to be pulled out into its own function, if it's really just for:

```coffee
preferenceNames = _.map Service.listing.preferences, (pref) -> pref.preferenceName
```

app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -81,6 +81,24 @@ ListingService = ($http, $localStorage, $modal, $q, $state) ->
       minMax = [min, max]
     return minMax
 
+  Service.hasPreference = (preference) ->
+    hasPreference = null
```
kaptron: not that it will have any real difference in outcome, but this should probably default to `false`, that way `hasPreference` either returns `true` or `false`, rather than `true` or `null`.

kaptron:
@chrisdolendo I was leaving a few comments but then I realized just making a very minor PR would be easier way to explain what I was going for, see 5bfeb6b and 559d51e

Anyway it works well! I tested out adding DTHP or Live/Work to "Test Listing". I saw they also created preferences in Salesforce QA and it looks like "Live or Work in San Francisco Preference" is the naming convention so I updated that in `ListingService`.

kaptron:
note I removed the Brighton hardcoding because by the time this feature goes live, that listing will no longer be accepting applications and we should just rely on `showPreference('displaced')` to know whether to show/hide that pref going forward.

chrisdolendo:
@kaptron yes thank you! I liked your refactor. I kinda made this PR prematurely, need to add some tests.. definitely unit tests, but maybe an integration test would be good too. I also haven't tested this thoroughly so will probably do that before delivering.

PR #554, opened 2016-11-23
[#134872227] - Check if sessionStorage is available so the browser do…

kaptron:
…esn't freak out and crash in the event that it's not (e.g. Safari Private Window)


chrisdolendo:
:+1:

PR #558, opened 2016-11-29
[#135145517] Resolve angular-translate version in bower.json

creativecoder:
`bower install` triggered a manual choice for angular-translate version, since other dependencies in bower.json require to different versions of this library. This PR sets bower.json to automatically resolve to use v2.13.0, currently the most recent version of angular translate.


kaptron:
@creativecoder good call, and I'm thinking we should freeze all the versions rather than using fuzzy ones. We did add a `.bowerrc` with `save_exact: true` so new packages are no longer added w/ the `~` but I think we should still clean up and freeze the versions for _all_ the packages. I'm no bower power user, but it seems like some of those `resolution` declarations might still be necessary? Or would it be solved if all of the dependencies had exact versions?

creativecoder:
@kaptron The resolutions section tells bower what to do in the case where the dependencies in bower.json have the same sub-dependency, but specify different versions of that sub-dependency.

Here's what I saw when I did bower install:
```
Unable to find a suitable version for angular-translate, please choose one by typing one of the numbers below:
    1) angular-translate#~2.11.1 which resolved to 2.11.1 and is required by angular-translate-loader-static-files#2.11.1
    2) angular-translate#2.13.0 which resolved to 2.13.0 and is required by sf-dahlia-web
    3) angular-translate#~2.13.0 which resolved to 2.13.0 and is required by angular-translate-loader-static-files#2.13.0

Prefix the choice with ! to persist it to bower.json
```

I responded with `!2`, which is what added the change to bower.json

creativecoder:
I'm not sure why two different versions of `angular-translate-loader-static-files` are referenced. Bower isn't great at handling these kinds of issues (compared to npm).

kaptron:
@creativecoder cool, thanks for the clarification. I just cleaned up the bower file to lock it at all the versions that I had locally (found this nifty trick on SO: `grep "version\"\:" lib/assets/bower_components/*/.bower.json`). I did a `rm -rf` on my bower folder and a clean `bower i` and now it doesn't ask me to resolve any dependencies. However it does mention that various packages seem to depend on slightly different versions of angular (i.e. some of them point to angular >1.5), but I don't think there's much we can do to resolve that without fiddling with all of the package versions (and everything should work exactly as it always has been, anyway):

```
Please note that,
    angular-pageslide-directive#1.0.13 depends on angular#1.4.6 which resolved to angular#1.4.6
    angular-clipboard#1.1.2, angular-cookie#4.1.0, angular-filter#0.5.11 depends on angular#* which resolved to angular#1.4.14
    angular-mocks#1.4.14, sf-dahlia-web depends on angular#1.4.14 which resolved to angular#1.4.14
    ng-idle#1.2.2 depends on angular#>= 1.2.0 which resolved to angular#1.4.14
    angular-scroll#1.0.0 depends on angular#^1.2.16 which resolved to angular#1.4.14
    angular-translate#2.13.0 depends on angular#>=1.2.26 <1.6 which resolved to angular#1.4.14
    ng-token-auth#0.0.29 depends on angular#>=1.4 <=1.5.3 which resolved to angular#1.4.14
    angular-ui-router#0.2.18 depends on angular#^1.0.8 which resolved to angular#1.4.14
    angular-ui-validate#1.2.2 depends on angular#>= 1.3.0 which resolved to angular#1.4.14
    angular-ui-mask#1.8.7 depends on angular#>= 1.3.6 which resolved to angular#1.4.14
    lifely-angular-carousel#1.1.0 depends on angular#>= 1.2.3 which resolved to angular#1.4.14
    ng-currency#0.9.6 depends on angular#>=1.3.4 which resolved to angular#1.4.14
    ng-file-upload#12.2.13 depends on angular#>1.2.0 which resolved to angular#1.4.14
    ngstorage#0.3.11 depends on angular#>=1.0.8 which resolved to angular#1.4.14
    angular-foundation#0.7.0 depends on angular#>=1.3.0 which resolved to angular#1.4.14
    angular-aria#1.5.8, angular-sanitize#1.5.8 depends on angular#1.5.8 which resolved to angular#1.5.8
Resort to using angular#1.4.14 which resolved to angular#1.4.14
Code incompatibilities may occur.
```

PR #562, opened 2016-12-01
[#118884683] - added toggle functionality to the headers of the listi…

kaptron:
…ng buckets


kaptron:
was just one tweak to the front-end, Jesse signed off on this.

PR #570, opened 2016-12-07
[#132294081] - adds listing name and link to gis email notifications

catcall:


app/mailers/emailer.rb
```diff
@@ -70,6 +72,8 @@ def geocoding_error_notification(data, log)
     @error = data[:errors].first
     setup_geocoding_notification
 
+    @log[:listing_url] = "#{base_url}/listings/#{@log[:listing_id]}"
```
kaptron: @catcall minor detail, but I would change this to:
`@listing_url = "#{base_url}/listings/#{@log[:listing_id]}"`

and refer to it as listing_url within the template. Since `listing_url` is not really a field on `GeocodingLog` it doesn't feel appropriate to add it onto the log object.

kaptron:
hello slack notifications, I added a review.

PR #575, opened 2016-12-15
[#136061827] - removed old hardcoded elements for 280 Brighton and re…

kaptron:
…purposed sidebar hardcoding for L Seven


chrisdolendo:
:+1:

PR #581, opened 2016-12-16
[#136227357] - attach any temp session files to user upon registration

kaptron:
* [#136227357] - attach any temp session files to user upon registration

* [#136227357] - some additional fixes and catches for making sure the right proof files are attached to the appropriate user

### DIFF
--- a/app/assets/javascripts/short-form/FileUploadService.js.coffee
+++ b/app/assets/javascripts/short-form/FileUploadService.js.coffee
@@ -1,7 +1,8 @@
 FileUploadService = ($http, Upload, uuid, ShortFormApplicationService) ->
   Service = {}
-  Service.session_uid = ShortFormApplicationService.session_uid
   Service.preferences = ShortFormApplicationService.preferences
+  Service.session_uid = ->
+    ShortFormApplicationService.session_uid
 
   Service.hasPreferenceFile = (fileType) ->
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/short-form/FileUploadService.js.coffee
```diff
@@ -1,7 +1,8 @@
 FileUploadService = ($http, Upload, uuid, ShortFormApplicationService) ->
   Service = {}
-  Service.session_uid = ShortFormApplicationService.session_uid
   Service.preferences = ShortFormApplicationService.preferences
+  Service.session_uid = ->
```
chrisdolendo: is there a reason why it's a function as opposed to an assignment?

kaptron: @chrisdolendo yep! because I added an update so that `session_uid` can get refreshed every time you start a new short form app (stops you from accidentally picking up a file upload from a short form session in your same browser window, if you start a new application). So if the value is an assignment, then FileUploadService.session_uid will always be assigned to the initial value. If it's a function then it will run the function, and pick up whatever is the _current_ value in SFApp service.

chrisdolendo: gotcha!

PR #585, opened 2016-12-20
Features/mobile nav accessibility #135638981

kaptron:


kaptron:
@creativecoder what should it do in that case? it seems like if you do that, it just "blurs" the focus so that nothing is in focus, with the menu close button becoming the first tab item e.g. if you tab into it.

but yeah if that's a bug, I think that might be separate -- this story was just to resolve the focus of the mobile nav for accessibility (e.g. voiceover / keyboard), so that the menu took focus when you opened it (previously the focus remained on the page behind the menu).

creativecoder:
@kaptron You're right, with keyboard navigation, it's not an issue. With iOS voice navigation, tapping within the menu strangely moves the focus back to the page and starts reading elements on the page, behind the open menu (this happened to me during testing when I accidentally tapped instead of swiped when the menu was open). To get focus back on the menu without tapping a link on it, you then have to swipe back up the page to the menu icon and activate it, but because the menu is already open, the menu closes, instead. So, then you have to open the menu again, to actually get focus back the menu.

Maybe a good thing to do right now is remove the toggle function and instead have separate open and close functions, since those actions are handled by separate buttons anyway. What do you think?

kaptron:
@creativecoder I agree that having separate open/close vs. just one generic toggle (with logic to tell _if_ it's closing) makes more sense. updated -- I think I followed the same steps you were talking about, but let me know if it seems fixed now.

creativecoder:
This is kind of maddening... now it works like this (with voiceover):
1. Swipe and double tap to open mobile menu
1. Focus goes into mobile nav menu
1. "Accidental single tap" on bottom of menu
1. Focus goes back to page even though menu is open
1. Swipe back to open menu button and double tap
1. Voiceover announced "Open Menu" but focus stays on open menu button, behind menu. Menu stays open. (Just like the original problem, but you can't close the menu!)

kaptron:
@creativecoder check out the last commit adding `ng-blur=trapFocus()` to all the mobile nav links, like I said it doesn't feel the cleanest but I couldn't figure out a more universal way to do it. https://github.com/Exygy/sf-dahlia-web/pull/585/commits/feaec454139e82b0b52cb21ae438155712a6e6af

It does seem to work pretty well though both in tab-navigating and mobile + voiceover contexts. 

it would be nice to do something more globally e.g. as outlined here https://www.nczonline.net/blog/2013/02/12/making-an-accessible-dialog-box/
`document.addEventListener("focus", function(event) { ...`

but adding a 'focus' eventListener didn't seem to trigger anything, and I couldn't find any jQuery-ish equivalent to adding a global event listener like that. Let me know if you have any thoughts or if we should just stick with this as is...

PR #589, opened 2016-12-21
[#130994559] - Update heroku buildpack config

chrisdolendo:


chrisdolendo:
Got this from: https://devcenter.heroku.com/changelog-items/670

PR #592, opened 2017-01-04
Chores/e2e ci #134472139

chrisdolendo:
Semaphore is running the e2e tests headless - meaning it starts its own rails and selenium server and runs protractor from the semaphore environment. This way we don't need to wait for the PR review app to run the tests, which prevents us from running into a lot of obstacles such as waiting for the review app to finish deploying before running the tests. It also means we don't need to pay for and rely on saucelabs to run it for us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "grunt-contrib-copy": "0.8.1",
     "grunt-replace": "0.11.0",
     "grunt-sort-json": "0.0.5",
-    "protractor": "4.0.10"
+    "protractor": "4.0.10",
+    "npm-run-all": "4.0.0"
   },
   "dependencies": {
     "bower": "1.7.9",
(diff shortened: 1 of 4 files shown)


chrisdolendo:
READY TO GO!!!

package.json
```diff
@@ -9,18 +9,24 @@
...
   "engines": {
     "node": "4.1.x",
     "npm": "2.1.x"
   },
   "scripts": {
-    "postinstall": "bower install"
+    "postinstall": "bower install",
+    "webdriver-start": "./node_modules/protractor/bin/webdriver-manager start",
+    "protractor": "./node_modules/protractor/bin/protractor spec/e2e/conf.js",
+    "e2e": "npm-run-all -p -r webdriver-start protractor"
```
chrisdolendo: we run `npm run e2e` on semaphore which allows web driver to run in the background so that protractor can run.

chrisdolendo:
@creativecoder yes this is correct, updated the PR description and pushed up an update on the readme as well

kaptron:
👍  paired w/ CD on this one. Merging this so that other PR builds can now use semaphore + E2E as well!

PR #596, opened 2017-01-09
Chores/e2e ci #134472139

chrisdolendo:


chrisdolendo:
it's alive!

chrisdolendo:
@kaptron do you think we can merge this first and then update the cucumber app (we can do that together too)

kaptron:
bitmoji

PR #602, opened 2017-01-13
Features/singlelisting sidebar waitlist #131920617

chrisdolendo:


chrisdolendo:
@kaptron @creativecoder  for tests, I'm noticing that the fakeListing on the ListingService is an old one from Salesforce -- thoughts on updating the tests and updating fakeListing to one that has multi-family fields? Should we capture it as a chore or make it as part of this story

kaptron:
@chrisdolendo hm, I don't think we should point our tests to the salesforce-multifamily instance, which would pick up those new fields (it should always point at QA) but it does make sense that we may need to update the fake listing at some point -- seems like that would belong in its own chore though.

kaptron:
@chrisdolendo just added a couple of small updates to this... all of our "listings" related text in `locale-en.json` are in "LISTINGS" and not "LISTING" so I just moved those to be consistent. Also there is a 3rd case (in your review instr. you noted case 2 and 3 are the same, they are actually just _almost_ the same), the case where _all_ units are available is considered a "new building" and just has slightly different language. Actually it looks like you started to do it this way (where total == available) but then removed that part, haha... probably because it wasn't totally clear about why that case was different.

Anyway, with those updates: 👍

PR #605, opened 2017-01-19
[#136294971] - Bugfix - reassign primary applicant to preference on s…

chrisdolendo:
…ave draft


kaptron:
@chrisdolendo Looks great -- and also looks like a good example to add a test 😜  -- even just a small unit test that just checks when you run `importUserData` that it has changed the values as expected.

chrisdolendo:
@kaptron ha! you are right. fair enough.

chrisdolendo:
@kaptron just wrote the spec!

kaptron:
great! 👍

PR #613, opened 2017-01-23
[#136967891] - Show lottery results accordion when no open listings

chrisdolendo:


kaptron:
nice and easy solution! Shouldn't `Service.displayLotteryResultsListings` get initialized at the top of the Service (and I guess to `false`) just so that it's not undefined?

chrisdolendo:
nice @kaptron just updated to initialize it in the service as false :)

PR #614, opened 2017-01-24
Features/prompt sign in before submitting #128266733

kaptron:


app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
```diff
@@ -425,6 +425,49 @@ ShortFormApplicationService = (
       return
     )
 
+
+  # ----- TODO: call this function via AccountController and ShortFormApplicationController
```
chrisdolendo: should we delete this code?

chrisdolendo:
i remember we paired on this. looks good just had that one question about commented out code

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
kaptron: @chrisdolendo yes technically, it's being removed in the next feature story where the `TODO`'s get implemented (that's the one I still have in flight)

PR #616, opened 2017-01-25
Bugs/pre lottery missing start date time #138292105

kaptron:


kaptron:
note this one is only failing semaphore because anything off the `production` branch doesn't have e2e tests available.

PR #623, opened 2017-01-31
Features/mf browse listings units table v2 #132255853

kaptron:


app/assets/javascripts/listings/templates/_unit-summary.html.slim
```diff
@@ -1,26 +1,36 @@
-tr ng-repeat="unit_summary in summary"
+tr ng-repeat-start="unit_summary in summary" ng-if="unit_summary.minRentalMinIncome || unit_summary.maxRentalMinIncome"
```
chrisdolendo: cool! didn't know about `start` and `end`! super useful

kaptron: yeah had to look that up too 😁 , but it's good for this use case of having a loop, but wanting the loop to contain multiple things.

PR #629, opened 2017-02-03
[#135676979] Adds logic for displaying "Waitlist/Available" in browse grid

kaptron:
… using listing data


chrisdolendo:
rad yay for deleting code! :+1:

PR #630, opened 2017-02-06
[#114738921] - Autogenerate sitemap

chrisdolendo:


lib/tasks/sitemap.rake
```diff
@@ -0,0 +1,22 @@
+require 'sitemap_generator'
+
+desc 'This task is called by the Heroku scheduler add-on'
+
+task generate_sitemap: :environment do
+  listings = ListingService.listings(nil)
```
kaptron: this can just be `ListingService.listings()` don't need to pass the `nil`

kaptron:
looks great! a couple of modifications: 
 - Can remove `sitemap.xml`, and put it in the gitignore, since it will be created by the rake task
 - Update some of the priority/changefreq settings in the sitemap:
   - homepage doesn't need to be `always`, can be `weekly` (the homepage itself doesn't update much) 
   - /listings should be `daily` and have a higher priority, like 0.75
   - each /listings/ID should also be `daily`, and have a higher priority, like 0.75
   - the rest look good

creativecoder:
1. Agree with keeping sitemap files out of git.

2. I did a little research on priority/changfreq, since I didn't have an answer of the top of my head. Latest version of WordPress SEO actually removed those, b/c Google has said they don't use them, and updating the timestamp for specific pages to indicate an update, instead (https://www.seroundtable.com/google-priority-change-frequency-xml-sitemap-20273.html). So we could just remove them... or leave them and and use @kaptron's suggestions for settings. Google is not the only search engine, after all.

3. That brings up another question: how are we setting lastmod date? Seems like we would need to update that every time the sitemap is regenerated, since we don't have a good way to do it automatically when content actually gets updated.

creativecoder:
And one more thing: I don't see the translation pages--they should be part of the sitemap. When we do move to a multi-lingual site, we'll need to specify that in the sitemap using `hreflang` (https://support.google.com/webmasters/answer/189077)

chrisdolendo:
@kaptron thanks for the feedback, I removed sitemap.xml and updated the priority and changefreq

@creativecoder Thanks for the feedback on the priority/changefreq, I'll just keep them with Dave's recs, and we can remove them if it's not making a diff.

Last mod is automatically generated by the gem that generates the sitemap. So it will update overtime the heroku task runs. I've also added the translation pages to the task, but unfortunately the gem doesn't yet support `hreflang`. We can keep looking into it to continue optimizing how we generate this.

kaptron:
👍

PR #632, opened 2017-02-06
[#138308775] Removes all references to claim submitted application

kaptron:
also regenerated some short form VCR specs


chrisdolendo:
:+1:

PR #638, opened 2017-02-15
Features/new page for live in neighborhood #138684311

kaptron:


chrisdolendo:
this is a big one! can we get this merged asap?

chrisdolendo:
this just got accepted so i think we can merge! i'll let you do it @kaptron

PR #642, opened 2017-02-18
Features/easypost timeout #138280079

kaptron:
includes some debugging code to simulate the timeout -- remove before merging! 

(it also kills the RSpec tests because we've purposely broken address validation)


chrisdolendo:
we paired on this so :+1:

PR #658, opened 2017-03-04
[#139510873] Makes NRHP results visible until lottery results available

catcall:


kaptron:
👍

app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -197,12 +197,7 @@ ListingService = ($http, $localStorage, $modal, $q, $state, $translate) ->
       "#{Street_Address}#{City} #{State}, #{Zip_Code}"
 
   Service.showNeighborhoodPreferences = (listing) ->
-    return false unless listing.NeighborHoodPreferenceUrl
-    now = moment()
-    lotteryDate = moment(listing.Lottery_Date)
-    begin = lotteryDate.clone().subtract(9, 'days')
-    end = lotteryDate.clone().subtract(2, 'days')
-    return now > begin && now < end
+    !!listing.NeighborHoodPreferenceUrl && !listing.Lottery_Results
```
creativecoder: @catcall We already have `Service.listingHasLotteryResults()` method. Let's using that method instead of `listing.Lottery_Results` property, since the property is only true if someone remembers to manually set the lottery to complete in Salesforce.

PR #667, opened 2017-03-10
Features/new timer events for apply online #139762185

catcall:


creativecoder:
@catcall Looks like we need to get this branch back in sync with master. I think this is the best way to do that:

1. Checkout branch
1. Note the hashes of the commits you actually need from this branch
1. Do a hard reset back to the first shared commit with master (should be the one before [#138684311] Adds initial version of standalone NRHP page eaa5980)
1. `git cherry-pick` the commit hashes that you need for this feature (that you noted in step 2)

catcall:
@creativecoder cool, will do

PR #670, opened 2017-03-14
[#139762319] Adds GTM event for account creation

kaptron:


creativecoder:
@kaptron Looks like Lukas has approved this, so we can merge.

PR #672, opened 2017-03-14
Feature/mf community building q #136989917

chrisdolendo:


app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -636,6 +640,9 @@ ListingService = ($http, $localStorage, $modal, $q, $state, $translate) ->
     if opts
       Service.stubPreferences(opts)
 
+  Service.stubListingFeatures = ->
+    Service.listing.STUB_CommunityRestriction = 'All People'
```
chrisdolendo: I wonder if it makes sense to make `STUB_CommunityRestriction` values as 'Everyone' and 'Anyone'.

Or `STUB_CommunityRestrictionAll = true` for everyone and `STUB_CommunityRestrictionAll = false` for anyone.

Thoughts?

PR #673, opened 2017-03-14
Features/mf community building q #136989917

kaptron:


app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -636,6 +640,9 @@ ListingService = ($http, $localStorage, $modal, $q, $state, $translate) ->
     if opts
       Service.stubPreferences(opts)
 
+  Service.stubListingFeatures = ->
+    Service.listing.STUB_CommunityRestriction = 'All People'
```
kaptron: @chrisdolendo @catcall re: your earlier comment, the "All People" comes from the way they said it will eventually be set up in Salesforce, with "All People" or "One Person" being the values: https://www.pivotaltracker.com/n/projects/1397652/stories/139237931

chrisdolendo: @kaptron gotcha thanks for clarifying!

app/assets/javascripts/short-form/directives/communityScreeningHeader.js.coffee
```diff
@@ -0,0 +1,27 @@
+angular.module('dahlia.directives')
+.directive 'communityScreeningHeader', ['$translate', 'ListingService', ($translate, ListingService) ->
+  replace: true
+  scope: true
+  templateUrl: 'short-form/directives/community-screening-header.html'
+
+  link: (scope, elem, attrs) ->
+    listing = ListingService.listing
+    restriction = listing.STUB_CommunityRestriction
+    age = { minAge: listing.Reserved_community_minimum_age }
```
creativecoder: My intuition might have been to pass these in as attributes rather than get from Listing service--would love to discuss to best practices you've learned for directives at some point.

kaptron: @creativecoder yeah looking back I'm actually not sure why we set it up this way, usually I'd inject a service like that if it was a way to give the directive its own functionality without having to pollute the controller with extra things (if it's just to serve the directive). However the way we generally use directives (i.e. with `scope: true` which inherits the parent scope) it's more like a glorified `ng-include` that is a little more compartmentalized and somewhat "component-like". Anyway, given that, I updated this to just refer to `scope.listing` rather than needing to pull in the ListingService, which is unnecessary.

Good catch!

PR #674, opened 2017-03-15
Multifamily: Qualification Questions for Household section #136990399

creativecoder:
~~Note that this stubs the new preferences in listing_service.rb. These stubbed preferences will need to be removed before merging this feature into master.~~

Updated pull request to use multi Salesforce environment and set up Arc Mercy listing with new preferences for testing.


creativecoder:
@kaptron @chrisdolendo Does the monthly rent code include "stubbed saving" of total rent for the application and rents per address to the metadata field? I didn't see either of those in the pr diff.

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
```diff
@@ -32,6 +32,7 @@ ShortFormApplicationService = (
       assistedHousingRentBurden: null
     householdIncome: { incomeTotal: null, incomeTimeframe: 'per_year' }
     groupedHouseholdAddresses: []
+    prioritiesSelected: {}
```
kaptron: @callaghanc after all our rigorous checkbox error testing we missed _one_ case which is right when the application initializes, if `prioritiesSelected` isn't initialized to an empty object then it gets confused when you try to check for the properties `prioritiesSelected.no`, etc. which means that you were able to hit Next without selecting anything (meanwhile JS errors popped up about object being undefined). 😖 

Anyway, I updated this branch, and the review instructions since ashley was already testing the feature last night.

callaghanc: lol can't believe we missed that in the 1000 test cases we tried! nice catch @kaptron

PR #677, opened 2017-03-16
[#141844541] Fixes "create account" after submission

kaptron:
Resurrected some of the "claim_submitted_application" functions which were actually relevant to this feature, but made it much simpler and less convoluted than it was before. Uses formMetadata to verify that the application belongs to the same user session as the person trying to "claim" the application by creating an account after the app has already been submitted.

### DIFF
--- a/app/assets/javascripts/account/AccountService.js.coffee
+++ b/app/assets/javascripts/account/AccountService.js.coffee
@@ -50,18 +50,19 @@ AccountService = (
 
 
   Service.createAccount = (shortFormSession) ->
+    # loading overlay will be cleared on success by a state transition in the controller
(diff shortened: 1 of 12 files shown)


creativecoder:
@kaptron I didn't manually test the functionality, but the changes look good in the sense it seems cleaner and simpler.

I think I remember that we had some trouble before with proof of preference file uploads being correctly attached to claimed applications--is there any extra (manual) testing we should do here to check that?

kaptron:
@creativecoder don't totally recall the previous issue, however this shouldn't be a problem. "Claiming" in this case doesn't have anything to do with preference uploads, those have already been uploaded when you submit anonymously.

PR #678, opened 2017-03-17
Make "Do not apply online and also send in a paper application" on the sidebar dynamic #140878885

callaghanc:
…ppear as needed


app/assets/javascripts/listings/templates/listing/_panel-apply.html.slim
```diff
@@ -91,7 +91,9 @@
 
   .content-group.bg-snow
     p.c-charcoal.t-small
-      | Do not apply online and also send in a paper application. All applications containing any person who appears on more than one application will be removed from the lottery.
+      span ng-if="listing.Accepting_Online_Applications && !listing.Lottery_Results"
```
kaptron: @callaghanc is the `&& !listing.Lottery_Results` just to ensure that this listing is still open? I don't think that part of the `ng-if` is necessary because the whole `_panel-apply` section will already be hidden once the listing is closed.

callaghanc: @kaptron oh I see, that's a good point. I was using the exact criteria for the display of the "Apply Online" button (see line 34 in _panel-apply.html.slim) since the story AC were written using the presence/absence of the button as the criteria for the presence/absence of the extra "do not apply online and on paper" phrase. and it does look like the "Apply Online" button display logic checks for lottery results. there may be some business logic here that i'm not totally clear on - thoughts?

kaptron: @callaghanc ah, gotcha, makes sense that you copied the logic from there. Then the extra check is perhaps redundant in both of those cases, but I'll take a look at how the `_panel-apply` is getting loaded in the ng-include and make sure my understanding above is correct.

PR #679, opened 2017-03-20
Features/mf sf monthly rent #136990689

chrisdolendo:


kaptron:
@chrisdolendo it seemed like this should use #674 as the base branch since it is branched off that one, but I tried doing that and it seemed to put it in a weird state with other commits (maybe needs to be rebased off that source branch?) so I set it back to `multifamily`. Maybe we can take a look later.

kaptron:
also this may have been accidental (?) but it looks like the file naming/ordering got changed for the two pages in the income section, @creativecoder had renamed those `d1` and `d2` but it looks like they're `d6` and `e1` in this branch.

PR #684, opened 2017-03-23
[#136132929] - Request additional info on geocoder

chrisdolendo:


chrisdolendo:
@kaptron -- tried to update the VCR tests by deleting and then regenerating the cassettes but it still failed. Can you chime in? I'll review this PR when I get back so I get that knowledge transfer.

kaptron:
@chrisdolendo not sure what happened when you deleted the cassettes but all I had to do was:
- delete the 2 cassettes in `vcr/geocoding` dir (invalid.yml, valid.yml), re-run `rake spec`
- one more test was failing, had to add a migration for `geocoding_logs` so that they can save `state` param 

then ✅

PR #695, opened 2017-04-06
Fix listing link in print review screen

callaghanc:


kaptron:
looks good to me, just asked @slowbot one question in pivotal about the removal of the browser warning

PR #696, opened 2017-04-06
Features/rework lottery results #128698459

chrisdolendo:


chrisdolendo:
lottery ranking works, still need the buckets info filled in on the portal so I can test the buckets as well.

kaptron:
looks good, agreed we need that to test!

PR #698, opened 2017-04-07
Multifamily Application: Assisted Housing Preference #136990399

kaptron:


creativecoder:
@kaptron Added one more commit to update proof uploader instructions and classes to match patterns here: http://sf-dahlia-pl-mf-shortform.herokuapp.com/application/application-preferences--multifile.html

PR #701, opened 2017-04-10
Saves extra fields in application metadata #143458175

kaptron:
Also saves stubbed fields so that we can still simulate saving/loading via Salesforce


kaptron:
merging this chore into `multifamily` so that other MF features can inherit this feature. reviewed by @chrisdolendo

PR #703, opened 2017-04-12
Multifamily Application: Rent Burden Preference

chrisdolendo:
don't merge. used for review app purposes only.


chrisdolendo:
@kaptron tests updated

PR #707, opened 2017-04-18
Features/dynamically change plural singular #136990503

chrisdolendo:


app/assets/javascripts/short-form/ShortFormHelperService.js.coffee
```diff
@@ -83,12 +83,19 @@ ShortFormHelperService = ($translate, $filter, $sce, $state) ->
...
+  Service.addressTranslateVariable = (address) ->
     { address: address }
 
-  Service.membersTranslationVariable = (members) ->
+  Service.membersTranslateVariable = (members) ->
     { user: $filter('listify')(members) }
 
+  Service.youOrHouseholdTranslateVariable = (membersCount, wholeHousehold) ->
+    value = if membersCount > 0
+      if wholeHousehold then 'does your household' else 'you or anyone on this application'
```
kaptron: @chrisdolendo @callaghanc this story has been approved, and everything looks like it works fine -- but I hadn't really checked out the implementation. The problem with this setup is that phrases like "does your household" are hardcoded into English, and they are fragments of the sentence, so in a way it makes the whole text (probably) difficult to translate. Just for example, as a translator looking at the sentence to translate, this would look weird: 
`"TITLE": "How much {{members}} pay in monthly rent?"`

I can see how this is sort of a complicated scenario to fit into the translation strings, because of the variability, so I can't think of a lot of great options here. However perhaps the way to fix it would be:
 - put these hardcoded phrases as translation strings themselves, so that "your household" could be translated. Might still make for slightly difficult sentence translation but better than nothing. 
 - remove "do/does" from the strings (if possible) so that the "do/does" remains part of the overall sentence, and the translated part is just "your household, you or anyone, you, etc" which would fit more translation-friendly as just the subject of a sentence rather than sometimes mixing in a verb. 
Honestly though this might mean that in some cases it's easier to just have an `ng-if` with two different sentences, something like:

```slim
h2.app-card_question ng-if="members > 1" translate="C5_HOUSEHOLD_MONTHLY_RENT.TITLE_HOUSEHOLD"
h2.app-card_question ng-if="members == 1" translate="C5_HOUSEHOLD_MONTHLY_RENT.TITLE_YOU"
```

where `TITLE_HOUSEHOLD` would say "How much does your household pay in monthly rent?" and `TITLE_YOU`would say "How much do you pay in monthly rent?"

Anyway, we can chat at standup -- this could just mean merging this one as is, and adding an additional chore to clean it up as part of multilingual, so that we don't rabbit hole on this story which was already from 1-2 sprints ago.

chrisdolendo: @kaptron Good call! I like the approach of removing the do/does and having a an ng-if='' depending if it's singular or plural. Can we create this as a chore for the next sprint since we've already set a goal for this sprint?

PR #709, opened 2017-04-20
Updates to phantomjs 2.1.1 from version 1.9.8

kaptron:
Now that semaphore supports easily switching versions of phantomJS by adding one config command, this makes sense to do. Angular 1.5+ doesn't play well with phantomJS v1.


creativecoder:
@kaptron To add to that, the sole maintainer of PhantomJS announced that he's stepping down from the project. Looks like headless Chrome, while not quite ready yet, is the future of headless browsers. https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE

PR #710, opened 2017-04-20
Features/rent burden dashboard #141737079

chrisdolendo:


chrisdolendo:
@kaptron @callaghanc wrote tests for all of FileUploadSpec so this is done too

PR #717, opened 2017-04-26
[#141440837] Adds more info links to preference screens

callaghanc:


kaptron:
@callaghanc as we had discussed I updated this to pull from the Listing Preference `readMoreUrl` rather than being hardcoded, otherwise this is accepted and good to go -- merging now.

PR #718, opened 2017-04-27
ensure saving geocoder fields even after loading draft

kaptron:
quick fix to make sure geocoder fields get preserved, not only when you input your address but after loading a draft where you had previously stored geocoder metadata.


creativecoder:
@kaptron Do we need to do the same for household members, since they can have geocoder information as well if they have a different address?

PR #720, opened 2017-04-28
Features/mf review screen update #140887967

chrisdolendo:
This feature was branched from `features/rent-burden-dashboard-errors-#141737051`

Not finished -- still need at add tests


kaptron:
@chrisdolendo you switched `application.STUB_householdPublicHousing` to `listing.STUB_householdPublicHousing` but the first way was actually correct -- it's a stubbed field of a person's _answer_ (did they answer Y/N, however it is stubbed since we can't save it to salesforce until they add the field), and not of a feature on the listing. That's why the test is breaking. 

I just merged w/ the base branch, and also reverted it back to `application.STUB_householdPublicHousing`, which should fix E2E tests 🤞

chrisdolendo:
thanks @kaptron! that makes sense now lol. That should make it easier on the review instructions too. Just gotta wrap up some additional tests and will be delivering this today.

PR #734, opened 2017-05-08
Multilingual: Non-Latin character input error #143664229

chrisdolendo:


chrisdolendo:
delivered - need to update protractor tests

kaptron:
@chrisdolendo 👍  I guess there were actually less full text input fields than I was worried about, so not too bad! Added one minor fix to regex because spaces were getting rejected as "not english"

PR #741, opened 2017-05-15
Chores/translate plural singular shortform #144724381

chrisdolendo:


chrisdolendo:
co from wrong branch, cherry picking to right co and re-opening.

PR #743, opened 2017-05-15
[#144724381] - translated singular/plural dynamic rendering

chrisdolendo:


public/translations/locale-en.json
```diff
@@ -91,7 +91,8 @@
         "HOW_MUCH_DO_MEMBERS_PAY": "How much do {{members}} pay at {{address}}?",
         "HOW_MUCH_DO_YOU_AND_MEMBERS_PAY": "How much do {{members}} pay at {{address}}?",
         "HOW_MUCH_DO_YOU_PAY": "How much do You pay at {{address}}?",
-        "TITLE": "How much {{members}} pay in monthly rent?"
+        "TITLE_HOUSEHOLD": "How much does {{members}} pay in monthly rent?",
+        "TITLE_YOU": "How much do {{members}} pay in monthly rent?"
```
kaptron: @chrisdolendo I guess I'm not seeing why `{{members}}` really needs to be a variable for these ones. Isn't `TITLE_YOU` _always_ `"How much do you pay in monthly rent?"` ? And I think that was my point on the original story, it will always be easier for a translator to translate: `How much do you pay in monthly rent?` rather than translate `How much do {{members}} pay in monthly rent?` and then separately translate the word `You`, but knowing that it's the version of `You` that grammatically fits into that first sentence as `{{members}}`.

kaptron:
@chrisdolendo see my comment above, let me know what you think

chrisdolendo:
@kaptron ah yeah,  I did a mix of the two lol! Updated.

app/assets/javascripts/short-form/templates/c4-household-public-housing.html.slim
```diff
@@ -1,14 +1,22 @@
...
+      | {{'C4_HOUSEHOLD_PUBLIC_HOUSING.TITLE_HOUSEHOLD' | translate }}
     p.form-note.margin-bottom
       | {{'C4_HOUSEHOLD_PUBLIC_HOUSING.P1' | translate}}
 
   .app-inner.inset
     .form-group ng-class="{ error: inputInvalid('STUB_householdPublicHousing') }"
       p.form-note.margin-bottom translate="LABEL.PLEASE_SELECT_ONE"
-      .radio-group.padding-top--half role="radiogroup" aria-label="{{'C4_HOUSEHOLD_PUBLIC_HOUSING.TITLE' | translate}}"
+      .radio-group.padding-top--half role="radiogroup" aria-label="{{'C4_HOUSEHOLD_PUBLIC_HOUSING.TITLE_YOU' | translate}}" ng-if="householdMembers.length == 0"
+        radio-block-item value='Yes' user='application' name='STUB_householdPublicHousing' ng-required='true' on-change="publicHousingYes"
```
kaptron: @chrisdolendo sorry to have to keep hammering this one story 😩  but it looks like this whole block of Yes/No gets repeated just because the `aria-label` can change based on `householdMembers.length`? Isn't there a better way to do this, like using `aria-labelledby` to point to the `h2` that already has the right text?

chrisdolendo: @kaptron yeah I think that's why we did a function to render the right words.. but yeah! would it still serve the right purpose of having `aria-labels` if we did it this way:

```
h2.app-card_question ng-if="householdMembers.length == 0" aria-label="{{'C6A_HOUSEHOLD_RESERVED_UNITS_VETERAN.TITLE_YOU' | translate}}"
  | {{ 'C6A_HOUSEHOLD_RESERVED_UNITS_VETERAN.TITLE_YOU' | translate }}
h2.app-card_question ng-if="householdMembers.length > 0" aria-label="{{'C6A_HOUSEHOLD_RESERVED_UNITS_VETERAN.TITLE_HOUSEHOLD' | translate}}" 
  | {{ 'C6A_HOUSEHOLD_RESERVED_UNITS_VETERAN.TITLE_HOUSEHOLD' | translate }}
```

that will remove the radio buttons.

kaptron: @chrisdolendo yes, right idea but not quite, the aria-label doesn't belong on the h2, the aria-label is the accessibility label for the radio group to indicate which text is labelling it. so more like

```slim
h2#question-title.app-card_question ng-if="householdMembers.length == 0"
  | {{ 'C6A_HOUSEHOLD_RESERVED_UNITS_VETERAN.TITLE_YOU' | translate }}
...
.radio-group.padding-top--half role="radiogroup" aria-labelledby="question-title"
```

and I just used `question-title` as an example ID, basically the ID of the h2 has to match with the labelledby attr.

chrisdolendo: @kaptron Ah I see, didn't know you could point the area label to an ID of another element. Cool! I'll do that.

chrisdolendo:
@kaptron updated aria labels :)

kaptron:
bitmoji

PR #750, opened 2017-05-23
[#142534427] - Implement autofill get started sign in page

chrisdolendo:
- Directs user to sign-in page and then redirects back to welcome page after sign in


kaptron:
👍

PR #751, opened 2017-05-24
Features: Autofill reset data #141451371

kaptron:


kaptron:
@chrisdolendo this was the branch I was just showing you -- you may just want to add your work into this same branch? again for `completedSections` it should just be a simple change, that would go into `Service._autofillReset` in ShortFormDataService

PR #756, opened 2017-06-01
Chores/translate repeater fields dynamic options #120224485

mluedke2:
I didn't include tests for this because it's just UI strings. If I should do so, just let me know (I need some help figuring out how to do that).


chrisdolendo:
:+1: nice!

chrisdolendo:
@mluedke2 I think this can change the base to master since it's not really a multilingual story

creativecoder:
@mluedke2 I don't think there are any tests needed for this story

@chrisdolendo I see your point about basing off of master. However translated strings will only start being used in the multilingual branch, so I'd vote for merging it to multilingual since that's what the story indicates (from tag), that's where product will be expecting to test it.

creativecoder:
Plus we need to send these new strings to the `translations` branch so they sync up with POEditor, and that branch is based off `multilingual`. Now that I think about that, we need to make sure that `translations` gets updated as soon as this branch is merged!

PR #761, opened 2017-06-02
Features/multiple reserved types #137567887

mluedke2:
Copied the `icons.html` over from the `multifamily-shortform` branch of the pattern library [here](https://raw.githubusercontent.com/Exygy/sf-dahlia-pattern-library/multifamily-shortform/src/views/layouts/includes/icons.html).

Has a bugfix Grant found to get `reservedDescriptor` in the `listings` object from the Salesforce API on the listings page.

I'm not wild about the hardcoded 0/1/2/3, so if anyone has ideas to improve that, I'm open. But it's only in 3 spots so I didn't worry about it.

I made the `or` translated, requiring that I inject the translator into `ListingController`. If I should move this into the `ListingService` instead, let me know and I'll make that change.


kaptron:
I couldn't think of a better way to do the 0,1,2,3... but then I did! 😁  I set it up to be an array so that rather than checking for 0,1,2,3 it just looks up in the index in the array of icons.

PR #766, opened 2017-06-09
Features/language landing pages #146652657

mluedke2:
Just a note: The first few commits had a lot of work to translate the housing counselor cards (use Idiomas instead of Languages, for example). In this final version, the cards don't fully match the Google Docs on the Pivotal story because all dynamic words are still in English. I will note this in the Pivotal review instructions.

If we do a squash-and-merge, the commit history will be cleaner but that work will be lost except on my local machine, in case we need it later. It was only placeholder for when we move the translations to POEditor anyway, but if the product owner would prefer all the little strings to be translated, it may come in handy.


kaptron:
👍  Looks good, just made some minor style updates e.g. the `housing-counselor ul li` was getting bullets added to it because that's how we globally set up `.page-content ul li` to behave, however that was really only to serve for making bulleted lists on the privacy page. So I updated that to be more specific (i.e. `ul.bullet-list`) when we want that style. 

@mluedke2 Also to your note, squash-and-merge + deleting the branch doesn't totally wipe the commits from history, they're still in the actual git/github history, see for example this PR which the 2 commits were squash-and-merged, but you can still view the PR and its individual commits (which is nice). You just wouldn't be able to view the file in `master`, view the blame/history for a file and see those individual commits any more (instead you'd see the single squashed commit). 
https://github.com/Exygy/sf-dahlia-web/pull/736

mluedke2:
interesting-- glad there is some history somewhere @kaptron

PR #767, opened 2017-06-09
Chores/e2e proof attachments #144587401

mluedke2:


spec/e2e/step_definitions/short_form_steps.coffee
```diff
@@ -138,6 +138,25 @@ module.exports = ->
...
 
+  @When /^I upload a "([^"]*)" as my proof of preference$/, (documentType) ->
+    # open the proof option selector and pick the indicated documentType
+    element.all(By.id("liveInSf_proof_option")).filter((elem) ->
+      elem.isDisplayed()
+    ).first().click()
+    element.all(By.cssContainingText('option', documentType)).filter((elem) ->
+      elem.isDisplayed()
+    ).first().click()
+    filePath = '/Users/mattluedke/Desktop/projects/dahlia/sf-dahlia-web/public/images/arrow-down.png'
```
mluedke2: this needs to be made not local

mluedke2: @kaptron interestingly, it doesn't have to be a valid file here for the tests to pass. It passes on all the following options:

`filePath = '/literally-whatever.png' #'/arrow-down.png' # '/images/arrow-down.png' #'/public/images/arrow-down.png'`

kaptron: @mluedke2 hah, weird... I guess that's actually ok though to just have some fake filepath? as long as the E2E test has made the browser "think" it uploaded some kind of file, enough to pass the test, that's all we're really going for. Basically uploading a "mock file".

mluedke2: Ok. I'll at least leave a comment in the test so it's clear that the file upload itself is not important to succeed.

creativecoder:
Not actually loading a file feels like cheating. But if that's the only way we can get it to work...

Did you try a relative file path? When I've seen (server-side) e2e tests do this in the past, usually it's by uploaded a test file committed in the repo.

kaptron:
@creativecoder you're exactly right, it actually was cheating 😈 . I fixed it to use an actual file path (see 661d6c6) and in doing so it actually uploads a file (i.e. the server processes the file upload and it shows the uploaded date) whereas the fake file was just somehow making the file input valid, it showed the fake file name but it did not actually hit the server and uploaded date was blank.

mluedke2:
@creativecoder refactored the element id to kebab-case

PR #768, opened 2017-06-12
Features/activejob setup #146963669

kaptron:


spec/jobs/short_form_attachment_job_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
```
chrisdolendo: rad

chrisdolendo:
right on 👍

kaptron:
@creativecoder you're exactly right, good catch -- I just fixed `schema.rb`. It was because of me jumping between branches and my database being in different states, the schema ended up getting combined/confused.

app/services/cache_service.rb
```diff
@@ -0,0 +1,51 @@
+# service class for pre-caching salesforce listings
+class CacheService
```
kaptron: @creativecoder would be interested to discuss `CacheService` with you and get your thoughts. As I mentioned last week we had to make some decisions/compromises as how to best pre-cache listings on a regular basis, and how frequently to do it. I think it's in a "good enough to deliver" place but I think it's worth one more conversation to finalize.

PR #769, opened 2017-06-13
Features/modals not popups #141341079

mluedke2:


kaptron:
@mluedke2 nice E2E test! And I'm realizing the E2E tests don't need to change as drastically as I thought... the `getUrlAndCatchPopup` having to close the browser alert is for the `$window.beforeunload` i.e. the alert that happens when you try to close the browser, which happens between tests. And since this behavior is necessarily always going to be a native browser alert (because that's just the `beforeunload` works), all of this is still relevant. 👍

kaptron:
Also I modified one part of it, I think we _do_ want to keep the "OK" button for alerts (i.e. that's how the standard browser alert has it, both a closing "x" and an "OK") but it's just the "OK / cancel" combo that only appears when there's an `onConfirm`, see cb50930

mluedke2:
Keeping the "OK" button makes sense to me 👍

mluedke2:
CSS edits in [PR #38 on pattern library](https://github.com/Exygy/sf-dahlia-pattern-library/pull/38)

chrisdolendo:
:+1: great job on creating modal service tests.

chrisdolendo:
When this is merge, go to `angularProviders` file and change the timeout code `alert(alertMessage)` to use the Modal Service

PR #770, opened 2017-06-13
Features/Update privacy policy text #143887647

mluedke2:
Just FYI: I didn't see an example of another translated string having a hyperlink in `locale-en.json` before this, but apparently it's fine, as long as you do the `translate="key"` method. If you try to do `{{'KEY' | translate}}`, the HTML doesn't render.


chrisdolendo:
:+1:

kaptron:
@mluedke2 that's right... the `translate` directive (i.e. `translate="KEY"`) will allow you to do HTML whereas the filter will escape it. Not made super obvious in the angular-translate docs but we did come across that before as well.

creativecoder:
@mluedke2 I'm going ahead and merging this for you since we want to add it to sprint 40 release that's currently being tested.

PR #774, opened 2017-06-14
Features/mobile horizontal scroll #141712899

mluedke2:
The diff looks larger than the amount of actual change involved because of indentation. All of these changes are just tables being wrapped in a `div style="overflow-x: auto;"`, or main page content being wrapped in `div style="max-width:100vw;"`


kaptron:
@mluedke2 rather than indent those individual pages and add the `max-height: 100vh` I think the intention was to literally add that globally to `section#main-content` i.e. in `application.html.slim`, that way that view height style is just available globally rather than having to stick it on certain pages. 

also, even though putting inline styles (`style="xxx"`) is nice for prototyping a style change, generally this would warrant adding CSS classes (i.e. adding something like `.view-height` and `.overflow-x` classes in the pattern library, and bringing that over) -- though I'd defer to @creativecoder / @slowbot, for these one-off utilities is that the right thing to do? or is just inlining these kinds of specific styles actually the 2017 way to do it?

mluedke2:
@kaptron Ah ok, I see. I'll make the `section#main-content` change so it's available globally.

I'll hold on making the CSS format change just yet, but my intuition says I should make it because that makes a lot of sense too.

creativecoder:
@kaptron Yes, let's avoid inline styles. While they are all the rage for css in js solutions, this is not that.

I'd probably do something semantic, like
```html
<div class="scrollable-table-container'>
  <table>
    ...
```
and
```css
.main-content {
  max-width: 100vw;
}

.scrollable-table-container {
  overflow-x: auto;
}
```

mluedke2:
Is there a specific CSS file in the pattern library where I should put these, or just do my best to pick the most appropriate? For the table container I'm thinking [components/_tables.scss](https://github.com/Exygy/sf-dahlia-pattern-library/blob/master/src/assets/toolkit/styles/components/_tables.scss), and for the `main-content` I'm thinking I'll make a new file called `components/_section.scss`. Figured I might as well check before I do it...

mluedke2:
@kaptron @creativecoder Redid using the pattern library CSS. Those changes are in [PR #37](https://github.com/Exygy/sf-dahlia-pattern-library/pull/37) in that repo.

mluedke2:
@kaptron reconfigured to use `main-content` class instead of id

PR #780, opened 2017-06-26
Chores/Test out google crawling #111657410

kaptron:
Follows steps listed [here](https://robots.thoughtbot.com/block-web-crawlers-with-rails) to prevent  Google from crawling non-prod sites by setting an ENV var that will switch the robots.txt and `X-Robots-Tag` settings delivered to search engines.


creativecoder:
@kaptron I assume we need to add `DISALLOW_ALL_WEB_CRAWLERS` env variable to all our development apps?

kaptron:
@creativecoder that's right -- and thanks for adding that `deployment config` label to the relevant stories! I've prepped our existing non-review / non-prod apps to have this ENV setting, and I added it to `app.json` in this PR so that future review apps (once this is merged) will automatically pick up this setting from dahlia-qa.

kaptron:
closing + re-opening to test new app.json

PR #785, opened 2017-06-29
Chores/full/e2e tests preference holder name change #147545945

creativecoder:
- Updates e2e dependencies
- Adds e2e test
- Takes steps toward refactoring and better organizing e2e test scenarios

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,15 @@
 {
+  "private": true,
   "author": "Exygy",
   "description": "DAHLIA is the affordable housing portal for the City and County of San Francisco.",
+  "license": "GNU GENERAL PUBLIC LICENSE",
   "devDependencies": {
     "bower": "1.7.9",
-    "chai": "3.5.0",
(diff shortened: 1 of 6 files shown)


kaptron:
@creativecoder I noticed this updates some of the versions of protractor/cucumber/etc, I'm assuming that was you trying to get it working? Anyway if all the tests still work fine then that looks good to me. I don't remember exactly why we needed to fork `protractor-cucumber` but I think it just included an unmerged fix that someone had proposed, and we needed it to resolve some bug/error -- again if the new version seems to work fine then I think we're ok, and probably better off moving back to a legit version.

creativecoder:
@kaptron Good question. Yes, I updated those while I was trying to get e2e tests working, and they work fine, so I left them updated. One thing I did **not** update was cucumber to version 2, since it works quite differently compared to version 1.

`protractor-cucumber-framework` did end up finding a maintainer and getting all pending PRs merged, so we don't need to use/maintain our forked version anymore.

PR #788, opened 2017-06-29
features/full/do not reply email #146741427

mluedke2:
Note that the default email was in two places, and the one in `devise.rb` was being overwritten by `emailer.rb` anyway. I removed the overwritten one (to be DRY) and left a comment.


config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = 'dahlia@housing.sfgov.org'
+  # config.mailer_sender = 'example@email.com'
```
kaptron: good catch @mluedke2, I just set it back to be a default commented out config option rather than just the "note", the comments above already explain what's going on there

chrisdolendo:
:+1:

kaptron:
marked this `WIP` just because I'm organizing our PRs and this one still has a pending city/content question: do we want to use `donotreply@housing.sfgov.org` or does that actually get us more likely to land in spam?

chrisdolendo:
Tasks:
1. Complete this chore: https://www.pivotaltracker.com/story/show/155236537
2. Fix semaphore tests

PR #790, opened 2017-06-30
Chores/qa/change neighborhood match field name #147354805

creativecoder:


kaptron:
@creativecoder approved -- although it doesn't seem like it belongs on `multilingual` I get it, since that is what's currently on QA (could also branch to `mf-shortform` which is the base of `multilingual`). Basically since we don't have a branch that straightforwardly represents "QA" this seems fine.

PR #799, opened 2017-07-12
Features/qa/mf dynamic pref no proof #133548407

chrisdolendo:


kaptron:
@chrisdolendo see 8e0fbc9, I think that is enough to make this work on saving/loading drafts! I also merged upstream w/ multifamily to incorporate some fixes. However it looks like there is at least one broken jasmine test, so please take a look at that.

chrisdolendo:
@kaptron thanks doing it now!

PR #802, opened 2017-07-14
Chores/full/Run e2e tests in ie #144303197

creativecoder:
Pre-merge tasks, update Semaphore sauce project:
1. SAUCE_URL that points to dahlia-full
1. Add master branch to project


kaptron:
@creativecoder I see how you got it to disable the `onunload` alerts, nice... makes the tests run a little smoother locally too! 👍 

Anyway, I think we're good to go here with the plan being:
 - all PRs still go to their individual semaphores (QA, Full) 
 - merges to master will build on Semaphore Full (`rake spec, rake jasmine, npm run protractor`) 
 - merges to master will _also_ build on Semaphore Sauce (just `npm run saucelabs`) 

The one thing to note is that with the Live/Work + NRHP tests, even in chrome it takes a few additional minutes. On IE, it's currently running, so we'll see how long it takes now, but... I bet it may take closer to an hour 😬  
Which is ok, since it's really just a safeguard for when we merge things to master. But just something to keep an eye on as we add more E2E tests, if it ever becomes too cumbersome for IE then we may limit which cucumber scenarios are run by `npm run saucelabs` to just some of the core ones (e.g. submitting full application). 

Will wait and see how long this one takes now (and hopefully tests still all pass in IE).

creativecoder:
@kaptron That all sounds good.

I think the advantage of keeping IE in a separate SauceLabs project is that we don't have to wait to see green in github if all the non-IE tests pass.

The disadvantage is that the semaphore/github integration isn't really set up to handle more than 1 semaphore project for a given branch. I'm not sure how it will handle the 2 projects... I'd guess we'll see in github the semaphore project that had its webhook called most recently.

If we want to keep the 2 project structure in semaphore to separate the long running saucelabs tests, maybe we should remove the web hook for the sauce project? Then we'd have to check manually before releases to make sure they are passing. Unless you have any other ideas?

kaptron:
@creativecoder oh I see what you mean, I guess the 3 project setup can still exist but perhaps be more like:
 - Semaphore QA (PR branches only)
 - Semaphore Full (PR branches only)
 - Semaphore Master (any merges/commits to master, setup is similar to Full, runs all tests, but also runs saucelabs) 

I'd be good with that as well.

creativecoder:
Looks like these are now timing out `The test with session id e597de96a5734dbfb27acd84804327c2 has already finished, and can't receive further commands.` I'll look into this.

kaptron:
@creativecoder the test is [limited to 30 mins by default](https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options#TestConfigurationOptions-MaximumTestDuration) and it sounds like that is based on their recommendation 🤔  

So perhaps we should already implement my suggestion of limiting down the IE tests to a set of "core" ones and not all the other little back/forth interactions e.g. for Live/Work.

kaptron:
@creativecoder nice looks like you already did this and it works! I just merged the conflict w/ master, so assuming the tests pass _one more time_ I'll get this all merged.

chrisdolendo:
:+1: looks good, rebuilding the semaphore again since it failed before even running the tests

PR #804, opened 2017-07-17
Features/mf single listing lottery preference #147453601

chrisdolendo:
displays pdf link to a lottery preference for listing


kaptron:
closing this branch to use the renamed one with `qa` so it connects to semaphore

PR #805, opened 2017-07-19
Bugs/full/Autofill release mobile CSS fix #149122341

kaptron:


kaptron:
will merge this so the release can be fixed

PR #808, opened 2017-07-19
Features/qa/mf single listing lottery preference #147453601

kaptron:


kaptron:
@chrisdolendo 👍  on this, just made some minor edits to it e.g. `showPreferenceListPDF` sort of became a redundant function now that it is really based on the field being available in the ng-repeat and `listingHasLotteryResults` -- I also made the ng-repeat on the `.content-group` which makes it look a little nicer in the rare case that there happens to be more than one. It also makes the logic a little cleaner -- the entire `.content-group` shouldn't be shown _at all_ unless there is a PDF. 

I also updated the review instructions to point to `dahlia-qa-pr-808`.

chrisdolendo:
@kaptron thank you so kindly for those edits, I agree the function becomes redundant since the logic became simpler. Right on!

bitmoji

chrisdolendo:
Updated for: https://www.pivotaltracker.com/story/show/149900328

PR #809, opened 2017-07-20
Features/full/maintains scroll position when navigating back/forward

creativecoder:


chrisdolendo:
:+1:

creativecoder:
Looks like I missed merging this after it was accepted--will merge it now.

PR #811, opened 2017-07-21
Features/full/update sogi questions #147369275

mluedke2:
This one PR contains the work for both #147369275 and #147369129, since the work was very related.

It also contains some of the work done in `multifamily` to translate all the dynamic options for the optional demographics dropdowns, so those merge conflicts can be handled now.


chrisdolendo:
:+1: thanks for using ShortFormHelperService more!

kaptron:
Merging this because we know it needs to be part of the next release. Will update review instructions to point to dahlia-full.

PR #812, opened 2017-07-22
Features/mf nrhp autoclaim upload proof #146338037

chrisdolendo:


app/assets/javascripts/short-form/components/preferenceProofUploader.js.coffee
```diff
@@ -72,6 +72,10 @@ angular.module('dahlia.components')
           opts = @rentBurdenOpts()
         FileUploadService.uploadProof($file, @preference, @listingId, opts).then =>
           @afterUpload()
+        if @preference == 'neighborhoodResidence' || @preference == 'antiDisplacement'
+          # if we're uploading for NRHP/ADHP, it also copys info and uploads for liveInSf so that the file info is saved into DB
+          ShortFormApplicationService.copyNeighborhoodToLiveInSf(@preference)
```
chrisdolendo: @kaptron I had to call this function on upload in order to get the file info saved into the database. Before this, we only transfer the info during `copyNeighborhoodToLiveInSf` which is called as part of the callback `checkAfterLiveInTheNeighborhood` when you click next (on ADHP/DTHP page). Now that this order of operations makes more sense, I wonder if we even need to called `copyNeighborhoodToLiveInSf` in the callback. My sense is no, so I want to remove it. Thoughts?

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -370,8 +370,6 @@ ShortFormApplicationController = (
   $scope.checkAfterLiveInTheNeighborhood = (preference) ->
     # preference is either neighborhoodResidence or antiDisplacement
     if ShortFormApplicationService.applicationHasPreference(preference)
-      # Neighborhood provides automatic liveInSf preference
-      ShortFormApplicationService.copyNeighborhoodToLiveInSf(preference)
```
chrisdolendo: @kaptron  this is the redundant function i suggested to remove from my first comment.

kaptron: @chrisdolendo it is _almost_ redundant but not entirely, just thinking of edge cases... like if the person filled out Jane Doe + uploaded proof, all that would get copied over. If they then switched the dropdown to Jim Smith, that wouldn't get copied over to Live in SF. So we'd either need to add a callback to that dropdown or keep this one..... actually yes maybe a callback on the member dropdown is actually preferred because it would solve the case where they change the dropdown but jump around in the nav without hitting "Next".

chrisdolendo: so yes you agree to this approach @kaptron? 😄

kaptron: @chrisdolendo yeah I agree we can get rid of this happening in `ShortFormApplicationController` but I'm saying we should also add a callback on the member select dropdown, basically the same behavior as above: `if @preference == 'neighborhoodResidence' || @preference == 'antiDisplacement' ... copyNeighborhood...`

chrisdolendo: @kaptron Ah I see yes because we need to sync up the member too! Good catch, will implement.

chrisdolendo:
closing this to open a new one that is namespaces with QA

PR #815, opened 2017-07-24
Bugs/full/missing image error #147486157

mluedke2:
Worked with @callaghanc and @creativecoder


public/404.html
```diff
@@ -1,67 +1,83 @@
...
+      -webkit-margin-after: 0px;
+      -webkit-margin-before: 0px;
+      font-size: 15px;
+      color: #7F828B;
+      line-height: 21px;
+      margin-bottom: 4px; }
+    }
+    </style>
+  </head>
+  <body><script id="__bs_script__">//<![CDATA[
```
creativecoder: @kaptron @mluedke2 This 404 page looks nicer! I don't think we need this browser sync markup, though.

config/routes.rb
```diff
@@ -63,6 +63,7 @@
 
   # catch all mailer preview paths
   get '/rails/mailers/*path' => 'rails/mailers#preview'
-  # required for Angular html5mode
-  get '*path' => 'home#index'
+
+  # catch all to send all HTML requests to Angular (html5mode)
+  get '*path', to: 'home#index', constraints: -> (req) { req.format == :html }
```
creativecoder: @kaptron Always showing us a more elegant way, thank you!

public/404.html
kaptron: Yeah noticed that browser sync and was getting rid of it, must not have pushed that (I corrected the same for our maintenance pages on S3 which is where I got the layout).

config/routes.rb
kaptron: The other way didn't fully work 😊 because it only sent ".html" requests to angular. 

So I was playing around with this 404 behavior a bit and it works great for the most part the way Rails handles it out of the box -- the only thing that is perhaps a little annoying is that in the logs it still shows a RouterError and a stack trace (but the fact that it's serving a 404 and not a 500 should fix the "false" errors in NewRelic at least). 

So I looked into a couple alternatives to handle it if we wanted cleaner logging:
- Final catch all route that goes to an error handler controller which serves up a 404. Pro: no error w/ stack trace. Con: adding extra setup e.g. responding in the correct format that is basically replicating what Rails seems to be doing for you in its prebuilt 404 behavior. 
- Logger initializer that captures RouteErrors and just logs them differently / without stack trace. Also feels like somewhat unnecessary/ overconfiguration but I like it better as an option. 
- or just leave it alone?

chrisdolendo:
:+1:

creativecoder:
@kaptron @mluedke2 I checked New Relic after seeing another error alert yesterday. Looking through the errors, I some additional cases where we are getting a 500 response when a 404 would be appropriate

- `/api/v1/listings/.json`
- `/.well-known/dnt-policy.txt`
- `/api/v1/short-form/listing-application/`

The review app is now returning a 404 for those--nice!

However, any idea how people would come to request the API urls formed like that?

For example, the `listing-application` request, (which I think error'ed because it didn't have an application id like the route specifies) came from `https://housing.sfgov.org/applications/a0o0P00000FFVXEQA5`... seems legit.

Here's the New Relic page https://rpm.newrelic.com/accounts/1181758/applications/17861453/traced_errors/3fc86572-70f0-11e7-8a49-0242ac11000f_0_15201?original_error_id=2f25d090-716a-11e7-8a49-0242ac11000f_0_15184

kaptron:
@creativecoder yeah I'm not sure how those would happen... examples 1 and 3 are both valid URLs but with a listing id missing, not sure if it's just some javascript error that may be happening on the browser side or what, or how that would be getting malformed.

creativecoder:
I think I figured it out. If you come back to a previously submitted application through your account and you can get in a state where the listing is not in memory, and so all of the properties on the listing object (including ID for generating the link) are null. I'll file a bug!

PR #817, opened 2017-07-25
Features/full/auto resize listing images #128104539

creativecoder:
This PR comes with some new environmental variables

shared across all environments
`S3_ACCESS_KEY_ID`
`S3_ACCESS_KEY`

unique to each salesforce environment
`S3_BUCKET`
`RESOURCE_URL`

These heroku environments have been updated with the new env settings
`dahlia-full`
`dahlia-qa`
`dahlia-production`
`dahlia-full-multilingual`
`dahlia-qa-mf`
`dahlia-qa-multilingual`

When these code is deployed to dahlia-full and to production, run `rake listing_images:process_images` to populate the database with image_urls for each listing

Also, I've set up a log alert to check for `ListingImageService` errors on production in papertrail and email/slack us if that occurs in production. This will alert us to unreadable or missing images for listings.


chrisdolendo:
liking the structure of this feature -- is it still WIP? Or are the config vars updated on the pr app? Opened it and it looks like there's only one listing that has an image.

creativecoder:
@chrisdolendo You're right it's still WIP--I thought it was finished, but found a few more things that need to be resolved, like adding separate s3 buckets for each Salesforce environment.

creativecoder:
@kaptron @chrisdolendo would love to get your feedback on this, particularly on the listing image service and its tests.

Based on the article that Dave shared (http://multithreaded.stitchfix.com/blog/2015/06/02/anatomy-of-service-objects-in-rails/) I think this could be a pattern for improving some of our existing services and adding some tests.
- Services should primarily be instances (rather than singleton classes)
- Minimal public methods

I did keep FileStorageService as a singleton because it think we only need 1 remote connection to S3, rather than a new connection each time the service is called. Would need to reconsider this if we ever moved to a multi-threaded context, though.

Also, the ListingImageService tests seems a little strange to me in places (stubbing private methods inside the test in one instance, depending in their behavior in another place). On the other hand it tests all of the things I want to make sure happen when processing an image, so maybe it's just about right.

app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -103,7 +103,7 @@ ListingController = (
   $scope.imageURL = (listing) ->
     # TODO: remove "or" case when we know we have real images
     # just a fallback for now
-    listing.Building_URL || 'https://placehold.it/474x316'
+    listing.imageURL || 'https://placehold.it/474x316'
```
kaptron: @creativecoder should we finally just get rid of this function and refer directly to `listing.imageURL`? I think it's only referred to in one place anyway. And then I think our practice should be, if a listing layout is messed up because of no image (e.g. this sometimes happens on fake listings on QA) then just add an image in salesforce rather than relying on this placeholder to correct the layout.

creativecoder: I like keeping this as a fallback, just in case we ever had a problem with images on production, either because of a missing/malformed image url, or because we couldn't process (resize and optimize) the image for some reason. That way the layout would still flow okay until the problem was resolved and the cache refreshed.

PR #818, opened 2017-07-25
Features/full/dynamic sign in before submit #146654295

kaptron:


app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -127,6 +128,16 @@ AccountService = (
         msg = $translate.instant("ERROR.PASSWORD_UPDATE")
       Service.accountError.messages.password = msg
 
+  Service.checkForAccount = (email) ->
+    $http.get("/api/v1/account/check-account?email=#{encodeURIComponent(email)}").success((data) ->
```
chrisdolendo: Hai - why did you need to `encodeURIComponent` the email?

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
```diff
@@ -505,9 +505,14 @@ ShortFormApplicationService = (
...
     # check if this user has already applied to this listing
     Service.getMyAccountApplication().success((data) ->
-      if !_.isEmpty(data.application) && Service._previousIsSubmittedOrBothDrafts(data.application)
+      previousApplication = data.application
+      if !_.isEmpty(previousApplication)
         # if user already had an application for this listing
-        return Service._signInAndSkipSubmit(data.application)
+        if opts.type == 'review-sign-in' && previousApplication.status.match(/draft/i)
+          # override draft ID and proceed, so that we eventually call submitApplication...
+          Service.overridePreviousDraftId()
```
chrisdolendo: Is overriding the draft ID with that was already saved happen after then choose which draft to use?

chrisdolendo:
just had a couple of Qs!

app/assets/javascripts/account/AccountService.js.coffee
kaptron: @chrisdolendo noticed particularly for when using `+` in the email like `dave+test@exygy.com`, it was breaking.

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee.erb
kaptron: @chrisdolendo so yeah the same `overridePreviousDraftId` is used for choosing a draft, however in this case they have finished/reviewed their latest draft and are ready to submit, so it is sort of counterproductive to say "hey you had an old draft, do you want to use it?" since that one is probably less finished than this one, and they'd have to start back over. So in this case we basically silently overwrite their old draft (without asking them) so that they still only have _one_ application but it is replaced with this newly finished one. (we checked with Ashley that this is how we should proceed)

Will add a better clarifying comment in the code than the current one.

kaptron:
@chrisdolendo it's on master/full because it's not a multifamily-specific feature (`review-sign-in` page exists on `master)

chrisdolendo:
bitmoji

PR #819, opened 2017-07-26
Features/qa/mf nrhp autoclaim upload proof #146338037

chrisdolendo:

PR #821, opened 2017-07-26
Features/full/resize proof images #148837125

mluedke2:
@chrisdolendo and I paired on this. We figured out that we need to do the resizing in JS instead of in HTML to get it to work correctly. If the file is over 5MB, it resizes to 2112 max w/h. Otherwise, it uploads same as before.

This is one option to structure the JS. Chris and I had originally considered making the `_uploadProof` method private, and calling the `processProof` directly from the HTML. It makes more intuitive sense, but breaks future work that depends on `uploadProof()` in #819. I was thinking of switching back to that and Chris agrees, but now I'm going back and forth.

Another option is to use a `Promise` instead of the callback method.


mluedke2:
Thanks @creativecoder 

I'll add a check for 5MB either after resizing (or not if it's PDF and wasn't resized) in.

I didn't test with any PDFs so I'll test that too, and make sure if they aren't resized they are at least size-limited.

It appears that it is already the case that PNGs are converted to JPEGs. When I was testing I used a PNG file and it is saved in Salesforce as a JPEG.

chrisdolendo:
Thank you for the clarification, @creativecoder :+1:

Though we interpreted the eng notes as ways to implement the acceptance criteria, when we uncovered (via github issues) that the directives suggested in the engineering notes weren't fully functioning, we focused on other ways we can meet the acceptance criteria.

So for future stories, it would be helpful to translate critical engineering notes into acceptance criteria to ensure clarity. E.g. `ng-validate-after-resize` engineering note and the other clarifications you offered, can be written as an acceptance criteria "1. Resize image larger than x size. 2. Validate size of resized image before saving into database. 3. All images should be saved as JPGs 4. Because pdfs can't be resized, do not resize, and just validate before upload etc etc" (And still include helpful engineering notes) 

Although, I do have a question about resizing large images _and then_ validating the size. If resizing a large image always resulted to a smaller image file, so why validate if we already know it's going to be smaller than 5MB?

Also for clarification, looks like the story said 2MB not 5MB. I think we did 5MB coz it was already in the html validation. Which one should we stick to?

chrisdolendo:
@mluedke2 let me know if you want to continue pairing on this, it looks like the logic is a little more intricate with resizing, validating, and selecting which kinds of files we should resize. I wonder about browser support as well if uploading will break if the browser doesn't support resizing. Might be good to test.

kaptron:
yeah I think we should sync up on this, in terms of what are the intentions of the feature, explaining how image resizing already works on the backend, what are the limitations (e.g. why we still want to limit to 5MB because we can't guarantee that resizing will be possible, particularly for PDF). To me it seems like the benefit of doing the in-browser resizing is mainly to reduce the upload size + time (like it's great that the server will optimize the image _after_ upload, but if you have a 4 MB image you're trying to upload from a slow connection then it'll go a lot quicker if your browser is able to resize it down to 1 MB first)

Anyway why don't we chat about it post-standup.

creativecoder:
To add some context for my requested changes: the purpose of this story is to prevent users from getting stuck when trying to upload a proof of preference file. This was frequently happening before when the upload limit was set to 2mb and people tried to upload photos from their phones without resizing them. This is only going to continue being a problem as phone images get larger and larger (mo' megapixels is better, yo!)

So I believe ideal functionality from a user's standpoint is
1. User uploads whatever image they have
1. We resize, if at all possible, it to something manageable to upload with enough detail to read any documents so that don't have to worry about that themselves.
1. If we can't resize it and if it's still too big, we trigger a validation error (if we have to)

@chrisdolendo You're right, resize before validation should have been in the acceptance criteria, not just the engineering notes--I apologize for that oversight. If you think we need to pull that out into a separate story, we can talk about that.

We'll talk more after scrum.

chrisdolendo:
bitmoji

mluedke2:
Alright. I added a check for 5MB after the resizing happens, in case it doesn't actually happen. In the case of PDF files, the resize function still "runs" and passes the file back to the callback, just unchanged. I tested on iOS in Firefox and Safari with large files and the behavior is the same as on Desktop.

However, I had a weird edge case where my proof file got uploaded as `liveInSf - null.pdf` (instead of `liveInSf - Electric bill.pdf`). See this Application and scroll down to see it. I can't figure out how to reproduce though-- other PDF proof uploads seem to have correct filenames. I believe it was an Electric Bill. https://sfhousing--qa.cs51.my.salesforce.com/a0o4B000000h7ju

Any objections to me marking my current story as delivered, or should we try to chase this down while looking at file uploads?

chrisdolendo:
@mluedke2 Chase it down! lol jk. Not sure, maybe we can timebox 30 mins to pair on it - might be helpful to have some fresh eyes. I'm prioritizing finishing a feature now but will review in the afternoon. Awesome!

kaptron:
@mluedke2 @chrisdolendo I feel like I've seen that `null` in the filename before... so I realized that there was one fix we made which is only on `multifamily`, which is that you are forced to choose a document type before you upload the file, and then you can't change it afterwards -- this is because the document type gets stored with your file at the point of upload, so it can be buggy if those inputs are decoupled as they are on `master`. However when I tested out that theory I was able to get `liveInSf - .jpg` but it didn't say `liveInSf - null.jpg` 🤔 

Anyway, I will at least file a bug to capture this. And @mluedke2 yes feel free to mark as delivered, as I don't think the `null` is related to this story and was probably a pre-existing bug.

mluedke2:
Sounds good-- I was wondering if it was a separate bug. I've marked the story as `Delivered` in Pivotal now.

chrisdolendo:
This looks great @mluedke2!

PR #823, opened 2017-07-29
Features/full/Google Page Speed optimizations #147236379

kaptron:
this branch is combining all the page speed optimization stories:
 - Optimize Images: #147236379
 - Serve Static Assets from a CDN: #144250921
 - Eliminate blocking CSS: #147236063


app/helpers/application_helper.rb
```diff
@@ -1,3 +1,19 @@
 # View helpers that should be available application-wide.
 module ApplicationHelper
+  def image_asset_paths
+    asset_paths = {}
+    if Rails.env.development?
+      Rails.application.assets.each_file do |f|
```
chrisdolendo: Cool~ What's the difference between `.assets` and `.assets_manifest`?

kaptron: @chrisdolendo had to learn this myself but `Rails.application.assets` is something that is available during development, it's basically a list of all your assets (which aren't "precompiled" in development") however in prod it works differently, you have to refer to the `assets_manifest` which is like a record of all the assets that have been precompiled. In both cases the `image_asset_paths` result is the same, which is an object with keys being the name of the file and values being the hashed asset path, so we can reference this in JS:

```javascript
var IMAGE_ASSET_PATHS = {
 "arrow-down.png": "/assets/arrow-down-a443598bdf41cfdc56f7c3bab8ccf69425e5e4ebbaa47dd4a217ad05201f2cbb.png"
}
```

creativecoder:
@kaptron I noticed that no one is seeing our os specific favicons because they are tucked away in a subfolder. Can we move these to be served from the root? Otherwise we can add metatags for each one that points to the existing location.

creativecoder:
@kaptron Nevermind, my bad, looks we already have the meta links.

PR #825, opened 2017-08-01
Features/full/optimize assets for pagespeed #144250921

kaptron:
continuation of #823 so using that as base branch


kaptron:
merging this branch into #823 to combine into one review app

PR #826, opened 2017-08-01
Features/full/salesforce error timeouts #147286031

chrisdolendo:
checked out from matt's modal popup alerts branch


chrisdolendo:
Feature is working, currently trying to figure out failed tests -- it's saying `Faraday client error missing`. Seeing it latest commit will fix.  Will continue tomorrow.

app/assets/javascripts/config/angularProviders.js.coffee
```diff
@@ -11,17 +11,25 @@
   $httpProvider.defaults.headers.common["Content-Type"] = "application/json"
   $httpProvider.defaults.headers.get = {}
 
-  $httpProvider.interceptors.push ["$location", "$rootScope", "$q", ($location, $rootScope, $q) ->
-    success = (response) ->
-      response
-    error = (response) ->
-      if _([400, 401]).includes(response.status)
```
creativecoder: @kaptron @chrisdolendo Did we loose 400/401 handling here? Or did something else change that we don't expect those error codes anymore?

kaptron: @creativecoder it was actually pretty silly, @chrisdolendo and I were digging into this code to setup the other error code handling, and realized the `success / error` handlers weren't actually doing anything! (the format was incorrect, or old) We're pretty sure it came over with some of the very first DAHLIA commits where we took some of the Angular config setup from CCP, and hadn't been touched since then. 

So first of all we set up the correct `responseError` handler, and secondly our angular token auth libraries are already handling `400/401` errors just fine.

creativecoder: This PR makes so much more sense now! Thanks for the context.

PR #828, opened 2017-08-03
Features/full/eliminate blocking css #147236063

kaptron:


kaptron:
merging this into #823 to combine all Google Page Speed optimizations

PR #830, opened 2017-08-03
Features/full/salesforce timeout alerts #147286031

chrisdolendo:


chrisdolendo:
The protractor tests seem to keep failing coz of the timeout :joy:

chrisdolendo:
Removing the modals label because this is only using the alert box since Matt's modal story hasn't been merged yet. Will switch to modal after it has been merged.

kaptron:
unfortunately there are some problems with our approach to rescue from these timeouts, and that's why semaphore is failing... we will need to adjust the way we're doing this. The problem is that `Faraday::ClientError` is the same exception that also gets thrown if there is a Restforce error (e.g. invalid listing ID, etc). Ran into a couple of dead ends my first quick attempts of trying to move this around, and will have to take another look.

kaptron:
@chrisdolendo @creativecoder finally got this in a good place! in the end it is adding better handling to more than just timeouts, for example when hitting the listing API with an invalid ID, that was previously just being rescued by the SalesforceService and returning a `200 OK` but with empty data, even though behind the scenes Salesforce was actually throwing an error. That has been changed to return a `404` (API response) and then handled properly by Angular. 

Likewise, learned an interesting lesson about `408 client timeout`, which is that your browser may treat that as "this appears to be my fault, let me try again", at least that's what Chrome does and someone else experienced here: https://github.com/angular/angular.js/issues/4806
It was really weird because I couldn't figure out why it would keep re-trying the request over and over, but apparently it is native browser behavior. So I changed that to a `504 gateway timeout` which seems more appropriate. 

TL; DR: All timeouts and Salesforce errors are now handled gracefully! The user will get an error alert, and then we log the error (i.e. to papertrail)

```
2017-08-16T17:35:05.590253+00:00 heroku[router]: at=info method=GET path="/api/v1/listings/foofoo.json" host=dahlia-qa-pr-830.herokuapp.com request_id=3c3e9a62-e6b2-4da7-be7d-432238470681 fwd="173.239.228.88" dyno=web.1 connect=1ms service=1253ms status=404 bytes=684 protocol=https
2017-08-16T17:35:05.585127+00:00 app[web.1]: << API Error >> Faraday::ClientError, APEX_ERROR: System.StringException: Invalid id: foofoo
2017-08-16T17:35:05.585138+00:00 app[web.1]: 
2017-08-16T17:35:05.585140+00:00 app[web.1]: Class.ListingService.getListingByIDs: line 446, column 1
2017-08-16T17:35:05.585140+00:00 app[web.1]: Class.ListingService.getListingDetailByIDs: line 547, column 1
2017-08-16T17:35:05.585141+00:00 app[web.1]: Class.ListingService.getListingDetailByID: line 541, column 1
2017-08-16T17:35:05.585142+00:00 app[web.1]: Class.ListingDetailAPI.doGet: line 34, column 1
2017-08-16T17:35:05.585561+00:00 app[web.1]: Completed 404 Not Found in 1157ms (Views: 0.2ms | ActiveRecord: 0.0ms)
```

creativecoder:
@kaptron Awesome! This feels so much better, thanks for seeing it through. Very good to know about 408 vs 504

chrisdolendo:
woah that explains it! learned a lot thanks @kaptron

PR #834, opened 2017-08-10
Features/qa/google translate outside apps #145890855

kaptron:
This is the first in the line of google translate features so it will probably serve as the base branch for the whole "mini-epic"


chrisdolendo:
Team work! Looks cool

PR #838, opened 2017-08-16
Adds radio and bus ad referral options to short form

callaghanc:


callaghanc:
Closed because of incorrect setup. New PR with correct setup will be submitted.

PR #844, opened 2017-08-17
Features/qa/Format lottery number in modal-#132710333

callaghanc:


app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -578,6 +578,7 @@ ListingService = ($http, $localStorage, $modal, $q, $state, $translate) ->
     lotteryNumber = lotteryNumber.replace(/[^0-9]+/g, '')
     if (lotteryNumber.length < 8)
       lotteryNumber = '0'.repeat(8 - lotteryNumber.length) + lotteryNumber
+    lotteryNumber
```
kaptron: @callaghanc everything looked good but I was wondering why some values were seeming to blank out the lottery number altogether -- lack of a `return` value! That's where coffeescript can sometimes get you with the implicit return.

PR #846, opened 2017-08-17
Features/full/Salesforce timeouts v2 #147286031

kaptron:


kaptron:
Rather than specifically set a different timeout for the short form `POST` method, this is setup to adjust the timeout generally for `api_post`(which is really just used by short form, and account create/update) to wait 25 seconds instead of the default amount. 

We can improve this later with: 
 - implementing better handling of submission timeouts on salesforce side (searching for session id, currently in the icebox until post-MF) 
 - refactoring our service classes into service objects (timeouts could be more easily adjusted on a per-method basis rather than being a class attribute)

creativecoder:
👍 This looks good--I like the refactoring with per method settings for timeouts.

PR #851, opened 2017-08-23
plus housing #150111282

chrisdolendo:


kaptron:
@chrisdolendo my fault, since I merged this without checking it out first -- but we actually don't want the `plus_housing.css` in our asset pipeline. Because in `application.css` we do `require_tree .` that means it includes _all_ files in the stylesheets directory (which is really just `toolkit.css`), and so it was bringing the `plus_housing.css` into our main angular app as well. Granted we could change this setting in `application.css`, but since this CSS really has nothing to do with our app, I just moved it to be inline in the head of the +housing layout so it was self-contained (pushed that to master since this PR is already closed)

chrisdolendo:
@kaptron makes sense! Best not risk it overwriting the other styles within Dahlia

kaptron:
@chrisdolendo yep, it was making some of the short form dropdowns look funny and at first I couldn't figure out why... 🤔  💡

PR #853, opened 2017-08-23
Features/qa/Timeout browser alert #149616677

callaghanc:


kaptron:
@callaghanc 👍  looks good, see my last commit 8ff2dbc I just refactored the service slightly, once `ModalService.alert` started getting some extra params I turned it into an `opts` object so that it's more clear e.g. `ModalService.alert('hi', {nativeAlert: true})` instead of `ModalService.alert('hi', null, true)`

kaptron:
closing to reopen without `qa`

PR #854, opened 2017-08-24
[#149973165] - Update lottery pdf url logic

chrisdolendo:
Nothing was working during the implementation, but then upon investigation, I noticed that the lottery modal footer was still referencing `lotteryBuckets` when the Service was using `lotteryBucketInfo` 😮  Anyway, fixed that naming and then the implementation worked as expected. Also, searched the codebase for any other `lotteryBuckets` reference in case any was left out. This should be the last of it.


kaptron:
@chrisdolendo oh interesting! must have been some additional changes from the new lotteryBuckets setup w/ multifamily that we didn't catch? If that's the case, I would recommend that we move this into the multifamily release, if the feature was actually broken.

kaptron:
👍  😊

PR #861, opened 2017-08-29
Features/mobile short form nav bar #121621661

chrisdolendo:


chrisdolendo:
Updated to refactor. 

Maybe someone can chime in but focusing on `short-form-wrapper` was the closest thing to focusing straight to the application, and it looks like the image below:

screen shot 2017-08-29 at 12 03 40 pm

Note that `short-form-wrapper` is the container _below_ the app-header/nav, yet it shows the nav part of the header, but doesn't include the blue header naming the building. When I try to make it focus on anything above the form, it shows website header. I kept it as you see here but I welcome any other suggestions if it's necessary to show the app building title.

kaptron:
@chrisdolendo for what it's worth, on my iPhone (6S) Safari and Chrome, it does scroll directly below the short form nav (i.e. "First, what's your name?") appears exactly at the top, and it looks good to me.

creativecoder:
@chrisdolendo I just tried this on an Android device in browser stack and I think it scrolls a bit too far. The mockup has the Listing Name at the top of the screen, but I see the form header. 
what i expect to see
what i actually see

kaptron:
@creativecoder thanks for catching that, I hadn't noticed the mockup in the story comments but that makes sense as a better place to scroll to. I've updated this, and it's working well for me on responsive chrome, iOS and android (browserstack). 

Also chatted quickly with Jesse... this is one of those "designed mobile-first" features that I actually like the behavior on desktop as well (I sometimes find myself scrolling down when paging through the form, even on desktop). We agreed that it's a nice global feature and doesn't need to be limited based on 768px width, so I've updated that as well.

chrisdolendo:
@kaptron thanks for the update!

chrisdolendo:
Finally get to merge a PR! lol

chrisdolendo:
oh whoops, my bad.

PR #866, opened 2017-08-31
Bugs/full/toggle more fix #150566235

kaptron:


chrisdolendo:
@kaptron - ah yes. I was thinking about the significance of that function, cool!

PR #868, opened 2017-09-05
Features/qa/upload custom preference #147142889

chrisdolendo:
WIP


kaptron:
@chrisdolendo I'm wondering why you had to pull the back page logic out of `ShortFormNavigationService`, since now that logic is a little disjointed (sometimes it comes from nav service, and sometimes from the multiple back buttons with ng-ifs). Perhaps I can take a look at how that same logic could be refactored so it's all in the same place again

chrisdolendo:
@kaptron, I had started down that path, but it's because the Nav service function (`Service.backPageState`)  that determines the back page returns a `string` for the ui-sref. And the `custom-proof-preferences` back logic are dynamic based on the `prefIdx` param, and a lot of other things. It wasn't as simple as returning a string. Maybe there is a way we can extend the `Service.backPageState` function to accommodate the logic I placed in the `ShortFormAppController`. Let's dig in when we pair!

chrisdolendo:
closing and re-opening on `full`

PR #869, opened 2017-09-06
Chores/redirect translations from previous route #150851553

creativecoder:


creativecoder:
@kaptron could use your help setting this up to work correctly with Cloudfront

creativecoder:
@kaptron This was triggerin 500 errors for any missing/malformed translation files (like `/translations/locale-zz.json` so I modified to account for that. Also seemed redundant to check dev environment in application controller and application helper, so refactored that slightly.

Please review and let me know if you see anything else that should be improved or added before merging. Thanks!

kaptron:
@creativecoder ah good call -- just tested it out (dev + on heroku), and it all looks good to me! 👍

PR #882, opened 2017-09-13
Features/full/disable features mf release #150937154

chrisdolendo:
branched from master, PRed to prod


chrisdolendo:
closing - reopened one that's co from prod

PR #883, opened 2017-09-13
Features/full/prod disable features temp #150937154

chrisdolendo:
opening this version as well - branched from prod, PR-ed to prod

### DIFF
--- a/app/assets/javascripts/account/templates/sign-in.html.slim
+++ b/app/assets/javascripts/account/templates/sign-in.html.slim
@@ -1,31 +1,32 @@
 .dash-container.tall
   section.row.collapse
     .medium-6.medium-centered.columns.fixed-width
-      back-to-application-link
+      / back-to-application-link
 
-      form#form-wrapper.dash-card novalidate="" name="form.signIn" ng-submit="signIn()"
-        header.dash-header
-          span.header-badge
-            span.ui-icon.ui-static.ui-medium.i-color
-              svg
-                use xlink:href="#i-profile"
-          h1.dash-title.no-margin Sign In
+      p We are in process of adding new features to the site. Sign in is not available right now. Please try back later.
+      / form#form-wrapper.dash-card novalidate="" name="form.signIn" ng-submit="signIn()"
+      /   header.dash-header
+      /     span.header-badge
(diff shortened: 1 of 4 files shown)


chrisdolendo:
I notice that this breaks a bunch of E2E tests as a result. My commits are most patterns, should I just  comment the breaking tests or let them fail? :thinking:

kaptron:
@chrisdolendo no need to change the tests around, you can let them fail 😈  
I think they're actually only failing because `production` and Salesforce Full don't play together nicely, Full is setup for multifamily, but `production` is not yet. So nothing to do w/ your code.

plus since this is going to get reverted anyway, no need to worry about it. 

lastly I believe your Q about the background, this is the same thing that gets fixed in #857. We didn't schedule that in the release -- but not a big deal, just a minor visual thing.

kaptron:
👍  looks good to me. 

also it was nice of you to leave all the notes of where to revert stuff, but we can literally just `git revert` this whole change, so no biggie either way. 😸

chrisdolendo:
true!

creativecoder:
👍

Just a reminder: you don't need to specify `/full/` anymore. Everything defaults testing on full unless there is `/qa/` in the branch name, then it will test on QA.

Let's think about what (if any) part of this functionality should live behind a feature flag in the future. I can imagine us wanting to do a similar thing again the next time we freeze caching for a joint Salesforce/web app deploy. (I'll make a pivotal story.)

kaptron:
good point re: feature flags. This use case was pretty specific though, in that pref refactor specifically had breaking changes e.g. with viewing your past applications and past lottery results, so I'd be wary of leaving too much of this in there with conditionals (i.e. _over_-configuring things). 

That being said, the nicely disabled sign in page seems like a good potential one to feature flag.

chrisdolendo:
:+1: thanks for the reviews!

Msol just accepted this so it's ready to go whenever.

kaptron:
@chrisdolendo you can go ahead and merge this! that will auto-deploy it to `dahlia-pre-prod` and then later after 5pm we will push it live.

kaptron:
@chrisdolendo it was getting to be time, so I merged it.... can't say I didn't give you the chance to press the button :grin:

bitmoji

PR #884, opened 2017-09-13
Bugs/angular routes trailing slashes #151032576

kaptron:


creativecoder:
Looks like this is accepted. One question: is there any analytics impact on having 2 routes (one with trailing slash, one without) go to all our pages? If so, we should redirect in some way? (WordPress always redirects you to a route with a trailing slash, for example, unless it's a file)

kaptron:
Yeah it's an interesting one, [strictMode](https://ui-router.github.io/ng1/docs/latest/interfaces/url.urlmatcherconfig.html#strictmode) seems to be sort of all or nothing, you either allow slashes or you don't (in the `strictMode(true)` case it doesn't even recognize the route).

I'm sure there's some way we could hack our way into angular-ui router logic to have it still recognize the route but then always redirect to un-slashed, but I'm not sure of an easy way to do that app-wide off the top of my head. I'm wondering what the downside is... seems like the effect on SEO would be low (like who/what would be indexing the pages with slashes, we just want the URLs to work without breaking, but we're not actually serving up any of those routes/urls ourselves)

creativecoder:
I'm more worried about duplicates in analytics reporting, like discussed here: https://magazine.getelevar.com/how-to-fix-the-dreaded-duplicate-url-in-google-analytics-b1fdc8ff87f8

There's a suggestion in ui-router docs. States it's for older versions of UI router but maybe we could use it for this issue as well? https://github.com/angular-ui/ui-router/wiki/Frequently-Asked-Questions#how-to-make-a-trailing-slash-optional-for-all-routes

kaptron:
Ah, makes sense... Yikes their suggestion says that we'd have to switch everything the other way around, and now we _always_ use trailing slashes. However perhaps that hack points us in the right direction... Either that or I wonder which version of ui-router we could upgrade to? We're on 0.2.x, there is now 0.3.x, 0.4.x and 1.0. I'm pretty sure at least one of those versions introduces some breaking changes to the way routes are constructed, but might be worth a look at some of the interstitial ones...

PR #885, opened 2017-09-13
Chores/Run database migrations before deploy #150753047

creativecoder:


creativecoder:
@kaptron Tested twice by running `rake db:rollback` on heroku pr app and then pushing a dummy commit. Works as expected!

kaptron:
excellent... I'd say let's merge!

PR #886, opened 2017-09-15
Bugs/selecting yes on live work #151111920

kaptron:
found this while doing QA testing yesterday

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
@@ -386,8 +386,6 @@ ShortFormApplicationService = (
     if type == 'liveWorkInSf' || type == 'all'
       Service._updatePreference('liveInSf', Service.liveInSfMembers())
       Service._updatePreference('workInSf', Service.workInSfMembers())
-    else if type == 'workInSf'
-      Service._updatePreference('workInSf', Service.workInSfMembers())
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
```diff
@@ -393,8 +393,6 @@ ShortFormApplicationService = (
     if type == 'liveWorkInSf' || type == 'all'
       Service._updatePreference('liveInSf', Service.liveInSfMembers())
       Service._updatePreference('workInSf', Service.workInSfMembers())
-    else if type == 'workInSf'
```
chrisdolendo: Not sure why this was deleted? Is it because it get's called within `if type == 'liveWorkInSf' || type == 'all'`?

kaptron: it got removed because it didn't seem like we're ever calling it for just `workInSf` any more... in fact, looking again, it actually seems like we really just do it for `liveWorkinSf` or `all`. This is one of those things that could probably be refactored a bit... `checkPreferenceEligibility` gets used as a page callback which then calls this function with `refreshPreferences('liveWorkInSf')`. Otherwise various things in the Service just call `refreshPreferences('all')`. 

I think at some point we were like, hmm if you change your address/etc, it might change around several different requirements... no harm in checking out all your preferences and making sure you're still eligible (the blunt force approach).

PR #891, opened 2017-09-20
Features/navigating back and forth accordions #149904145

kaptron:
note that this hijacks the `keepScrollPosition` directive to do one thing related to `ListingService`, since it seemed like the best place to add this same logic regarding `usedBrowserNavigation` -- if we wanted we could rename that to be something more like `browserNavigationHandler` since that is more generically what it is doing now.


creativecoder:
A couple thoughts on this:

It seems like it breaks unit features in some way:

1. Go to /listings
2. Navigate to a listing (call this listing 1)
3. Click on Browse in the nav menu
4. Navigate to another listing (call this listing 2)
5. Click on Browse in the nav menu
6. Navigate back to listing 1

Result: Unit features expander is missing for listing 1

Secondly, the behavior is inconsistent with the accordion expander, which doesn't clear accordion state on every window.location change, only when you refresh the page in browser--I wonder if it would make more sense if these were consistent (story description is confusing, here).

kaptron:
@creativecoder that bug is weird, not sure what was going on. In any case @callaghanc and I paired on this and refactored it a bit and agreed that the state should be remembered based on which listing you viewed, and not whether or not you used browser navigation (also got Jesse's buy-in on this). I've amended the AC of the story to fit this description.

chrisdolendo:
@creativecoder since you started the review on this, could you do an official review with `approve` or `request changes`?

app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -60,10 +62,7 @@ ListingController = (
     $scope.showApplicationOptions = !$scope.showApplicationOptions
 
   $scope.toggleTable = (table) ->
-    $scope["active#{table}Class"] = if $scope["active#{table}Class"] then '' else 'active'
-
-  $scope.isActiveTable = (table) ->
-    $scope["active#{table}Class"] == 'active'
```
creativecoder: `isActiveTable` is still used in `_features.html.slim:29` https://github.com/Exygy/sf-dahlia-web/pull/891/files#diff-4bdfc1842a1371d26e8dbea081644fd1L29

creativecoder: Fixed in https://github.com/Exygy/sf-dahlia-web/pull/891/commits/a63ca3a7f35bc6c72857be7e53b8964c53542ec7

creativecoder:
@callaghanc I modified so that the "What to Expect" expander on the listing sidebar remembers the expanded/unexpanded state for each listing separately, rather than sharing the same state for all listings. This keeps behavior consistent with the rest of the expander behavior. https://github.com/Exygy/sf-dahlia-web/pull/891/commits/b3ea79efed37960e2bbf9c591c452411cd4e92d0

Let me know if you think that should not be the case for some reason.

Otherwise, this is good to merge.

callaghanc:
@creativecoder No that seems right, seems like it was an oversight not to also use the listing ID on the "What to Expect" sections. I'll merge this in.

creativecoder:
👍

PR #894, opened 2017-09-21
Bugs/resetting location preferences #151313508

kaptron:


creativecoder:
I didn't test on review app but the code change looks legit based on what we did in the original refactoring. Bummer that we missed this case :(

PR #895, opened 2017-09-22
Features/bypass cache with listing preview #150861671

kaptron:


chrisdolendo:
@creativecoder before I merge, does any of your refactoring go against this?

creativecoder:
@chrisdolendo Thanks for flagging this. The listing controller changes will need to be updated in that refactor branch. I can handle that when I merge this code from master.

chrisdolendo:
okay @creativecoder will merge now then.

PR #889, opened 2017-09-18
Features/full/email draft link #135807091

chrisdolendo:
New branch checked out from reopened `features/full/allow-draft-to-start-at-previous-page-#142887053` feature branch, PR #888. Cherry-picked feature-relevant commits from PR #824.

PR #900, opened 2017-09-26
Features/email draft link #135807091

kaptron:
see #889 for the original history of this whole feature. Most of it has now been merged because it was a continuation of #888 (which has been merged/accepted). This PR now just contains the specific "emailing link to the saved draft" part of the feature, as that still needs to be accepted.


callaghanc:
@kaptron I was testing the review instructions on the Pivotal story for this and the instructions worked, but when I used the link in the email to re-open my draft, the application page I landed on looked odd, see screenshot. It was the right page, the address confirmation page, but it didn't have any address in it. I went back to my account page, to my saved applications, and opened the draft that way, and doing that brought me to the address confirmation page of the application with my address correctly shown on the page. 
screen shot 2017-10-24 at 11 30 29 am

kaptron:
good catch @callaghanc! Those "verify address" pages are somewhat transient and not really a good place to save your progress -- it's expecting to display the address that was temporarily pulled into the GeocodingService. I've updated it so that if you "save and finish later" from one of those pages, it will actually save that you were on the previous page (e.g. bump you back to "contact" or "household"). Not the perfect solution, but hopefully good enough for this unique case.

PR #903, opened 2017-09-27
Features/update feedback form #151024522

chrisdolendo:


kaptron:
@chrisdolendo I wonder if version + footer should co-opt the NavController or if $scope.feedbackUrl should just be added to SharedController (which already wraps the whole page)? NavController seems like it's mainly handling navigational things, though I can see how the version bar + footer are somewhat related to the nav.

chrisdolendo:
@kaptron sharedcontroller seems legit. i'll migrate that scope :) thanks for the review!

chrisdolendo:
@kaptron thanks for pinging this, updated and ready for re-review

PR #905, opened 2017-09-27
Bugs/Fixes Choose Draft + loadAccountApplication functionality #151486626

kaptron:


db/schema.rb
```diff
@@ -48,7 +48,6 @@
     t.string   "listing_id"
     t.string   "document_type"
     t.integer  "user_id"
-    t.integer  "preference"
```
creativecoder: @kaptron Is this schema change right?

kaptron: @creativecoder ah yeah didn't leave an explanation on this, but yeah this was a correction for a field that's already been removed in MF (i.e. if you run all the migrations in order it is removed). I had fudged it to leave the field in my local db for easier switching between MF and pre-MF, so this is just setting it back to the way it should be.

creativecoder:
Code diff looks good. Glad to see an e2e test for this!

See my pivotal comment, getting an intermittent alert that shouldn't be there as part of the choose draft flow.

PR #914, opened 2017-10-04
Bugs/demographic survey custom answer #149847962

kaptron:
preferred fix would be that salesforce titled these picklists consistently, see comment: https://www.pivotaltracker.com/story/show/149847962/comments/180555342


kaptron:
did also add E2E tests because the demographic survey was not being tested very thoroughly.

chrisdolendo:
:+1:

PR #915, opened 2017-10-05
Bugs/community screening bypass #145653375

kaptron:


chrisdolendo:
:+1: tested and it works!

kaptron:
@chrisdolendo good point... some of the logic is in angularRoutes which currently doesn't have any test coverage. I added a chore in our icebox to add tests to that, I've been meaning to do that. So that's one place to start. Otherwise, could be a good fit for E2E but that would mean either: adding a 2nd listing w/ community building settings to our post-refresh + E2E setup process, or updating 280 Fell to be a community building, which will affect all our other E2E tests which start with "jump to the first page of the application" since that would now kick you out to the community screening page. 
So it seems not worth the hassle, for now...

PR #917, opened 2017-10-05
Bugs/skip to main content safari #150678472

kaptron:
Turns out it was just the `blur()` we were doing that was making it not work in Safari (note Firefox is still another separate issue). I'm not noticing anything visual/bad about the outline we were trying to remove by blur, you can see the focus is on the body by clicking "skip to main content" and then scrolling all the way down the page to barely see that the blue outline is at the bottom, but this seems like a small price to pay to have the feature work cross-browser.


chrisdolendo:
:+1:

PR #920, opened 2017-10-10
Chores/log app attachment params #151018487

chrisdolendo:


app/services/salesforce_service/short_form_service.rb
```diff
@@ -84,6 +84,7 @@ def self.attach_file(application, file, filename)
         ApplicationMemberID: _short_form_pref_member_id(application, file),
         ApplicationPreferenceID: _short_form_pref_id(application, file),
       }
+      logger.info "Attach job params: #{body}"
```
kaptron: @chrisdolendo running in the job it appears to want `Rails.logger.info` otherwise `logger` is undefined -- I think this is because our Service classes do not inherit from any ActiveRecord/ActiveXXX

other than that, I would just say let's make the log message a little more direct like: 
`"ShortFormService attach_file params: #{body}"`

creativecoder: How about mirroring our existing logging info as closely as possible? `Api::V1::ShortFormService.attach_file Parameters: #{body}`

chrisdolendo:
@kaptron thanks for the review I shall add a quick spec #stublife

chrisdolendo:
@kaptron - test added for `attach_file`, codeclimate is happy :)

PR #922, opened 2017-10-11
Chores/angular ng if optimizations #146883395

kaptron:
Sprinkles lots of `::` in templates, mainly for browse listings / single listing where bindings can be one-time instead of having watchers. 

Did not find as much in short form that required this, as a lot of those watchers are actually appropriate (based on dynamic input, etc), and there are generally less watchers per page so it's less of an issue.


chrisdolendo:
Agreed like that it was just on the listing page since short form will need refreshing of some data. Also flagging this is related to this [PR#929](https://github.com/Exygy/sf-dahlia-web/pull/929). :+1:

kaptron:
merging this optimization chore, thanks for the review @chrisdolendo!

PR #923, opened 2017-10-11
Bugs/email validation #151134233

callaghanc:


creativecoder:
@callaghanc Please update PR name using our standard format

callaghanc:
@creativecoder I'll look into block regular expressions and add the reference. If you check out the engineering notes on the Pivotal story, there's an explanation of why I'm using that regex.

callaghanc:
Oh didn't know about the CoffeeScript multiline regexes, good to know, thanks! Added e2e test. Also added regex validation to more email fields.

kaptron:
looks like everything @creativecoder requested is in there now 👍  story has been approved and product team wants in the release -- will check in at standup otherwise this should be ready to merge.

PR #925, opened 2017-10-12
Chores/uri encode listing id #151411944

chrisdolendo:


chrisdolendo:
@kaptron agreed, that will make it more dry when we get to instantiate a service class.

PR #928, opened 2017-10-16
Chores/rescue file resize #151425961

chrisdolendo:
Rescues png conversion and image optimization if there are any errors and continues file upload.


app/models/uploaded_file.rb
```diff
@@ -4,15 +4,21 @@ class UploadedFile < ActiveRecord::Base
...
-        # convert png -> jpeg
-        image = MiniMagick::Image.new(tempfile_path)
-        image.format 'jpg'
-        attrs[:content_type] = 'image/jpeg'
+    begin
+      unless attrs[:content_type] == 'application/pdf'
+        if attrs[:content_type] == 'image/png'
+          # convert png -> jpeg
+          image = MiniMagick::Image.new(tempfile_path)
+          if image.valid?
```
chrisdolendo: @kaptron did you implement the conversion/optimization feature? I actually have a Q about this, how come we don't use the image variable again past conversion and use tempfile_path?

chrisdolendo:
Code climate says X. Any recommendations on a useful test for this? the `uploaded_file_spec` tests the `create` method but looks like the `create_and_resize` method involves stubbing a bunch, I wonder how useful it would be 🤔

app/models/uploaded_file.rb
kaptron: @chrisdolendo ah -- `image` is just a MiniMagick object that represents the converted image that lives at `tempfile_path` (and has access to MiniMagick methods like `.valid?` and `.format(xxx)`. So once you do the conversion, the image at `tempfile_path` has been modified, and so when you want to refer to the actual file in the filesystem, you refer to that.

kaptron:
@chrisdolendo well I took a first stab at this -- I actually was curious about shoring up the tests myself and how to properly mock/stub things in this fashion, and you're right it needed to stub a lot of what's going on in this method. Unfortunately still under the 50% threshold because it's not testing the 'png' + `MiniMagick` part of the code, since I used `content_type: 'jpg'`. Why don't you take it from here, switch it to 'png' and mock the `MiniMagick::Image` calls, see if you can finish that up. It's one of those ones that maybe feels a little silly when it's mostly stubbing everything, but it's still good to have the test in place, that way if anything changes the code to `create_and_resize` the test will be there to point that out (I guess?)

app/models/uploaded_file.rb
chrisdolendo: makes sense thanks @kaptron!

chrisdolendo:
@kaptron I agree with you, I'll go ahead and do that :)

chrisdolendo:
test for this is in my todo today

chrisdolendo:
@kaptron I added a test to this (finally), and the checks passed. 🎉  Ready for your review.

PR #931, opened 2017-10-17
Features/remove geocoder fail emails #151690364

chrisdolendo:


app/mailers/emailer.rb
```diff
@@ -134,14 +134,7 @@ def setup_geocoding_notification
...
   def admin_email
-    # all heroku apps have Rails.env.production
-    # but ENV['PRODUCTION'] is only on dahlia-production
-    if Rails.env.production? and ENV['PRODUCTION']
-      # if listing has_nrhp_adhp it goes to the set of admins that includes HBMR staff
-      @has_nrhp_adhp ? 'dahlia-admins-hbmr@exygy.com' : 'dahlia-admins@exygy.com'
-    else
-      @has_nrhp_adhp ? 'dahlia-test-hbmr@exygy.com' : 'dahlia-test@exygy.com'
-    end
+    @has_nrhp_adhp ? 'dahlia-test-hbmr@exygy.com' : 'dahlia-test@exygy.com'
```
kaptron: @chrisdolendo hm I don't think this is what we wanted to do here? this will _always_ send emails to `test

we want to keep the whole `ENV['PRODUCTION']` piece but just remove the `-hbmr` ones, so in a nutshell: 

` if prod 'dahlia-admins' else 'dahlia-test' `

chrisdolendo: my bad @kaptron, thanks for catching that.

PR #934, opened 2017-10-18
Chores/e2e rent burdened tests #144587445

kaptron:


spec/e2e/step_definitions/short_form_steps.coffee
```diff
@@ -493,29 +529,11 @@ module.exports = ->
...
-
-  @Then 'I should see the Live Preference', ->
-    livePref = element.all(By.cssContainingText('strong.form-label', 'Live in San Francisco Preference')).filter((elem) ->
-      elem.isDisplayed()
-    ).first()
-    @expect(livePref.isPresent()).to.eventually.equal(true)
-
-  @Then 'I should see the Work Preference', ->
-    workPref = element.all(By.cssContainingText('strong.form-label', 'Work in San Francisco Preference')).filter((elem) ->
+  @Then /^I should see the "([^"]*)" preference checkbox$/, (prefName) ->
```
chrisdolendo: Really like the refactoring of this! if you got the time I would love to know what `([^"]*)` means -- if that's some type of regex?

kaptron: Yeah that's what allows you to pull out a variable e.g. `prefName` in this case and have it be dynamic. I think the regex basically means `everything contained within double quotes` so you can say `Then I should see the "assistedHousing" preference checkbox` and it will pull out `assistedHousing`. Very useful!

PR #936, opened 2017-10-20
Bugs/catch wrong format root request #151452799

callaghanc:
Applies the same format constraints that we already have on "*path" to the root route. This will make non-HTML requests to the home page trigger a 404 error instead of a 500 error. There wasn't any clear preferred alternative 4xx error for wrong formats in requests, so I decided to keep the existing 404 for wrong formats behavior that we already had and bring the root route into line with that.


kaptron:
looks good to me! 👍

PR #937, opened 2017-10-23
Features/salesforce error in alert #151138937

chrisdolendo:


app/controllers/api/v1/short_form_controller.rb
```diff
@@ -64,14 +64,16 @@ def show_listing_application_for_user
   def submit_application
     response = ShortFormService.create_or_update(application_params, applicant_attrs)
     if response.present?
-      attach_files_and_send_confirmation(response)
-      if current_user && application_complete
-        delete_draft_application(application_params[:listingID])
-      end
+      process_submit_app_response(response)
```
chrisdolendo: this was refactored due to cyclomatic complexity alert by rubocop -- other ways to refactor the rescue instead of this is welcome :)

app/assets/javascripts/config/angularProviders.js.coffee
```diff
@@ -13,7 +13,7 @@
         # This is set up to universally capture HTTP errors, particularly 503 or 504, when a bad request / timeout occurred.
         # It will pop up an alert and stop the spinning loader and re-enable short form inputs so that the user can try again.
         responseError: (error) ->
-          if error.status >= 500
+          if error.status >= 400
```
kaptron: @chrisdolendo why `400` error? I don't like the `error.status >= 400` because then for example a `404` will get an "uh oh!" which doesn't seem right. `4xx` error means "the client did something wrong", whereas `5xx` means "the server did something wrong" which is what seems to be the case here. So I think it should stay in the 500 range.

callaghanc: this question has been resolved

creativecoder:
Something about this is not working. I did quick test, by trying to save an application with a really long address (which triggers an apex error about the field being too long), but got a generic error alert with the api message "not found".
screen shot 2018-02-23 at 11 04 57

callaghanc:
@creativecoder Looks like since this PR was created and the story was tested and accepted in Pivotal, there's been a change in error handling in another branch that was merged to master, and master has been merged into this branch, and now a new option, `capture`, is required in the `opts` for `render_error` in api_controller.rb in order for the Salesforce error message propagation to work again. I will follow up on this with @chrisdolendo as needed.

creativecoder:
Makes sense. Seems like having `capture` default to `true` would be a good idea.

chrisdolendo:
TO DO: Retest, fix, redeliver to anyone

callaghanc:
@creativecoder Added the error rendering refactor, ready for re-review.

PR #939, opened 2017-10-24
Features/lottery buckets timeout #151458539

chrisdolendo:
TASK BEFORE MERGE: Revert listings_controller.rb code before merging.


chrisdolendo:
Note: The failed checks are because of a temp code change I made to re-create a salesforce error for `lottery_buckets`. The tests will pass when reverted before merging.

PR #942, opened 2017-10-27
Chores/refactor shortform data service #151726771

kaptron:


kaptron:
Done and ready for review, tests passing 😅  -- Would be good to get some other engineer eyes on this as it's an important part of the application! Some notes on implementation:
 - Creates shared `WHITELIST_FIELDS` to be used by `format` and `reformat` basically meaning, these are the fields that we copy 1:1 from our angular applicant <-> salesforce object because they have the same name/value. 
 - Modifies the way `formatApplication` kept passing the `application` along and modifying it in each function. Now, each function is more "functional" and just returns some output that can be attached to the application.
 - `reformatApplication` was basically already set up this way (building up parts of the object rather than passing along + modifying), so I didn't make any changes there.

PR #943, opened 2017-10-27
Bugs/submit duplicates #152157944

chrisdolendo:
- To prevent app submission dupes due to shared global retry values


kaptron:
👍  for implementation, it looks like tests are failing because of the refactored method call, hopefully an easy fix

chrisdolendo:
Yup, tests pass now :)

chrisdolendo:
not needed

PR #945, opened 2017-10-31
Features/listing unit data timeout #150277919

callaghanc:
…option


callaghanc:
This branch has an error trigger in it for testing purposes. The error trigger is causing Code Climate to fail. Once the story is accepted, I will remove the error trigger.

app/services/salesforce_service/base.rb
```diff
@@ -8,9 +8,11 @@ class Base
     class_attribute :timeout
     class_attribute :error
     class_attribute :force
+    class_attribute :unit_data_tries
```
kaptron: the whole `unit_data_tries` is purely for the fake error, right? otherwise, this looks good 👍

callaghanc: yep just for the error trigger

PR #948, opened 2017-11-08
Features/add lottery results button to my applications #152122952

kaptron:
the UX for this (particularly for the failure case where salesforce call didn't work, and we fall back to download) probably requires some discussion, otherwise the feature seems to work well.


creativecoder:
Note, request from product is to not merge this until we've handled the error case in this story:
https://www.pivotaltracker.com/story/show/152730034

chrisdolendo:
Okay to merge - need to fix conflicts first.

creativecoder:
This one can be merged--we decided not to wait for the subsequent work.

callaghanc:
ok @creativecoder, i'll resolve conflicts and merge

callaghanc:
We should merge this PR after the new modal work in my PR https://github.com/Exygy/sf-dahlia-web/pull/1011 is merged into master. Once that new modal work is in master, we can do a single merge conflict resolution in this PR that updates it to use new modal syntax. I will track the approval and merge of https://github.com/Exygy/sf-dahlia-web/pull/1011 and return to this PR to finish resolving conflicts after https://github.com/Exygy/sf-dahlia-web/pull/1011 is merged.

app/assets/javascripts/account/directives/myApplicationDirective.js.coffee
```diff
@@ -54,11 +60,51 @@ angular.module('dahlia.directives')
...
+        scope.isSubmitted() && scope.listing.Lottery_Results
+
       scope.isPastDue = ->
         moment(scope.listing.Application_Due_Date) < moment()
 
       scope.lotteryNumber = ->
-        { lotteryNumber: scope.application.lotteryNumber }
+        if scope.listing.Lottery_Results
+          html = """
+            <button class='button-link lined' ng-click='viewLotteryResults()'>
```
chrisdolendo: @callaghanc is there a reason why you have to return an html here? Can you just return the lottery number and do the ng-if on the markup for a button or to display the number?

app/assets/javascripts/account/directives/myApplicationDirective.js.coffee
```diff
@@ -54,11 +60,51 @@ angular.module('dahlia.directives')
...
+          ###
+          # NOTE: Even though we have the listing already "loaded" via API AccountController `map_listings_to_applications`
+          # this has one limitation which is that by using ListingService.listings() ("browse" API) we do not get LotteryResultsURL.
+          # So in the event that getLotteryRanking fails and we want to fall back to LotteryResultsURL, we have to call getListing.
+          ###
+          ListingService.loadListing({})
+          # have to restart the loader because the error would have stopped it
+          ShortFormNavigationService.isLoading(true)
+          ListingService.getListing(scope.listing.Id).then ->
+            # alert message: "Oops! Please download lottery results instead..." ??
```
chrisdolendo: what does this comment pertain to?

callaghanc: This is Dave's work. My best guess is that it references the uncertainty at the time about what kinds of error states or fallback options there should be in case of failure to load certain objects in this process. If you read the comments on the Pivotal story (https://www.pivotaltracker.com/story/show/152122952) you can get a sense of the discussion around this. I don't feel strongly either way about leaving this comment in or removing it. It may be better to remove it to avoid confusion. The error states issue is captured in a Pivotal story (https://www.pivotaltracker.com/story/show/152730034), so I don't think that we would lose any reminder about this by removing the comment.

app/assets/javascripts/account/directives/myApplicationDirective.js.coffee
callaghanc: This is Dave's work. I don't have immediate insight into his reasoning here without doing further investigation which I am out of time to do this week. I believe the best way to understand this further would be to try the alternative you propose in your comment and see if that works.

chrisdolendo: Hm then he must have tried it my way and found an obstacle to doing it that way - coz we always go for the standard way first. We have a few refactoring chores in the web app in the pipeline.. I'll include this in it so that we can merge.

app/assets/javascripts/account/directives/myApplicationDirective.js.coffee
chrisdolendo: I would like to remove it. Since you're out of hours, I'll do this and merge. Thanks!

PR #952, opened 2017-11-13
Features/sign in at short form start #151462171

callaghanc:


app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -308,6 +319,11 @@ AccountService = (
   Service.afterUserTokenValidationTimeout = ->
     Service.accountSuccess.messages.userTokenValidationTimeout = $translate.instant('SIGN_IN.USER_TOKEN_VALIDATION_TIMEOUT')
 
+  Service.emailAlreadyInUse = (email = '') ->
```
kaptron: @callaghanc I think this function was leftover as a placeholder? can remove if so.

chrisdolendo:
Rad. Usually, I wouldn't merge WIP commits. If it is a continuation I would just squash/amend the complete commit. No need to update this now if you're booked, but for future PRs :)

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -735,15 +729,22 @@ ShortFormApplicationController = (
...
       $scope.handleErrorState()
     else
-      $scope.clearEligibilityErrors()
-      $scope.goToAndTrackFormSuccess('dahlia.short-form-application.contact')
+      if $scope.loggedIn()
+        $scope.goToAndTrackFormSuccess('dahlia.short-form-application.contact')
+      else
+        ShortFormNavigationService.isLoading(true)
+        AccountService.checkForAccount($scope.applicant.email).then ->
+          if AccountService.shortFormAccountExists()
```
chrisdolendo: Wondering if we need a function to return this value? Or can we just say `AccountService.accountExists` since it gets assigned in `checkForAccount` anyway? Also just for naming, `shortFormAccountExists` makes it seems like there's more than one type of account.

chrisdolendo:
:+1:'ing @kaptron's review here, just had two comments/questions.

app/assets/javascripts/account/AccountService.js.coffee
callaghanc: @kaptron yep that's a leftover, good catch, will take it out

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
callaghanc: I believe used a function for a reason relating to making sure we got the current rather than initial value - @kaptron do you remember more about this?

kaptron: @callaghanc yeah that sounds familiar, I believe we were anticipating needing to use a $scope function in order to get the current value (i.e. rather than just saying $scope.value = Service.value which would just do a one-time assignment) however it looks like that wasn't really needed until chris's follow-up story here: https://github.com/Exygy/sf-dahlia-web/pull/955/files#diff-07eb5682f861e0fe07f9e3d209702a22R646

So yeah in the context of this `checkForAccount` function it could just be checking `AccountService.accountExists`.

callaghanc: @kaptron @chrisdolendo If I changed this PR to use `AccountService.accountExists` and removed `AccountService.shortFormAccountExists()`, would that mess up your work that was based on this PR?

kaptron: @callaghanc should be fine -- the other branch could be merged/rebased to include any commits you make on this one.

chrisdolendo:
@callaghanc @creativecoder Discussed this with Dave last week: When [Features/sign in and reconcile draft apps #152331095 #958](https://github.com/Exygy/sf-dahlia-web/pull/958) is finished (I will take it home), we will merge:

https://github.com/Exygy/sf-dahlia-web/pull/949
and
https://github.com/Exygy/sf-dahlia-web/pull/958

into the same PR (and use that as an 'epic' branch) and update the review apps on the stories. Since the subsequent PRs (949 and 958) we're branching off the other and this mother branch, it's easier to track and merge to `master` as a full-set feature when accepted.

PR #955, opened 2017-11-16
Features/dynamic app submit create account #152250445

chrisdolendo:
… confirmation


app/assets/javascripts/short-form/templates/g1-confirmation.html.slim
```diff
@@ -41,17 +41,18 @@
         br
           | {{ listing.Leasing_Agent_Phone }}
 
-  .app-inner.inset ng-if="!loggedIn()"
-    label.form-label.caps translate="G1_CONFIRMATION.CREATE_ACCOUNT_TITLE"
-    p.form-note translate="G1_CONFIRMATION.CREATE_ACCOUNT_BODY"
+  div ng-if="!loggedIn() && ::accountExists()"
```
kaptron: @chrisdolendo the syntax for this one time binding, even though it looks weird is:

`ng-if="::!loggedIn() && accountExists()"`

I'm not sure if you can stick the `::` next to one part of it like that (possibly?), but my understanding is that angular basically treats it like:

`ng-if="{expression}"`   <-- watch `expression` 
`ng-if="::{expression}"`  <-- one-time bind `expression` 

so it might be sort of all-or-nothing. I'm saying this with only minimal authority on the subject, just my experience from implementing that feature 😊 

last comment -- how does `accountExists` get set by the time you reach this page? I guess based on cat's story, if you were filling out anonymously, you would have _had_ to pass the "name" page at which point it would get set in the Service?

chrisdolendo:
task before merging PR to base branch:  check one-time bind expression syntax

app/assets/javascripts/short-form/templates/g1-confirmation.html.slim
callaghanc: @chrisdolendo I'm picking up where Dave left off on reviewing this PR - what's the status? Were you able to address Dave's comments above?

chrisdolendo: @callaghanc this story was accepted so it looks like the feature works with this syntax will retest again when merged.

PR #957, opened 2017-11-21
Features/disable buttons on submit #151966642

chrisdolendo:


app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -168,15 +168,13 @@ AccountService = (
...
     Service.clearAccountMessages()
     if infoType == 'email'
       params =
         user:
           email: Service.userAuth.user.email
       $http.put('/api/v1/auth', params).success((data) ->
-        bsLoadingOverlayService.stop()
         Service.accountSuccess.messages.email = $translate.instant("ACCOUNT_SETTINGS.VERIFY_EMAIL")
       ).error((response) ->
         bsLoadingOverlayService.stop()
```
callaghanc: @chrisdolendo Do we still need to call `bsLoadingOverlayService.stop()` here since the call to `bsLoadingOverlayService.start()` was removed? Also then maybe we can remove the comment above, `# have to later manually call overlay.stop() since this update doesn't result in a stateChange`.

chrisdolendo: good catch @callaghanc, I meant to remove both. Will remove that line and the comment.

chrisdolendo:
@callaghanc Thank you for your review! Implemented changes - ready for your re-review :)

PR #972, opened 2017-12-13
Chores/refresh procedure on full #153567003

callaghanc:
This PR makes tests pass after the recent Salesforce Full refresh.


creativecoder:
@callaghanc Regenerated the `update_application` vcr and then removed extra stubbed methods you added. The problem was that the application to test an update on needs to have `Draft` status or we don't allow updating it. Added a comment to that effect. Once this passes tests on semaphore, you are good to merge.

callaghanc:
cool thanks!

PR #973, opened 2017-12-14
Chores/Update rubocop #153300270

chrisdolendo:


chrisdolendo:
re code-climate total-coverage, I believe @callaghanc created a chore to cover the test covering from the base refactor. We may need to prioritize that chore so get all checks to go green :)

PR #976, opened 2018-01-02
Bugs/housing counselor page printing incorrectly #152625990

callaghanc:


creativecoder:
It looks like there have been updates to toolkit.css since this PR was created, so it should probably be refreshed with latest toolkit before merging.

PR #977, opened 2018-01-02
Bugs/can go straight to confirmation page #150870824

callaghanc:
…mitted apps


callaghanc:
@creativecoder Hmm...I now wonder whether this PR should even be open at all. Reviewing the Pivotal story, it seems like the actual decision may have been to scrap this bug. I may have failed to close this PR in tandem with scrapping the bug. Will investigate a bit further.

creativecoder:
@callaghanc Copied from Slack:
- For anonymous users: based on the comments in the story (https://www.pivotaltracker.com/story/show/150870824), we scrapped the redirect behavior and instead put the application header as part of the community screening page so a user would understand what happens (https://www.pivotaltracker.com/story/show/152007104). This seems right to me and is good as is on dahlia-full.
- For logged in users who already have a submitted application for the listing, I think we should redirect to the application summary page for that application (right now it goes to the homepage). This should probably be a new story.

callaghanc:
Summary of final decision on this PR, as copied from the related Pivotal story:(https://www.pivotaltracker.com/story/show/150870824):

The original request from design + product was to redirect direct accesses to the application confirmation page, when no application is present, to the home page.

After the discussion in the comments below, the direction from design + product changed. The final decision was that direct accesses to the confirmation page should be redirected to the first page of the application. This matches the redirect behavior for direct accesses to other short form pages that the user is not allowed to access.

At the time of the discussion in the comments below, direct accesses to the confirmation page were already being redirected to the first page of the application. So, at that time, it was determined that the behavior happening in this bug was no longer occurring and that the desired behavior was in fact already happening. So it was decided that this story could be closed without code changes.

However, at a much later date, when the still-open GitHub PR for this story containing the original home page redirect was reviewed by the engineering team, it was found that changes had been made in other stories since this bug was opened that changed the behavior in question. At time of reviewing the GitHub PR, the desired behavior of redirecting direct accesses of the confirmation page to the application start page was no longer occurring. 

Therefore, in order to restore the desired behavior that was decided on in this story's comment thread, a code update was made in the GitHub PR so that direct accesses to the confirmation page are once again redirected to the application start page.

app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -1014,11 +1018,8 @@
           templateUrl: 'short-form/templates/review-application.html'
           controller: 'ShortFormApplicationController'
       onEnter: [
-        '$state', 'ShortFormApplicationService',
-        ($state, ShortFormApplicationService) ->
-          applicationDataExists = !!ShortFormApplicationService.application.lotteryNumber
-          return if applicationDataExists
-          $state.go('dahlia.welcome')
```
chrisdolendo: ah i see where you moved this. :+1:

PR #979, opened 2018-01-08
Features/Application preference record types

creativecoder:


creativecoder:
Merging and delivering master so that we can unblock testing on full environment.

PR #980, opened 2018-01-08
Features/include session id in application #148860651

callaghanc:


chrisdolendo:
@creativecoder see commit 0539f89 for the test to pass. If you could review this update and merge if you're cool with the additional code change that would be great.

creativecoder:
I refactored this just a bit so that still save session_uid in metadata for backward compatibility, and claim application after creating an account works correctly (in e2e tests).

Discussed with @callaghanc and we still would like to improve this in the future to remove duplication between `session_uid` and `externalSessionId`. Follow-up chore is https://www.pivotaltracker.com/story/show/156234576.

PR #982, opened 2018-01-10
Features/hardcode sro text for merry go round #154188736

callaghanc:


creativecoder:
Looks good! Reminder to merge to production to master after this PR is accepted.

PR #983, opened 2018-01-11
Bugs/preference proof uploader stalls on too large files #154041775

callaghanc:


callaghanc:
This PR contains an error trigger for testing purposes. The error trigger is a garbage string added to the file path in uploaded_file.rb to cause an error in attempts to read the file. Tests will not pass with the error trigger in place. The error trigger should be removed before merge.

creativecoder:
@callaghanc Questions:

Did you look into adding tests for any of this?

When the file upload errors, it seems funny to have the error alert and see the error message. Is that something we could quickly correct, or should I write a new story?

creativecoder:
Answers after talking to @callaghanc 

> Did you look into adding tests for any of this?

Cat is going to at least write tests for the changes in `FileUploadsService.js`. If there's time, she'll also add tests for the `preference-proof-uploader` component, otherwise we'll break that out into a separate chore.

> When the file upload errors, it seems funny to have the error alert and see the error message. Is that something we could quickly correct, or should I write a new story?

This was already discussed when the original story was delivered, so will stay as is for now.

PR #987, opened 2018-01-19
Bugs/angular infinite retry loop #151027145

callaghanc:


callaghanc:
This PR contains an error trigger for testing purposes. The error trigger is a stray comma added in locale-en.json to cause an error in attempts to fetch it. Tests will not pass with the error trigger in place. The error trigger should be removed before merge. Tests will not pass with the error trigger in place.

PR #988, opened 2018-01-24
Chores/update nokogiri #154579758

chrisdolendo:


chrisdolendo:
E2E fail because branch is checked out of production env. Jasmine and Rspec pass. Merging to production.

PR #1002, opened 2018-02-14
Bugs/button overflows container in filipino #154188183

callaghanc:
Updates CSS from PL with change that makes button text wrap. PL PR: https://github.com/Exygy/sf-dahlia-pattern-library/pull/62.


chrisdolendo:
@callaghanc can you fix this conflict and merge? I am also happy to re-review after the conflict fix.

PR #1009, opened 2018-03-05
Bugs/Live-Work opt out canceled incorrectly #155714607

creativecoder:
Because of the change in bed4b75 preferences that weren't being claimed were still being reset. This had a side effect of cancelling the claiming and opting out of Live/Work preference, because NRHP was always being reset before application was sent or loaded from Salesforce.

When applicant is eligible for NRHP, this is the correct behavior, because claiming NRHP also claims Live/Work, if changing address resets NRHP it should also reset Live/Work, etc. However if applicant isn't eligible for NRHP, we don't want NRHP affecting anything with Live/Work preference.

Also, ADHP resets Live/Work in the same way that NRHP does, so additional conditions are now in place to make NRHP/ADHP function in the same way.


creativecoder:
@chrisdolendo @callaghanc Please take an extra close look at this one. I want to make sure I'm correct in all my assumptions about preference reset behavior. Preference bugs getting to production are particularly awful because they affect lottery rankings.

PR #1012, opened 2018-03-09
Features/Site wide alert message #155867833

creativecoder:
Set alert message with `ALERT_MESSAGE` env variable


creativecoder:
Note tests will not pass since this was branched from `production`

app/assets/javascripts/shared/SharedController.js.coffee
```diff
@@ -2,7 +2,7 @@
 ###################################### CONTROLLER ##########################################
 ############################################################################################
 
-SharedController = ($scope, $state, SharedService, GoogleTranslateService) ->
+SharedController = ($scope, $state, $window, SharedService, GoogleTranslateService) ->
```
chrisdolendo: good use of $window to access an ENV var!

PR #1017, opened 2018-03-19
Bugs/can go straight to confirmation page #150870824

creativecoder:


creativecoder:
@chrisdolendo @callaghanc I'm reopening this branch because I discovered that the intended functionality was working.

For example, this listing is on dahlia-full: https://dahlia-full.herokuapp.com/listings/a0W0P00000DZ3ObUAL

I should be able to go to https://dahlia-full.herokuapp.com/listings/a0W0P00000DZ3ObUAL/apply/confirmation and get redirected to the first page of the application, but I go to the home page instead. I think we were all fooled because if you go directly to the confirmation page of the app for a community building listing, there's another redirect that happens first, taking you to the community screening question (so it looks like the redirected added in this branch is working, even though it's not).

By injecting `listing` into the onEnter method (which is part of the `resolve` logic in the `short-form-application` parent route), the listing is loaded before this new redirect logic happens.

Also, if there isn't a listing, I figured all short form routes should redirect to the homepage, since none of them work without listing data, so added that logic to the parent route, instead.

PR #1021, opened 2018-03-27
Bugs/google translate resets during page load #155955218

callaghanc:
We previously were beginning Google Translation in the state change start callback. This was occasionally causing the GTranslate process to partially or wholly fail in instances where the page content did not load in the right time or with the right timing for GTranslate to pick up on the content. Moving the kickoff of Google Translation into the state change success callback gives the page content more time to load so that GTranslate has a better chance to pick up on all the content.


callaghanc:
I will add as much testing for the `ExternalTranslateService` to this PR as is reasonable without delaying the release.

callaghanc:
sounds good thanks @creativecoder

callaghanc:
Closing because we are not planning to work on this anytime soon due to prioritizing LAP lease up work.

PR #1022, opened 2018-03-28
Bugs/scroll position maintained on home page #156241951

creativecoder:
This bug was being caused by two specific conditions:

1. The home page url was specified as '/'. When combined with a language param, this required
the home page url to have a trailing slash, like '/es/', unlike any other routes. To link
back to the non-english homepage, used ng-href and appending a trailing slash to the route
url.

2. The keepScrollPosition directive uses a difference between $location.path() at the beginning
and end of state changes to detect if the browser back/forward buttons were in use or not. When
navigating to the home page, it treated that navigation as if the browser buttons were being
used, and so maintained scroll position for the page.

To solve this:
- Homepage url is changed from '/' to '' to not require the trailing slash for non-english home page
- `ui-state` directive is used to dynamically generate the links for the language switcher
- Logo home page link uses ui-sref with route name rather than a custom function

Also note that angular-ui-router was updated to 0.4.2 (most recent version without breaking
changes) to use the `ui-state` directive that let's us dynamically generate a state link


app/assets/javascripts/shared/components/languageSwitcher.js.coffee
```diff
@@ -1,10 +1,8 @@
 angular.module('dahlia.components')
-.component 'languageSwitcher',
+.component 'languageSwitcher', {
   templateUrl: 'shared/components/language-switcher.html'
   controller: ['SharedService', '$state', '$scope', (SharedService, $state, $scope) ->
-    ctrl = @
-
-    @switchToLanguage = (lang) ->
+    $scope.stateForLanguage = (lang) ->
```
callaghanc: @creativecoder this refactor makes me wonder about consolidating the welcome page language duties in angularInitialize. which would remove the need for this function entirely because then the language switcher would just always use the current state in its href. then angularInitialize, which is already parsing welcome page requests, would do the appropriate language-determining. but the one downside to that is that the language switcher would then be outputting technically incorrect paths for the welcome pages - e.g. `/es/welcome-chinese` - which would be correctly transformed by angularInitialize but still seems undesirable. so because of that i think how it is here is good.

creativecoder: 👍

app/assets/javascripts/shared/components/language-switcher.html.slim
```diff
@@ -1,14 +1,14 @@
 ul.inline-list.translate-bar_languages.t-semi.right
   li
-    a ng-class="{active: $ctrl.isSelectedLanguage('en')}" ng-href="{{$ctrl.switchToLanguage('en')}}"
+    a ng-class="{ active: isSelectedLanguage('en') }" ui-state="stateForLanguage('en')" ui-state-params="{ lang: 'en' }"
```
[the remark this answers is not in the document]
callaghanc: 👍

PR #1024, opened 2018-03-31
Chores/Update tests after Salesforce refresh #156412782

creativecoder:


creativecoder:
Seems to be a Salesforce bug preventing deletion of applicaitons. I can't do it manually, either.
screen shot 2018-03-30 at 17 12 59

PR #1032, opened 2018-04-19
Chores/google translate bug stopgap #156934838

callaghanc:
Quick, temp fix for Google Translate bug due to deprioritization of bug due to lease up taking precedence - please see the Pivotal chore for more info: https://www.pivotaltracker.com/story/show/156934838.


callaghanc:
@chrisdolendo Note that this is branched from the original bug story because we do need to include the work that has been done on that story so far. When this chore is accepted, I will merge this chore branch into the original bug branch, and then merge that into master.

PR #1033, opened 2018-04-20
Chores/add logging for geocoding issues #156935075

callaghanc:


chrisdolendo:
@callaghanc Code climate is complaining about test coverage for the additional lines, but since it's just temp logging and we have other priorities, I am okay with it being merged.

PR #1030, opened 2018-04-19
Features/accurate geocoder record #155959812

callaghanc:

PR #781, opened 2017-06-27
Features/Different error email for NRHP/ADHP Listings #134889393

kaptron:

PR #1034, opened 2018-04-24
Features/geocoder project id adhp #155640979

callaghanc:
@chrisdolendo @creativecoder In doing this story, I discovered that although we were passing the `has_nrhp_adhp` value from Angular to Rails in the geocoding process, that `has_nrhp_adhp` value was ultimately not used anywhere. I also found that the value of whether or not the listing in question has NRHP was not ultimately used in the geocoding/matching process. I therefore replaced `has_nrhp_adhp` with a single value called `adhp` that we now use to determine whether to use "ADHP" as the project ID during boundary matching. When reviewing, please check to make sure that I have not missed something about the old `has_nrhp_adhp` value or about the need to have information about NRHP passed through to the geocoder.


app/controllers/api/v1/geocoding_controller.rb
```diff
@@ -36,7 +35,7 @@ def geocoding_data
   def address_within_neighborhood?(address)
```
callaghanc: @chrisdolendo @creativecoder I feel like there should be a check before line 23 of this file (I can't leave a comment on line 23 because I didn't modify it in this commit) that checks whether the listing has ADHP or NRHP before calling `address_within_neighborhood`. My assumption here is that if the listing doesn't have either ADHP or NRHP, then the value for `boundary_match` is automatically nil because the concept of a match shouldn't apply. Do you know what the business rules are around this?

spec/javascripts/services/geocoding_service_spec.coffee
```diff
@@ -0,0 +1,78 @@
...
+      return
+    )
+
+    beforeEach inject((_GeocodingService_, _$httpBackend_) ->
+      httpBackend = _$httpBackend_
+      GeocodingService = _GeocodingService_
+      return
+    )
+
+    describe 'geocode', ->
```
callaghanc: Note that the following tests will need to be updated slightly when both this PR and the PR for updating the boundary match values are merged into master, because that PR adjusts what values we should expect to be returned for the boundary match.

creativecoder:
@callaghanc 
> `has_nrhp_adhp` value was ultimately not used anywhere
Correct. This was a relic of #781, when we used to send geocoding error notifications to different people, depending on if the listing actually had nrhp/adhp. That was removed in #931, but `has_nrhp_adhp` was left sitting around. I think your approach to modify it to our current needs (for ADHP) makes sense.

app/controllers/api/v1/geocoding_controller.rb
creativecoder: @callaghanc I believe this was set up to always return `false` for boundary match because we use that somewhere in AngularJS to know if the geocoding request has been made for that address, or not, but I'd have to dig into the code to be sure.

If that is the case, I'd recommend adding a property to the response, like `"geocoded": true` and using that instead, because I think it would be more helpful if the Salesforce data we store didn't incorrectly store false matches that weren't relevant or were errors.

callaghanc: @creativecoder This PR that is currently open and awaiting product acceptance: https://github.com/Exygy/sf-dahlia-web/pull/1030, affects the values returned for boundary match. It's a bit confusing because both this PR and that PR deal with the same bits of code, just doing slightly different things to them. Actually now I think of it this comment would have been better to leave on the other PR -__- but anyway, the changes in that other PR make it so the the boundary match now has three possible values it can return: true (to indicate a match), false (to indicate that the address was found not to be a match), and nil (to indicate that the address couldn't be put through the matching process due to error or blank address). So in the case of neither ADHP or NRHP being present, we could just return `boundary_match: nil` without even going into the `address_within_neighborhood?` function so that we can skip doing the whole matching process. But I am not sure how that meshes with the business rules.

creativecoder: @callaghanc It's a good question to ask. The boundary matcher can still return results for the building's project ID, but I don't know what meaning this has if the listing doesn't have NRHP. You could ping on slack or bring up at stakeholder or design/product meeting.

callaghanc: @creativecoder I'll do that, thanks for the suggestions of where to go with the q, that tends to be the fuzziest part for me.

callaghanc:
Thanks! Yep, I'll add a label to this to indicate that its merge is pending the resolution of that question.

app/controllers/api/v1/geocoding_controller.rb
callaghanc: @creativecoder Yay Ashley answered my question and confirmed that the approach I mentioned above to boundary matching for listings that have neither NRHP nor ADHP is correct. I'm going to add the change with a comment to this PR. I also am going to add a test for the controller which holds all this logic. I don't think we have this knowledge documented anywhere so a test would be doubly useful for both validation and documentation.

callaghanc:
@creativecoder This PR has been updated with the work from #1030, and with the changes that I discussed and confirmed with Ashley re: boundary match value when listing has neither ADHP nor NRHP. I've also added more rspec tests. So for now I am done updating the code on this PR. 

PR-admin-wise, looks like Code Climate has some complaints, and even though Semaphore had a passing build it's not marked passing here yet, so I'll need to look into that. This PR will need to be re-reviewed, so feel free to either dismiss your old review or start a new once, whichever you believe is more appropriate.

callaghanc:
@creativecoder Ok I refactored to pass Code Climate and Semaphore is passing, so this PR is now 100% ready for re-review!

callaghanc:
Great idea, yes I'll add the boundary matching rules to the wiki.

callaghanc:
Yeah that makes sense. `preferenceAddressMatch` doesn't have to live in the GeocodingService. In general it would be cool to be able to slightly adjust the overall setup of geocoding/boundary matching to make it clear what happens where/when/why, but I certainly don't want to get too much farther into that right now. Anyway do you wanna add those changes to tidy up the concerns?

callaghanc:
@creativecoder Errrrrr well I did say above that I didn't want to get into general refactor too much but er well here we are, just pushed all the refactoring. The one thing this still needs is to make sure all tests pass, and a check to make sure the tests that were already in the ShortFormApplicationService spec sufficiently test the "Matched"/"Not Matched"/"" values for the ultimate value of the application member's `preferenceAddressMatch` field. I would check on those now but really have to leave for this event where DAHLIA is getting an award 😃 so I will look in on this again tomorrow.

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
```diff
@@ -588,7 +585,7 @@ ShortFormApplicationService = (
       # special case for contact form
       if stateName.match(/contact/)
         applicant = Service.applicant
-        addressValid = applicant.preferenceAddressMatch != null
+        addressValid = !_.isNil(applicant.preferenceAddressMatch)
```
callaghanc: @creativecoder need to think through the logic here and make sure that intention.meaning is being preserved

app/assets/javascripts/short-form/GISService.js.coffee
```diff
@@ -2,37 +2,29 @@
 ####################################### SERVICE ############################################
 ############################################################################################
 
-GeocodingService = ($http, ShortFormDataService) ->
+GISService = ($http, $q, ShortFormDataService) ->
```
creativecoder: Doesn't look like we actually need `$q` here--I don't see it used in the diff.

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
creativecoder: @callaghanc I think this makes sense--we want to know if the address has already been geocoded with a valid value, meaning `preferenceAddressMatch` should have a value other than `null` or `undefined`.

app/controllers/api/v1/gis_controller.rb
```diff
@@ -1,76 +1,83 @@
...
-    log_msg = '<< GeocodingController.address_within_neighborhood? ' \
-      'send_notifications called but notifications not sent'
-    logger.error log_msg
+    return if notifications_sent
 
-    # default response
-    nil
+    log_msg = '<< GISController ' \
+      'send_boundary_match_error_notification attempted but notifications not sent >>'
+    logger.error log_msg
```
creativecoder: @callaghanc Do we loose anything by not returning a `nil` response here?

spec/javascripts/services/gis_service_spec.coffee
```diff
@@ -0,0 +1,131 @@
...
+      httpBackend = _$httpBackend_
+      GISService = _GISService_
+      $q = _$q_
+      $rootScope = _$rootScope_
+      return
+    )
+
+    describe 'getGISData', ->
+      describe 'when the attempt to get GIS data is successful', ->
+        afterEach ->
```
creativecoder: @callaghanc Could this be moved to just below `describe 'getGISData'` so it applies to all tests for this method?

creativecoder:
@callaghanc I left a few comments with minor suggestions. Otherwise I've retested this manually, automated tests are passing (after being restarted a couple times for the e2e tests), and it looks to merge!

app/controllers/api/v1/gis_controller.rb
callaghanc: @creativecoder good q, here we are ok to not to have to return anything/worry about the return value. the `send_boundary_match_error_notification` and `send_geocoding_error_notification` methods only send the error notifications and do some logging. they don't handle/affect geocoding data or boundary match data so they don't have to return anything. we do have to make sure that `address_within_boundary?` returns `nil` when an address can't be boundary matched - see line 51 above. the notifications/logging stuff used to be fully mixed into the `geocoding_data` and `address_within_boundary?` methods, so before this PR we were sending notifications, doing logging, and then returning geocoding/boundary match data all in one method. this PR chops up the previously existing methods and does a bit of rearranging to break that up into more manageable chunks.

spec/javascripts/services/gis_service_spec.coffee
callaghanc: @creativecoder yes you're right, i looked into the jasmine docs further and tested putting the `afterEach` on line 55 and removing all subsequent identical `afterEach`s and it does indeed run the block after every `it` within the line 54 `describe` regardless of nesting. i also learned that one can use `this` in `before/after` and subsequent `it`s to share state, which i didn't know before! was always putting general vars outside them for shared state. thanks for making me expand and correct my jasmine knowledge! 😊

app/assets/javascripts/short-form/GISService.js.coffee
callaghanc: 👍

PR #1038, opened 2018-05-30
Updating navigation to use flexbox as a modifier

slowbot:
Implementing fixes from PL 

https://github.com/Exygy/sf-dahlia-pattern-library/pull/77


creativecoder:
Actually, e2e tests will be broken because Alice Griffith preference was added to Automated Test listing for #1039. We can override and merge this anyway.

creativecoder:
@slowbot Yes, ready to merge from my perspective.

PR #1039, opened 2018-06-08
Features/alice griffith preference #157204515

creativecoder:


app/assets/json/translations/locale-en.json
```diff
@@ -789,6 +794,14 @@
         "WELCOME_FILIPINO": "Filipino",
         "WELCOME_SPANISH": "Español"
     },
+    "PREFERENCES": {
+        "ALICE_GRIFFITH": {
+            "TITLE": "Alice Griffith Housing Development Resident Preference",
+            "DESC": "For households in which at least one member was a resident of the Alice Griffith housing development. This includes baseline and current residents that lived in the targeted redevelopment site on or after the time of application for Choice Neighborhoods of October 26, 2010.",
```
callaghanc: @creativecoder is there a source of truth for this description text that we can verify against? couldn't find this text in the InVision screens linked in the Pivotal story or in the Pivotal story itself.

creativecoder: @callaghanc Good question, @chrisdolendo created this translation text when starting the story. I believe it's supposed to match the preference description in Salesforce, but we put it here so it can be translated: https://sfhousing--full.cs95.my.salesforce.com/a0m0P00000yuzO0

app/assets/json/translations/locale-en.json
```diff
@@ -560,7 +563,9 @@
             "PUBLIC_BENEFITS": "Public benefits record",
             "SCHOOL_RECORD": "School record",
             "TELEPHONE_BILL": "Telephone bill (land line only)",
-            "WATER_BILL": "Water bill"
+            "WATER_BILL": "Water bill",
+            "CA_LICENSE": "CA ID or Driver's License",
```
callaghanc: @slowbot the CA state official term for this is "Driver License" not "Driver's License" - check out CA DMV: https://www.dmv.ca.gov/portal/dmv/dmv/dl. would it be ok for us to change "CA ID or Driver's License" to "CA ID or Driver License" in the label here? @creativecoder would that be too confusing since the Salesforce picklist value does use the possessive?

creativecoder: We can update and I'd recommend changing the Salesforce picklist values as well to keep things consistent.

callaghanc: ok, if @slowbot is ok with it then i can ask Vertiba to update that picklist value

app/assets/json/translations/locale-en.json
callaghanc: ok thanks

callaghanc:
@creativecoder @chrisdolendo looks good! only thing is the Driver's/Driver License terminology correction. if @slowbot approves that, i'll put in a request to vertiba to change that value in the picklist and i will update the label value that we use as well. after that decision is resolved this PR will be approved.

chrisdolendo:
@callaghanc @creativecoder - do we need to update this value before doing the release?  Considering our tight timeline to deliver this feature.

creativecoder:
@chrisdolendo I checked in with msol via @slowbot last week. MOHCD calls it "Driver's License" in other places so I'm going to leave as is for consistency (even though technically incorrect, at least according to the state of CA).

chrisdolendo:
Great, thanks for the update @creativecoder - got nervous there for a second. 😂

spec/e2e/features/short_form/error_messages.feature
```diff
@@ -80,3 +80,15 @@ Feature: Short Form Application
...
         When I upload a too-large "Gas bill" as my proof of preference for "liveInSf"
         Then I should see an error about the file being too large
+        And I opt out of "Live/Work" preference
+
+        # error: address not entered for Alice Griffith
+        When I select "Jane Doe" for "aliceGriffith" preference
+        And I hit the Next button "1" time
+        Then I should see a blank address error
+
+# error: address not found for Alice Griffith
```
chrisdolendo: @creativecoder should we delete these?

creativecoder: Whoops, meant to use those in #1041... will go update that PR now.

PR #1043, opened 2018-06-19
Translations

callaghanc:


app/assets/json/translations/locale-en.json
```diff
@@ -319,8 +311,6 @@
         "EMAIL_CONFIRMATION": "Email confirmation does not match",
         "EMAIL_NOT_FOUND": "Email not found. Please make sure your email has an account with us and is confirmed.",
         "FILE_MISSING": "Please upload a valid document",
-        "FILE_UPLOAD": "The file is too large or not a supported file type",
```
chrisdolendo: This doesn't seem right `FILE_UPLOAD` and `FILE_UPLOAD_FAILED` are both being used at the FileUploadService no?

app/assets/json/translations/locale-en.json
```diff
@@ -221,7 +215,6 @@
         "INSTRUCTIONS_P2": "Just upload valid proof of this address and <strong>we'll also give you Live in San Francisco Preference.</strong>",
         "PREFERENCE": {
             "DESCRIPTION": "<strong>Neighborhood Resident Housing Preference</strong> means at least one household member lives within the same Supervisorial District or within a 1/2 mile buffer of the project.",
-            "NAME": "Neighborhood Resident Housing Preference",
```
chrisdolendo: The names of the preferences are gone too.. E.g. these are found in `preferencesSummary`.

It looks like translations is deleting the strings that don't exist anymore in the html files (from running `grunt translations`) , but do so incorrectly coz they are getting translated in the angular code 🤔 I think what we did before for this was to undo the unnecessary deletions.

callaghanc:
@chrisdolendo hmm that's not good. i'm going to see if i can change the deletion process so that it doesn't delete terms that are being used in Angular - we really don't want that!

callaghanc:
@chrisdolendo ok i think i see the issue. it's not that `grunt translations` is ignoring terms used in Angular files. it's that `grunt translations` is not set up to detect terms that are referenced outside of an explicit translation-related use. So, for example, it can detect when we do `$translate('LABEL')`, or `{{'LABEL' | translate}}`, or any other place where the term is explicitly used as a translation term. This is done in a grunt task that uses regexes to detect these translation-specific uses. The terms that are in use but that we see being deleted are terms that are never used explicitly as translation terms. They only show up as plain strings, e.g. the NRHP preference name term is only referenced in this line: `when 'neighborhoodResidence' then 'E2A_NEIGHBORHOOD_PREFERENCE.PREFERENCE.NAME'`. The grunt task doesn't detect that as a translation term, so it thinks that the term is unused and deletes it.

For this PR, I'm going to manually correct and we can merge. I am creating a chore to fix this problem in the grunt task and adding a warning to the POEditor translation workflow mentioning this issue.

callaghanc:
@chrisdolendo Ok I re-added the incorrectly removed terms to this PR, ready for re-review.

chrisdolendo:
Commited latest translations from POE -> the Github

PR #1046, opened 2018-07-11
Bugs/sign in at start loop #158572766

callaghanc:
Marks Intro section completed when user signs in and chooses draft

This resolves a bug where the user was getting looped back to the Name page when trying to proceed to the Household section after signing in and choosing a previous draft. The check to see whether the user is allowed to proceed onto the Household section requires that the Intro section be completed. When we were setting up the previous draft, we were resetting all the completed sections, so the Intro section was marked not completed, hence the user could not proceed to Household.


chrisdolendo:
:+1: Good catch!

chrisdolendo:
the story has been accepted so merging now :)

PR #1051, opened 2018-07-13
Bugs/alternate contact email validation #157204996

akegan:
Looks like it was just the wrong regex in use. Are there tests for form validation things somewhere? I'm not sure where to add a test.

### DIFF
--- a/spec/e2e/features/short_form/error_messages.feature
+++ b/spec/e2e/features/short_form/error_messages.feature
@@ -32,6 +32,14 @@ Feature: Short Form Application
 
         When I fill out the Contact page with an address (non-NRHP match) and WorkInSF
         And I confirm my address
+        # error: invalid email in alternate contact
+        And I fill out the AlternateContact page with the email "grant@exygy"
+        Then I should see an email error on the AlternateContact page
+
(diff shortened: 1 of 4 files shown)


callaghanc:
@akegan For form validation testing, we do that in the e2e tests, specifically in `spec/e2e/features/short_form/error_messages.feature`. If you feel comfortable adding a test in there yourself, go for it! We can also pair on adding the test if you would like, just let me know 🙂

spec/e2e/features/short_form/error_messages.feature
```diff
@@ -32,6 +32,14 @@ Feature: Short Form Application
 
         When I fill out the Contact page with an address (non-NRHP match) and WorkInSF
         And I confirm my address
+        # error: invalid email in alternate contact
+        And I fill out the AlternateContact page with the email "grant@exygy"
+        Then I should see an email error on the AlternateContact page
+
+        When I navigate to the "You" section
```
akegan: This seems a bit redundant to navigate back to the You page again, then skip it, but the alternate contact thing is weird because there are multiple pages so it's not as easy to refill this page. 

Let me know if you have any better suggestions

PR #1055, opened 2018-07-26
[#157264305] Undefined preference errors

akegan:
Fixes here:

### 1. Handled customProofPreferences length bug
If customProofPreferences are undefined, we assume that there are none and continue

### 2. If listing preferences are undefined when formatApplication is called (generally on save)
Then we call getListingPreferences again and wait for listing preferences to be filled in before returning the formatted application,


akegan:
replacing with lower case branch

PR #1056, opened 2018-07-26
Bugs/lottery preferences language loading #159279386

akegan:
It turns out that this was due to the "stubbing preferences" function. 

https://www.pivotaltracker.com/story/show/159279386

### DIFF
--- a/app/assets/javascripts/listings/ListingService.js.coffee
+++ b/app/assets/javascripts/listings/ListingService.js.coffee
@@ -601,8 +601,10 @@ ListingService = ($http, $localStorage, $q, $state, $translate, ModalService, Ex
 
   Service.getListingPreferences = (forceRecache = false) ->
     Service.loading.preferences = true
+    # Reset preferences that might already exist
+    angular.copy([], Service.listing.preferences)
     Service.error.preferences = false
-    # TODO: -- REMOVE HARDCODED FEATURES --
+    # TODO: -- REMOVE HARDCODED PREFERENCES --
     Service.stubListingPreferences()
     # if this listing had stubbed preferences then we can abort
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -606,7 +606,8 @@ ListingService = ($http, $localStorage, $q, $state, $translate, ModalService, Ex
     Service.stubListingPreferences()
     # if this listing had stubbed preferences then we can abort
```
akegan: This comment is incorrect. we abort if the listing preferences are already defined. 

Is this OK to do? Is there a case where the listing preferences could be defined, but for the wrong listing? Should we instead set them to null at the start of this function, then always call?

akegan: I verified that if you navigate to a different listing url, that the preferences refresh as expected, but I'm not sure if I'm missing anything

app/assets/javascripts/listings/ListingService.js.coffee
```diff
@@ -606,7 +606,8 @@ ListingService = ($http, $localStorage, $q, $state, $translate, ModalService, Ex
...
     if !_.isEmpty(Service.listing.preferences)
-      return $q.when(Service.listing.preferences)
+      return $q.when(Service.listing.preferences).then ->
```
akegan: this function is expecting a promise to be returned, is this still returning a promise if I tack on a then? Since it's not really waiting on anything I could probably just set the loading to false in the line above. thoughts?

akegan:
replacing with pr with lowercase branch name

PR #1059, opened 2018-07-30
Chores/you section undefined errors #159019798

akegan:
Talked to @callaghanc and we decided that the best move for this bug is to go ahead and return false if validatedForms doesn't exist, and send some logs to sentry if that's the case. 

https://www.pivotaltracker.com/story/show/159019798

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
@@ -609,6 +609,15 @@ ShortFormApplicationService = (
         applicant = Service.applicant
         addressValid = !_.isNil(applicant.preferenceAddressMatch)
         isValid = isValid && addressValid
+
+      # catch errors where validatedForms becomes undefined
+      if _.isEmpty(Service.application.validatedForms)
+        Raven.captureMessage('Validated forms is unexpectedly empty', {
+          level: 'warning',
+          extra: { stateName: stateName, application: Service.application }
+        })
+        return false
+
       Service.application.validatedForms[section.name][stateName] = isValid
 
   Service.authorizedToProceed = (toState, fromState, toSection) ->
--- a/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
@@ -135,7 +135,7 @@ ShortFormApplicationService = (
     validated = Service.application.validatedForms
 
     # catch errors where validatedForms becomes undefined
-    if isEmpty(validated)
+    if _.isEmpty(validated)
       Raven.captureMessage('Validated forms is unexpectedly empty', {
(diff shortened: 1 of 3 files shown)


callaghanc:
@akegan I've finished adding my updates, will you review the commits i added and approve/request changes? since this is your PR github doesn't allow you to actually add an official review but you can just review it and comment your conclusion here

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
```diff
@@ -609,6 +609,15 @@ ShortFormApplicationService = (
...
         addressValid = !_.isNil(applicant.preferenceAddressMatch)
         isValid = isValid && addressValid
+
+      # catch errors where validatedForms becomes undefined
+      if _.isEmpty(Service.application.validatedForms)
+        Raven.captureMessage('Validated forms is unexpectedly empty', {
+          level: 'warning',
+          extra: { stateName: stateName, application: Service.application }
+        })
+        return false
```
akegan: is returning false sufficient enough to avoid errors here? it doesn't look like it's returning true is the form state is valid in some way

callaghanc: yeah that's a good question! so looks like `checkFormState` is only used in two places: 

1) Line 148 in `angularInitialize.js.coffee`: `ShortFormApplicationService.checkFormState(fromState.name, fromSection) if fromSection`. Here it's being used for its effect of setting state in `validatedForms`. So the goal here if `validatedForms` is empty is to just do nothing, otherwise we get an error. The subsequent lines take care of redirecting the user when `validatedForms` is empty, so `checkFormState` basically just needs to get out of the way there.

2) Line 732 in `ShortFormApplicationService.js.coffee`: `isValid = Service.checkFormState($state.current.name, section)`. Here its return value matters because we use the return value in subsequent lines to decide whether or not to set `Service.application.completedSections[section.name] = false`. In this case, if `validatedForms` is empty, it seems correct that `checkFormState` should return false so that we do set the `completedSections` for the section in question to false.

that was my thinking on it - what are your thoughts?

callaghanc: that will return undefined - i think something that could be helpful is seeing the regular js that coffeescript produces. check out http://js2.coffee/. it allows you to convert btwn js and coffeescript. you can also get editor plugins that do this. paste the `checkFormState` coffeescript into the coffeescript side of that tool and check out the regular js that it translates to. if none of the returns above the last line are hit, then the return value of `checkFormState` is going to end up being the value of`isValid`. ideally i would like to refactor that entire function and its usage, but we already discussed that and agreed to wait on that until we do more general refactor. you could add an explicit return of `isValid` as a last line if you would like. i think we need to move this along so let's not spend too much more time on it atm

akegan:
OK LGTM @callaghanc

PR #1070, opened 2018-08-21
Test Ruby 2.3 upgrade

akegan:


akegan:
can't use PR to test since we'd have to update semaphore with the correct ruby version

PR #1071, opened 2018-08-23
[#159153029] Do not show lottery results if status is not complete

akegan:
This is just to document the changes that work on the frontend to hide test lottery results, we'll have to follow up with a ticket to fix.


akegan:
closing because this was just to demonstrate the changes.

PR #1072, opened 2018-08-31
Chores/update npm high vulnerability [#159945747]

akegan:
https://www.pivotaltracker.com/story/show/159945747

Updates in this PR:

- All package changes were made with `npm audit fix --force`, then cucumber was downgraded to v2 to minimize breaking changes
- Cucumber update led to breaking changes 
  - Characters `/` and `()` are now reserved characters, so I had to remove them from step names and replaced `/` with - in Live-Work and COP-DTHP
  - The format for defining a set of steps [changed](https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md#breaking-changes-4). Then the [syntax updated again in 2.3.0](https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md#new-features-11)

Other updates that could be made:

- Switch Live-Work to Live\/Work and make those steps regex steps


callaghanc:
@akegan is this ready for review, wip, or "do not merge"?

akegan:
work in progress, just updated!

spec/e2e/features/short_form/sign_in_while_applying.feature
```diff
@@ -6,6 +6,7 @@ Feature: Sign-in while filling out application
     Scenario: Setting up an account for sign in while applying tests
         # Birth date indicates < 65 years old for senior building tests
         Given I have a confirmed account for "Alice Walker" with birth date "1/1/2000"
+        When I wait "5" seconds
```
akegan: @callaghanc so this error was just being caused by a timeout. In general, the tests were very slow on my computer. Do you have any ideas for how to add a more consistent wait until done on the account confirmation?

akegan:
Update: I updated to use the syntax that just allows for importing Given, When, and Then. The diff looks terrible now, but the only update is removing the @, then shifting everything to the left by a tab. If you guys would prefer, I can get rid of that last commit and then do the update in a separate PR.

chrisdolendo:
This looks good to me, as for `I wait "5" seconds` - it's been used before, there might be one that doesn't need a wait but this should do for now.

PR #1076, opened 2018-10-19
test commmit for heroku

akegan:


callaghanc:
@akegan This seems like a PR we could close - if you agree, please close it.

akegan:
agreed!

PR #1083, opened 2018-11-08
Chores/Update easypost gem #161814625

akegan:
https://www.pivotaltracker.com/story/show/161814625


akegan:
@callaghanc can you confirm that the easypost gem is just used for address validation during the application process? I want to make sure that we test anything that might be affected.

callaghanc:
@akegan Yes, the EasyPost gem is only used for address validation during the short form app process.

PR #1084, opened 2018-11-12
Features/change pref zipcode to zip #161762312

callaghanc:


akegan:
@callaghanc is this still work in Progress?

callaghanc:
@akegan i marked it WIP while the tests were failing, am going to re-run tests and will remove label if they pass

callaghanc:
Ah right these test failures are due to priority types

app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -321,34 +321,11 @@ ListingController = (
     return type unless labelMap[type]
     return labelMap[type][modifier]
 
-  $scope.priorityLabel = (priority, modifier) ->
```
akegan: @callaghanc you mentioned on the SF ticket that you thought this might be no longer necessary. I checked and it looks like on production we were displaying the impair*ed* version even though this function looks like it was trying to make it display the impair*ments* version. 

I'm also not sure which we would like to display, but based on this function is looks like we were trying for impair*ments*.

callaghanc: my comment mentioned that the "impaired" language in this function should not be necessary, since we're switching to "impairments" to match the new language in salesforce. i haven't looked closely enough into this function to see exactly how it handles the language - lmk if you would like a pair on this

akegan: sure! maybe pairing could help!

app/assets/javascripts/listings/ListingController.js.coffee
```diff
@@ -323,19 +323,19 @@ ListingController = (
...
+      'Hearing impairments':
         name: 'Hearing Impairments'
         description: 'impaired hearing'
-      'Hearing/Vision impaired':
+      'Hearing/Vision impairments':
         name: 'Vision and/or Hearing Impairments'
         description: 'impaired vision and/or hearing'
-      'Mobility/hearing/vision impaired':
+      'Mobility/hearing/vision impairments':
         name: 'Mobility, Hearing and/or Vision Impairments'
```
akegan: @callaghanc after looking at this function more closely, there is actually a change in the name as well. It takes what is returned from salesforce and makes it a bit clearer with the "and/or"s

app/assets/javascripts/listings/templates/listing/_eligibility.html.slim
```diff
@@ -135,7 +135,7 @@ accordion-heading.lead
               ng-pluralize count="units.length" when="{'0': ' units', '1': ' unit', 'other': ' units'}"
           p.c-steel translate="LISTINGS.UNITS_ARE_FOR" translate-value-type="{{::reservedLabel(listing, type, 'unitDescription')}}"
 
-    li.custom-counter_item ng-if="::listingHasPriorityUnits()"
+    li.custom-counter_item ng-if="listingHasPriorityUnits()"
```
akegan: @callaghanc I was able to grab some time with chris this afternoon to learn a bit more about the angular data binding for this.

She suggested that there are 2 options for fixing the problem
1. Remove the one-time data binding (`::`) for this ng-if. 
This means that the function is getting called constantly (including on page scroll), but is also means that when the request for listing units completes, the value is updated. One concern with this is that it could negatively impact performance. I couldn't notice a difference with a quick test, but I'm not sure how we should best evaluate the potential performance hit. 
2. Add a boolean variable to the scope for hasPriorityUnits, then update it when the priority units are set. This seems less-ideal because then we are adding another variable to the scope, but has the benefit of not being called constantly because it'll only update once, when the value is set. 

What are your thoughts? We can discuss in a slack call tomorrow.

chrisdolendo:
I have reviewed and QA-tested the two stories and delivered the for you. Looks good!

PR #1085, opened 2018-11-19
Chores/update overcommit coffeelint #162065874

callaghanc:
The previous Coffeelint configuration in overcommit.yml was not working for me - Overcommit would throw an error when I ran it and it reached the stage where it tried to run Coffeelint. I used the information in this Overcommit GitHub issue: https://github.com/brigade/overcommit/issues/352, to update the configuration to something that worked without error.

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -44,8 +44,9 @@ PreCommit:
   CoffeeLint:
     enabled: true
     description: 'Analyzing with CoffeeLint'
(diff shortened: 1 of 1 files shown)


akegan:
@callaghanc any suggestions for files to change to get overcommit to run on the right kind of file? anything in coffeescript?

callaghanc:
@akegan yep a .coffee file change should trigger it

PR #1086, opened 2018-11-26
Chores/update rack #162024577

callaghanc:


akegan:
@callaghanc this has failed on sign-in-while applying ~3 times since everything else has been passing. I think it's worth looking into

callaghanc:
@akegan Ok I'll take a look

akegan:
Tests are finally passing! merging now.

PR #1090, opened 2018-12-27
Chores/Refresh 12-24 #162455987

akegan:
This PR also Make salesforce instance Url a hidden env var in vcr cassettes https://www.pivotaltracker.com/story/show/162455987


chrisdolendo:
merging since the tests passed and is a chore!

PR #1094, opened 2019-01-08
Chores/stub ownership listing #162585268

akegan:
https://www.pivotaltracker.com/story/show/162585268

This adds some stubbed data for listings and units. It's a very dumb solution but should at least give us the fields we need until we can pass them from salesforce.

As we're using the stubbed data, feel free to update it in any way to help with testing.


akegan:
@software-project tagged you for review on this one since chris has some work to do on our other app (partners) too

PR #1095, opened 2019-01-11
Chores/refactor listing service #162494687

chrisdolendo:
**ListingService Refactor Notes:**

The objective of refactoring the `ListingService` is to divide the service functions into other meaningful Services.

My first approach to doing this was to pull out constants declared in the `ListingService` and transfer them to a new Service called `ListingConstantsService`. I looked up every constant and references to it in the whole codebase (in other directives, services etc) and updated these constants to reference the `ListingConstantsService` instead.

When strategizing the next approach I noticed that the ListingService had groups of related functions:
1. Eligibility 
2. Preferences
3. Occupancy
4. Lottery
5. Listing Sorting/Grouping
6. Unit Sorting/Grouping
7. Seems tricky because of so many codependent functions but possibly creating an ListingApiService

So lottery functions, for example, can go to a `ListingLotteryService`, and anything related to listing sorting and grouping can be a separate service like `ListingGroupingService` or something. You’ll notice that a lot of these functions are already grouped together.

When moving functions out, always ensure that other parts of the application that use those functions are updated as well. For example, `ListingService.resetListingData` is not just used within `ListingService`, but also at `AngularRoutes`. And remember that two Services can’t be co-dependent into each other -- inherit wisely.

Other notes:
`my-favorites` template looks to be a placeholder feature. Can you investigate if it’s being used or needed? It doesn’t seem like it is being used at all, even if the route and template is setup. 

Bugs from Andrea Testing Saturday:
- [x] Labels error on reserved units (reservedLabel error)
- [x] Lottery results search by number not working
- [x] Download application not expanding with links
- [x] Lottery number not showing up in lottery results view from "my applications"


akegan:
hm. I tried to switch the base on this to master and it did not work out well. maybe it needs a rebase instead of just switching the base?

callaghanc:
let's rebase if that's not conflicty

callaghanc:
update: @software-project will take on breaking out areas #1-3 that chris outlined above, and i will tackle areas #4-7. if either of us finds parts of of the ListingService that involve areas that the other person is working on, we will reach out to the other person to collaborate on a solution

callaghanc:
update: this PR now contains all the lottery-related functions broken out from `ListingService` into a new `ListingLotteryService` (item #4 from the list above)

app/assets/javascripts/listings/components/listingContainer.js.coffee
```diff
@@ -2,85 +2,82 @@ angular.module('dahlia.components')
...
-      @error = ListingService.error
-      @toggleStates = ListingService.toggleStates
-      @AMICharts = ListingService.AMICharts
-      @favorites = ListingService.favorites
-      @listingDownloadURLs = ListingService.listingDownloadURLs
+  controller: ['ListingDataService', 'ListingEligibilityService', 'ListingIdentityService', 'ListingUnitService', 'SharedService',
+  (ListingDataService, ListingEligibilityService, ListingIdentityService, ListingUnitService, SharedService) ->
+    ctrl = @
+    # TODO: remove Shared Service once we create a Shared Container
+    @listingEmailAlertUrl = "http://eepurl.com/dkBd2n"
```
akegan: I wonder if this should be moved to the constants service

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
```diff
@@ -2,16 +2,16 @@ h4.h-caps-lined
   | What to Expect
 p.c-steel.t-small.margin-bottom
   ' Applicants will be contacted by the leasing agent
-  span ng-if="!$ctrl.parent.listingIsFirstComeFirstServe()"
+  span ng-if="!$ctrl.parent.isFirstComeFirstServe()"
```
akegan: seems like we may need to pass the listing in here now, right?

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
```diff
@@ -2,16 +2,16 @@ h4.h-caps-lined
...
     | in lottery rank order until vacancies are filled.
-  span ng-if="$ctrl.parent.listingIsFirstComeFirstServe()"
+  span ng-if="$ctrl.parent.isFirstComeFirstServe()"
```
akegan: and pass the listing in here too

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
```diff
@@ -2,16 +2,16 @@ h4.h-caps-lined
...
     | on a first-come, first-serve basis until the vacancy is filled.
 .show-for-print ng-show="$ctrl.parent.toggleStates[$ctrl.parent.listing.Id].whatToExpectOpen"
   p.c-steel.t-small
     | All of the information that you have provided must be verified, and your eligibility confirmed.  Any fraudulent statements will cause your application to be removed from the lottery.  You will also be removed from the lottery if you cannot prove a housing preference, or if a household member appears on more than one application.
   p.c-steel.t-small
     | Should your application be chosen
-    span ng-if="!$ctrl.parent.listingIsFirstComeFirstServe()"
+    span ng-if="!$ctrl.parent.isFirstComeFirstServe()"
```
akegan: pass listing here too (just want to make sure we don't miss any)

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
```diff
@@ -393,30 +392,30 @@ ShortFormDataService = (ListingService) ->
     preferences = Service._initPreferences(data)
     shortFormPrefs = angular.copy(sfApp.shortFormPreferences) || []
     shortFormPrefs.forEach( (shortFormPref) ->
-
-      listingPref = ListingService.getPreferenceById(shortFormPref.listingPreferenceID)
+      listing = sfApp || ListingDataService.listing
```
akegan: what's sfAPP here?

app/assets/javascripts/listings/ListingLotteryService.js.coffee
```diff
@@ -0,0 +1,96 @@
...
+      angular.copy(data, Service.lotteryRankingInfo[listing.Id])
+      Service.loading.lotteryRank = false
+      Service.lotteryRankingInfo[listing.Id].submitted = true
+    ).error( (data, status, headers, config) ->
+      Service.loading.lotteryRank = false
+      Service.error.lotteryRank = true
+    )
+
+  Service.resetData = ->
+    angular.copy({}, Service.lotteryBucketInfo)
```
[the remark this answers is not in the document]
callaghanc: i also wondered this, it feels like we should, but i also wonder if there's some specific reason why it's not reset?

app/assets/javascripts/listings/ListingUnitService.js.coffee
```diff
@@ -0,0 +1,128 @@
+############################################################################################
```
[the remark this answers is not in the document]
callaghanc: i didn't do a deep look into the internals of the unit grouping. i'd say it's worth putting a note in either a future story that deals with units, or making a new dedicated story to look further into salesforce vs web app grouping of units

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
callaghanc: my guess is that it's an application object, in a format that is recently retrieved from salesforce and hasn't been through web app-specific formatting yet. not 100% sure. but this line is a typo, it should be `listing = sfApp.listing || ListingDataService.listing`, i'll update that

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
callaghanc: these are an instance of the component controllers using an implicit listing object from the component controller. i think this is a larger question about that pattern of using `@listing` as the default argument in component controller functions that pass a listing as an argument to a service function. let's add this topic to a list of general discussion points that we should discuss after your initial review of this PR

callaghanc:
@akegan Let's create a list here of topics to discuss from review:

- [ ] Use of default `@listing` argument in component controller functions

app/assets/javascripts/listings/components/listingContainer.js.coffee
callaghanc: yeah i think so. this is another topic we should discuss: the intention around location of data, and also what the distinction is between business logic and display logic and the locations of those things, which related a bit to the question you had the other day. adding these to the PR discussion topics

app/assets/javascripts/listings/ListingConstantsService.js.coffee
```diff
@@ -0,0 +1,103 @@
...
+
+  # Create a mapping to Salesforce naming conventions
+  Service.RESERVED_TYPES = {
+    VETERAN: 'Veteran'
+    DISABLED: 'Developmental disabilities'
+    SENIOR: 'Senior'
+  }
+
+  # TODO: -- REMOVE HARDCODED FEATURES --
+  Service.LISTING_MAP = {
```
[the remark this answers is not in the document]
callaghanc: this is still being used, by `mapSlugToId`. i'm not fully certain what that whole setup is for. i recall that one reason for the use of `listingIs` and the `LISTING_MAP` is to check listing identity to apply special cases. cannot recall why a name/id mapping is used instead of directly checking a listing prop. may be partially to centralize all instances of special cases to make it easier to track them for future removal, since we do want to eventually remove all special cases. it may be worth looking at the git history around this to uncover what more `LISTING_MAP`, `listingIs`, and `mapSlugToId` are used for. i'm happy to do this as i have a little more historical context to apply. i'm inclined to put that work into a new story as it's not blocking this PR's functionality - what are your thoughts?

app/assets/javascripts/listings/ListingIdentityService.js.coffee
```diff
@@ -0,0 +1,42 @@
...
+# This service contains functions that check properties of a listing to determine
+# whether the listing is or is not a certain kind of listing.
+#
+# TODO: It's not fully clear that this service needs to be a separate service.
+# Look further into where/how the functions in this service are used to see if
+# there is a reasonable alternative way to manage these functions.
+ListingIdentityService = (ListingConstantsService) ->
+  Service = {}
+
+  Service.listingIs = (name, listing) ->
```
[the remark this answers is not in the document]
callaghanc: see comment above

app/assets/javascripts/listings/ListingConstantsService.js.coffee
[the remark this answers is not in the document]
akegan: I made a ticket going through to remove harcoded things https://www.pivotaltracker.com/story/show/163463841. We can just include that investigation when we do that ticket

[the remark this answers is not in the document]
callaghanc: sounds good!

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+      Service.getListingResponse(deferred, retranslate)
+    ).cached(
+      Service.getListingResponse(deferred, retranslate)
+    ).error( (data, status, headers, config) ->
+      deferred.reject(data)
+    )
+    return deferred.promise
+
+  # Remove the previous listing and all it's associated data
+  Service.resetListingData = () ->
```
akegan: looks like this is missing removing units and preferences

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+    openMatchListings = []
+    openNotMatchListings = []
+    closedListings = []
+    lotteryResultsListings = []
+
+    listings.forEach (listing) ->
+      if ListingIdentityService.isOpen(listing)
+        # All Open Listings Array
+        openListings.push(listing)
+        if listing.Does_Match
```
akegan: any idea what this actually does? 

Here's what the field looks like on Salesforce 
image

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+      return
+    )
+
+  Service.isAcceptingOnlineApplications = (listing) ->
+    return false if _.isEmpty(listing)
+    return false if ListingLotteryService.lotteryComplete(listing)
+    return false unless ListingIdentityService.isOpen(listing)
+    return listing.Accepting_Online_Applications
+
+  Service.getListingAndCheckIfOpen = (id) ->
```
akegan: this seems oddly specific and the name doesn't really communicate that if there's no listing or if it's not open it's going to change the state. It also seems to only be used in angularRoutes. Would it make more sense to move these if/else statements directly into angular routes?

app/assets/javascripts/listings/ListingDataService.js.coffee
callaghanc: looks like the original function didn't reset units, preferences, or lottery ranking info. it seems like it would be in line with the intention of this function to add resets of those, but can we think of a reason why those were not already being reset?

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+    charts
+
+  Service.priorityTypes = (listing) ->
+    Service.collectTypes(listing, 'prioritiesDescriptor')
+
+  Service.collectTypes = (listing, specialType) ->
+    _.map listing[specialType], (descriptor) ->
+      descriptor.name
+
+  Service.priorityLabel = (priority, modifier) ->
```
akegan: can this be moved to ListingConstantsService?

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+        name: 'Mobility, Hearing and/or Vision Impairments'
+        description: 'impaired mobility, hearing and/or vision'
+      'Mobility impairments':
+        name: 'Mobility Impairments'
+        description: 'impaired mobility'
+
+    return priority unless labelMap[priority]
+    return labelMap[priority][modifier]
+
+  Service.reservedLabel = (listing, type,  modifier) ->
```
akegan: this one too, would it make sense to move it to listingConstantsService?

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+        eligibility: 'People with developmental disabilities'
+        reservedFor: 'people with developmental disabilities'
+        reservedForWhoAre: 'developmentally disabled'
+        unitDescription: 'people with developmental disabilities'
+
+    return type unless labelMap[type]
+    return labelMap[type][modifier]
+
+  # used by My Applications -- when you load an application we also parse the attached listing data
+  Service.loadListing = (listing) ->
```
akegan: I'm not really clear on what's going on in this function. if the listing passed matches the listing we already have, return, else set the listing we have (from somewhere else) into the listingDataService?

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+    return if Service.listing && Service.listing.Id == listing.Id
+    # TODO: won't be needed if we ever consolidate Listing_Lottery_Preferences and /preferences API
+    listing.preferences = _.map listing.Listing_Lottery_Preferences, (lotteryPref) ->
+      {
+        listingPreferenceID: lotteryPref.Id
+        preferenceName: lotteryPref.Lottery_Preference.Name
+      }
+    angular.copy(listing, Service.listing)
+
+  Service.seniorMinimumAge = (listing) ->
```
akegan: I'm not sure this warrants its own function. Should it just move up to reservedLabel? otherwise it's not clear from the function name that it's actually returning a formatted string with '+' in it, rather than a number.

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+      Zip_Code = ''
+
+    if display == 'street'
+      return "#{Street_Address}"
+    else if display == 'city-state-zip'
+      return "#{City} #{State}, #{Zip_Code}"
+    else
+      "#{Street_Address}#{City} #{State}, #{Zip_Code}"
+
+  Service.getListingDownloadURLs = ->
```
akegan: it would be really great to rename this to something a bit clearer. like getApplicationDownloadUrl or getPaperApplicationDownload url. Thoughts?

We can save it until we add the paper applications for ownership listings if that's easier?

app/assets/javascripts/listings/ListingDataService.js.coffee
```diff
@@ -0,0 +1,434 @@
...
+  Service.getProjectIdForBoundaryMatching = (listing) ->
+    return unless listing
+    if ListingPreferenceService.hasPreference('antiDisplacement', listing)
+      'ADHP'
+    else if ListingPreferenceService.hasPreference('neighborhoodResidence', listing)
+      listing.Project_ID
+    else
+      null
+
+  Service.mapSlugToId = (id) ->
```
akegan: looks like this is only used as part of the hardcoding scheme. maybe we can add a TODO to this function that says "TODO: Remove when we remove hardcoded listings

app/assets/javascripts/listings/ListingDataService.js.coffee
software-project: It make sense, will do that.

app/assets/javascripts/listings/ListingDataService.js.coffee
akegan: ah, that's a good point. you're right that we should be strict about what a constant is. let's not move it

app/assets/javascripts/listings/ListingDataService.js.coffee
software-project: we can make it inline in reservedLabel like `listing.Reserved_community_minimum_age ? "#{listing.Reserved_community_minimum_age}+" : ''` 
but it's a bit long and going along with DRY spirit I would rename it to sth like `formatSeniorMinAge`. What do you think?

akegan: sure that works for me. Let's move it up so it's right next to reservedLabel too if possible

app/assets/javascripts/listings/ListingDataService.js.coffee
callaghanc: i've put a note to do that in the story about removing hardcoding in general: https://www.pivotaltracker.com/story/show/163463841

app/assets/javascripts/listings/ListingDataService.js.coffee
callaghanc: yes agree, the name is misleading, also agree let's leave it for that story where we deal with paper apps for ownership. i've put a note to do that in the relevant story: https://www.pivotaltracker.com/story/show/162975138

app/assets/javascripts/listings/ListingDataService.js.coffee
callaghanc: i would be fine with moving it to angularRoutes, would you like to do that?

app/assets/javascripts/listings/ListingDataService.js.coffee
callaghanc: i don't know, could you ask msol about this?

akegan:
small bug found when just trying to reproduce other bug locally 
image

akegan:
@callaghanc it's just on the main browse listings page, seems to be an issue with reserved units

akegan:
Starting testing now. 
Priority/Reserved units on listing directory page
image

Same issue on listing show page:
image


Download application buttons are missing:
image

Lottery results search does not work when you click the arrow button:
image

akegan:
I just pushed a fix to the reservedLabel display issue,

app/assets/javascripts/listings/LotteryModalController.js.coffee
```diff
@@ -5,26 +5,27 @@
...
+  ListingLotteryService,
   AnalyticsService
 ) ->
-  $scope.listing = ListingService.listing
-  $scope.application = ListingService.application
-  $scope.lotteryBucketInfo = ListingService.lotteryBucketInfo
-  $scope.lotteryRankingInfo = ListingService.lotteryRankingInfo
-  $scope.favorites = ListingService.favorites
+  $scope.listing = ListingDataService.listing
+  $scope.application = ListingDataService.application
```
akegan: this is the problem that causes the missing lottery number issue in the lottery results modal when accessed from my-applications page

This should be calling ShortFormDataService.application instead of ListingDataService Application. It was a bug introduced in the controller refactor:
https://github.com/Exygy/sf-dahlia-web/pull/1092/files#diff-d2adb468b27b224eb81d8ab85755aeabL35

I tried pulling in the ShortFormApplicationDataService into this controller but got an error for some reason. I have to head offline now but will keep looking into it tomorrow 
image

akegan:
It looks like one of my updates broke jasmine tests on this. I'm taking a look now.

app/assets/javascripts/listings/LotteryModalController.js.coffee
software-project: This one is resolved for both modals where LotteryModalController is used.

software-project:
@akegan download application should be working. Tests for LotteryModalController are fixed as well.

akegan:
Just did another round of testing with Emily. Found 2 things:

1. Download applications tab, characters aren't coming out right
This PR is on the left, master is on the right
image

2. When you do eligibility search, some units are showing up as a match and as a non-match on both Full and this PR. E.g. if you filter with 5 household members, and 50k/year, Hyde, Automated Testing listing Senior all, and Automated Test listing all show up in the match and the non-match section. This makes me concerned that we're maybe displaying listings that are not a match as a match? If no one has any quick ideas for that, we can address it in a follow-up bug ticket. 

Other than those 2 things, everything looks good!

callaghanc:
@akegan I can take a look at the download application characters! For the matching issue, if it also shows on Full, let's make a follow-up story for that.

akegan:
Here's the follow up story for the eligibility estimator https://www.pivotaltracker.com/story/show/163545100

callaghanc:
Just updated the paper app download URLs, will keep checking/re-running tests till they pass.

akegan:
Download application special characters look good to me now! @emilyshore mentioned that she wanted to check on it again before we merge though. I also think that I'd be more comfortable if we can get the review app up and running before we merge

callaghanc:
We'll update when we know more about the account billing issue on Heroku and when we can get the review app up again.

PR #1104, opened 2019-01-28
Features/add ownership listings #162561558

software-project:
https://www.pivotaltracker.com/story/show/162561558

DON'T MERGE this is a branch off from listing service refactoring. Rebase before merging.

### DIFF
--- a/app/assets/javascripts/config/angularRoutes.js.coffee
+++ b/app/assets/javascripts/config/angularRoutes.js.coffee
@@ -77,6 +77,20 @@
           $translate('PAGE_TITLE.LISTINGS')
         ]
     })
+    .state('dahlia.listings-for-sale', {
+      url: '/listings/for-sale'
+      views:
+        'container@':
+          templateUrl: 'listings/templates/listings.html'
+      resolve:
+        listings: ['$stateParams', 'ListingDataService', ($stateParams, ListingDataService) ->
+          ListingDataService.getListings({checkEligibility: true, retranslate: true, params: {Tenure: 'sale'}})
+        ]
+        $title: ['$translate', ($translate) ->
+          # translate used without ".instant" so that it will async resolve
+          $translate('PAGE_TITLE.LISTINGS')
+        ]
+    })
     .state('dahlia.listing', {
       url: '/listings/:id?preview',
       params:
--- a/app/services/force/listing_service.rb
+++ b/app/services/force/listing_service.rb
@@ -25,7 +25,7 @@ module Force
       imageURL
       Tenure
     ].freeze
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -77,6 +77,20 @@
           $translate('PAGE_TITLE.LISTINGS')
         ]
     })
+    .state('dahlia.listings-for-sale', {
```
akegan: we're going to have different directory headers and content based on whether the page has rental or ownership lisitngs.

What do we think is the best plan for differentiating between the two? As a placeholder for now we could do the same thing that I have for show ownership listing where there's a bare h2 header at the top that lets you know that it's the ownership page?

callaghanc: Is that work part of this story or part of an upcoming story? If part of an upcoming story, I'd say let's do that work when we get to it, as it will be more clear then exactly what is needed.

akegan: I think that having an example of how to differentiate between the two pages is part of this story.

Actually wiring up the changes can be in this story: https://github.com/Exygy/sf-dahlia-web/pull/1108, or in a follow-up story.

callaghanc: The AC for this story don't mention changes to the page itself: https://www.pivotaltracker.com/story/show/162561558. They actually say the opposite, "Have an identical layout to /listings (or /listings/for-rent)". I'm leaning towards keeping this story to the AC as written, and approaching the new ownership-specific page content in one go when we get to that - it feels less fragmented to me.

akegan: OK. I think that's a mistake in the AC to be honest. I think when we were writing it and estimating it (at 3 points!) we were probably anticipating having a totally new html template for the ownership page and didn't specify that the page had to be modifiable in the future because of that. 

If you think this story is ready to go as is and we don't have a solution for how to differentiate between the pages ready to go, then we can move it into a future story.

akegan: Here's a ticket to follow-up on it: https://www.pivotaltracker.com/story/show/163633411

callaghanc: Seems like you're probably right about the original assumption being that we would create whole new template, hence the estimation. The AC of this story: https://www.pivotaltracker.com/story/show/162561700, say to add the sale- and rental-specific headers to the two separate listings pages - it sounds like this is the story where that content gets added and hence where a differentiation must be introduced - does that sound right? Definitely want to make sure we don't lose any work that we had planned on doing, so want to be certain that the work we're talking about here is captured.

PR #1107, opened 2019-01-30
[#160000126] Hide lottery results when status is not 'Lottery Complete'

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/160000126


akegan:
I need assistance getting a listing setup to QA this one. I've reached out to salesforce folks for help

akegan:
looks pretty good so far!

Here's the listing that I'm using for QA: 
https://dahlia-full-pr-1107.herokuapp.com/listings/a0W0P00000HMHrwUAH
You can see that no matter what I do (incl resetting the cache) I can't get the lottery results modal to stop showing up for this listing on full https://dahlia-full.herokuapp.com/listings/a0W0P00000HMHrwUAH.

One note is that there's a weird behavior where the download lottery results button shows up disabled then disappears (see gif). Can we get rid of that?
download_lottery_results_button

Next step in testing for me is to make sure that it works as expected from the my-applications page

akegan:
I changed the lottery status for alice griffiths phase 3 to not-lottery-complete and was able to confirm that lottery modal is hidden from that view too

image

akegan:
just to add more tests to your workload, do we have any tests around showing the lottery results button?

software-project:
Tests added.

akegan:
@software-project did you see my comment about the "download lottery results" button showing up then disappearing while the page is loading. have you had a chance to look into that?

software-project:
@akegan sorry, I got an impression it's resolved. Will have a look.

spec/javascripts/services/listing_lottery_service_spec.coffee
```diff
@@ -96,6 +96,16 @@ do ->
         spyOn(ListingLotteryService, 'lotteryDatePassed').and.returnValue(false)
         expect(ListingLotteryService.lotteryIsUpcoming(fakeListing)).toEqual true
 
+      it 'returns true when the listings lottery is not complete', ->
+        fakeListing.Lottery_Status = 'Not Yet Run'
+        spyOn(ListingLotteryService, 'lotteryDatePassed').and.returnValue(false)
+        expect(ListingLotteryService.lotteryIsUpcoming(fakeListing)).toEqual true
+
+      it 'returns true when the listings lottery is complete', ->
```
akegan: is this meant to be "returns false"?

software-project: yes, thanks for catching this.

software-project:
@akegan I fixed the value in the test description. Also as we discusses, in the case when lottery is not complete, so button should show. In that case I went ahead to hide the section. As a side result flickering is gone. Is it ok?

akegan:
@software-project I think that makes sense. 

@jrubenoff do you want to give this a quick look to see if the behavior makes sense to you?

app/assets/javascripts/listings/templates/listing/_lottery_results.html.slim
```diff
@@ -1,7 +1,7 @@
 h4.t-gamma.t-serif Lottery Results
 p.content-group_day.margin-bottom
   | {{ $ctrl.parent.listing.Lottery_Results_Date | date : 'longDate' }}
-.padding.bg-dust
+.padding.bg-dust ng-if="$ctrl.lotteryComplete()"
```
akegan: do we still need this if we have the control a level up?

akegan: for line 11 below:
```
  a.button.primary.expand.no-margin ng-if="$ctrl.showDownloadLotteryResultsButton()" ng-href="{{$ctrl.parent.listing.LotteryResultsURL}}" target="_blank" ng-disabled="$ctrl.loading.lotteryResults" data-event="gtm-download-lottery-results"
```
Should we add in the `$ctrl.loading.lotteryResults` to the ng-if so we only see it if the results are no longer loading and there's still no buckets?

jrubenoff:
Ideally, the disabled button should say "View Lottery Results" while loading.

But this isn't new behavior (I noticed it months ago), so feel free to fix this later if there's significant effort involved.

app/assets/javascripts/listings/ListingLotteryService.js.coffee
```diff
@@ -40,7 +40,7 @@ ListingLotteryService = ($http, ListingIdentityService, ModalService) ->
 
   Service.lotteryIsUpcoming = (listing) ->
     return false unless listing
-    !listing.Lottery_Results && !Service.lotteryDatePassed(listing)
+    !listing.Lottery_Results && !Service.lotteryDatePassed(listing) && !Service.lotteryComplete(listing)
```
software-project: @akegan I forgot to ask about this one. Should I remove `lotteryDatePassed` to fulfill condition from ticket's acceptance criteria?

akegan: no I think this is good as is

PR #1109, opened 2019-01-31
Bugs/fix listings display with eligibility filter #163545100

chrisdolendo:
https://www.pivotaltracker.com/story/show/163545100

Simple enough fix!


chrisdolendo:
@akegan good Q, since it is a fix in the template, I can create a protractor e2e test for it. The consideration here is if this is worth writing an e2e test for - on one hand yes to ensure it's covered and on the other hand protractor is already very slow. I looked up how we can unit test it but doesn't seem to be an easy way for this framework. If you think it's worth doing an e2e test, I can write one for it.

akegan:
hm. Since the eligibility filter is a core functionality of the app and we don't currently have any testing on it, I am tempted to add a test for it. @callaghanc what are your thoughts?

callaghanc:
@akegan this definitely seems like a great opportunity to stick to our commitment of tests in every PR and to add some tests that begin to address the eligibility filter! let's think about how we can do this in this PR in a way that gets the ball rolling for eligibility-related testing and also is quick enough to get this PR moved along in a timely fashion. my first thoughts are that this is a place where view unit testing may be suitable, perhaps piggybacking off the view unit testing work that will be done in #1115, and also that we don't need to test in this PR for everything involving eligibility - it would be appropriate to begin by just testing the work done in this PR, i.e. that the correct open matched listings are shown. what are your thoughts @chrisdolendo @akegan?

akegan:
@chrisdolendo I've added a WIP label back to this PR since we want to at least some level of testing

chrisdolendo:
Posting the updated I announced on standup: 

Unit View testing turned out to be needing more work than usual.
- setup Karma so we are able to compile slim files - 2-3 pointer effort
- integrate it with jasmine tests
- other considerations -- need to stub parent components being used

Upside to doing a chore on this is that it'll be faster than e2e tests once we get a few written and the foundation is setup.

PR #1113, opened 2019-02-01
Features/ownership paper app links #162975138

callaghanc:
https://www.pivotaltracker.com/story/show/162975138


callaghanc:
Still need to get tests passing on this PR but it is ready for someone to begin review on it.

spec/e2e/features/listings/listing.feature
```diff
@@ -7,5 +7,10 @@ Feature: Listings
       Then I should be redirected to the welcome page
 
     Scenario: Attempting to go to the Ownership listings page
-      Given I try to go to the Ownership listings page
+      Given I go to the Ownership listings page
       Then I should see available units
+
+    Scenario: Viewing and interacting with a listing page
```
callaghanc: This is a very general name for this scenario, even though the scenario currently only contains a check for going to the listing page and interacting with the download application feature, because I envision that it will be expanded and divided as we progress in developing the listing-related parts of Ownership.

spec/e2e/features/listings/listing.feature
```diff
@@ -7,5 +7,10 @@ Feature: Listings
...
+      Given I go to the "Test Listing" listing page
+      And I click the Download Application button
+      Then I should see at least one paper application download link
```
callaghanc: I wanted to add another step here of actually clicking on the links and checking that they open in new tabs, but I felt wary of the time that would add to the e2e tests. Thoughts on this are welcome.

callaghanc: In general I would prefer some kind of render testing for this, but as we've discussed we are a little ways off from implementing that new kind of testing.

akegan: is there a way we could check that the links are valid without having to visit the page?

maybe we could do some subset of one of these: https://stackoverflow.com/questions/29212375/how-to-test-html-links-with-protractor

callaghanc:
thanks for review @akegan! i'd just like to check out that testing link you posted and see if that leads to any little test improvements we could add here before merging

spec/e2e/features/listings/listing.feature
callaghanc: after chatting with @akegan we decided to test the download application URLs with a Jasmine test of the `getListingPaperAppURLs` function in the `ListingDataService`

PR #1114, opened 2019-02-02
Features/rearrange listing details elements #162565302

chrisdolendo:
https://www.pivotaltracker.com/story/show/162565302


app/assets/javascripts/listings/components/features-section.html.slim
```diff
@@ -14,6 +14,8 @@ accordion-heading
     feature-row title="Appliances" description="{{::$ctrl.parent.listing.Appliances}}"
     feature-row title="{{$ctrl.parent.isOwnershipListing($ctrl.parent.listing) ? 'Services Covered by HOA Dues' : 'Services Onsite' }}" description="{{::$ctrl.parent.listing.Services_Onsite}}"
     feature-row title="Parking" description="{{::$ctrl.parent.listing.Parking_Information}}"
+    feature-row title="Smoking Policy" description="{{::$ctrl.parent.listing.Smoking_Policy}}"
```
akegan: why not use locale-en.json for this if we already have it defined?

app/assets/javascripts/listings/templates/listing/_legal.html.slim
```diff
@@ -1,31 +0,0 @@
-accordion-heading
```
akegan: was this not in use anywhere?

chrisdolendo: @akegan it was converted into a component and the partial was not deleted -- from the refactor.

app/assets/javascripts/listings/components/features-section.html.slim
chrisdolendo: I wasn't 100% sure how to integrate the use of `translate` in there, and google translate runs through this page anyway. But I can try to see if if something like `title="{{SMOKING_POLICY | translate }}"` will work!

PR #1115, opened 2019-02-02
Features/update max occupancy logic in listing #162565299

chrisdolendo:
https://www.pivotaltracker.com/story/show/162565299

Logic is implemented for both use cases - but need to wait for https://www.pivotaltracker.com/story/show/163472473 to be implemented to test for `maximum occupancy is blank`


akegan:
@chrisdolendo larry is deploying the change to full now so it should be ready for testing soon!

akegan:
@chrisdolendo after discussing at our eng meeting this morning, we think that this PR is a good candidate for trying out the view testing you linked to in your comment above.

If you can time-box it to creating an example test for this PR into tonight and (maybe some of tomorrow night) then I think we should go for it.

chrisdolendo:
@akegan sounds good to me thank you for communicating! I will timebox this implementation as you suggested.

chrisdolendo:
Looking closer, we didn't need a unit view test for this specific feature because the occupancy value was isolated into its own component function 👍 Unit view testing can come in the bugfix on eligibility #1109

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
```diff
@@ -12,8 +12,10 @@ angular.module('dahlia.components')
       @error = ListingPreferenceService.error
 
       @occupancy = (unitSummary) ->
-        return '1' if unitSummary.maxOccupancy == 1
-        unitSummary.minOccupancy + '-' + unitSummary.maxOccupancy
+        if unitSummary.minOccupancy == 1 || unitSummary.maxOccupancy == null
+          unitSummary.minOccupancy
+        else
```
akegan: should we put the "at least" in here to make it easier to test and to keep the logic all in the same place?

akegan:
@chrisdolendo let me know when this is ready for review again

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
chrisdolendo: Done!

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
```diff
@@ -13,9 +13,9 @@ angular.module('dahlia.components')
 
       @occupancy = (unitSummary) ->
         if unitSummary.minOccupancy == 1 || unitSummary.maxOccupancy == null
-          unitSummary.minOccupancy
+          "at least #{unitSummary.minOccupancy}"
         else
-          unitSummary.minOccupancy + '-' + unitSummary.maxOccupancy
+          "at least #{unitSummary.minOccupancy}-#{unitSummary.maxOccupancy}"
```
akegan: in this case it should actually just show minOccupancy-maxOccupancy

spec/javascripts/components/listing/eligibilitySection_spec.coffee
```diff
@@ -46,12 +46,22 @@ do ->
...
+        describe 'maxOccupancy is 1', ->
+          it 'returns minOccupancy value', ->
+            unitSummary = { minOccupancy: 1 , maxOccupancy: 1 }
+            expect(ctrl.occupancy(unitSummary)).toEqual(1)
+        describe 'is an SRO unit', ->
+          it 'returns 1', ->
+            unitSummary = { minOccupancy: 1 , maxOccupancy: 1 }
+            expect(ctrl.occupancy(unitSummary)).toEqual(1)
+        describe 'maxOccupancy is null', ->
+          it 'returns minOccupancy value', ->
```
akegan: I think this test needs to be updated too

app/assets/javascripts/listings/components/browse-listings.html.slim
```diff
@@ -83,7 +83,7 @@ section.results-section#listing-results ng-if="!$ctrl.parent.openListings.length
     / Listings: Open Matched Listings
     .row.margin-top--2x
       .large-12.columns
-        property-card listing='listing' ng-repeat="listing in $ctrl.parent.openListings"
+        property-card listing='listing' ng-repeat="listing in $ctrl.parent.openMatchListings"
```
akegan: why did this update get made in this PR too?

akegan:
@chrisdolendo just did a quick QA on this and it looks like we're 1. missing a space between the number and "people" and 2. need to use "person" instead of "people" if it's "at least 1.."

image

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
```diff
@@ -12,8 +12,10 @@ angular.module('dahlia.components')
...
       @occupancy = (unitSummary) ->
-        return '1' if unitSummary.maxOccupancy == 1
-        unitSummary.minOccupancy + '-' + unitSummary.maxOccupancy
+        if unitSummary.minOccupancy == 1 || unitSummary.maxOccupancy == null
+          "at least #{unitSummary.minOccupancy}"
+        else
+          "#{unitSummary.minOccupancy}-#{unitSummary.maxOccupancy}"
 
       @occupancyLabel = (maxOccupancy) ->
         return $translate.instant('LISTINGS.PERSON') if maxOccupancy == 1
```
akegan: looks like this is what you can update to have it say person instead of people

chrisdolendo: @akegan @occupancyLabel is being used in multiple places and used in other ways which needs this to stay the way it is, and so I just updated @occupancy for this use case.

chrisdolendo:
@akegan there is a space there and the person, people is on there -- if you look at files changed and https://dahlia-full-pr-1115.herokuapp.com/listings/a0W0P00000DYxphUAD

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
```diff
@@ -12,8 +12,10 @@ angular.module('dahlia.components')
       @error = ListingPreferenceService.error
 
       @occupancy = (unitSummary) ->
-        return '1' if unitSummary.maxOccupancy == 1
-        unitSummary.minOccupancy + '-' + unitSummary.maxOccupancy
+        if unitSummary.minOccupancy == 1 || unitSummary.maxOccupancy == null
```
akegan: I think this logic is still wrong

### If max Occupancy is not null
We want
- "1 person" if min and max occupancy == 1
- "1-2 people" (minOccupancy-maxOccupancy) if min and max occupancy are both defined

### If max occupancy is null 
We want
- "at least 1 person" if minOccupancy is 1
- "at least 3 people" if minOccupancy > 1

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
akegan: oh I see. I think that we should update the function to make that clearer since it's not always maxOccupancy that's going in there.

spec/javascripts/components/listing/eligibilitySection_spec.coffee
```diff
@@ -46,12 +46,22 @@ do ->
         ctrl = $componentController 'eligibilitySection', locals, {parent: fakeParent}
 
       describe 'occupancy', ->
-        it 'returns 1 for SRO', ->
-          unitSummary = { minOccupancy: 1 , maxOccupancy: 1 }
-          expect(ctrl.occupancy(unitSummary)).toEqual('1')
-        it 'returns a range for all other unit types', ->
-          unitSummary = { minOccupancy: 1 , maxOccupancy: 3 }
-          expect(ctrl.occupancy(unitSummary)).toEqual('1-3')
+        describe 'maxOccupancy is 1', ->
```
akegan: nit, I think this one and is an SRO unit below are duplicates

app/assets/javascripts/listings/components/eligibilitySection.js.coffee
jrubenoff: @akegan Thanks for this!

spec/javascripts/components/listing/eligibilitySection_spec.coffee
chrisdolendo: Done! see latest commit

PR #1117, opened 2019-02-06
[#161808349] Remove address2 field

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/161808349


akegan:
Unrelated to this ticket, we just got some additional information about geocoding addresses that might shed some light on why this second address line existed in the first place. 

From charlie over at MOHCD:

> Chiming in on the errors and the addressing related issues. Most of the DAHLIA errors that came in yesterday and today are the same two applicants using non-conforming address.
There is an outstanding address with non valid address being validated by EZ post that cannot be geocode: General Delivery and Exact PO Box Numbers.
The locator cannot also cannot handle special character’s (/,#). The EZ post validator fixes most of these issues by creating a well formed address to pass through the locator service. However, the freeform apt entry field allows an applicant to include additional address details beyond what is validated that may be malformed, including the inclusion of additional “#”. For example, if I put in 901 Fairfax #307#B, I will get “901 FAIRFAX AVE APT 307 # B” as a validated address. Taken even further. I could also enter 901 Fairfax #307 and then enter “I live with Nancy Pelosi” and get “901 Fairfax #307 I live with Nancy Pelosi” as a validated address, which then gets saved into Salesforce.
In general, the locator itself does not do well with apartments. In that, including them may not result in a match because the apartment does not exist reference data. That’s why the web application strips them out before passing is through the locator service.  For some reason, this one is making it though (I think because it has two #s).
The three geocoder service errors that came this morning may be related to the server being unavailable for a short time (though it may also related to how it is being queried). It does happen from time to time and we have back up process for manually geocoding those. This back up process also address when an address is not found.

>For reference, the more critical path error is for the NRHP Boundary Check service, which would require a manual review of the application to determine preference eligibility. For the most part this has been addressed. The last error was back in June 2018
So a couple of recommendations
•    Prevent non valid address being validated by EZ post
•    Limit the number of special characters from being input in the freeform apt field or limit them entirely
•    Limit the total number of characters that can be input in the freeform apt field

It sounds like we pulled apartment number onto its own line to help with the geocoding through easypost. @callaghanc or @chrisdolendo do you have any historical knowledge on this?

callaghanc:
already answered the above question in conversation with @akegan, but to close the loop here, i don't have any historical knowledge about the configuration of the address1 and address2 fields. @akegan and i are continuing to discuss address- and geocoding-related issues.

akegan:
looks good so far. One note is that the alternate contact address is another mailing address that's not validated, so we probably want to remove the field from that address too

image

software-project:
I tested alice griffiths address and it looked fine (with validation and address2 field). Removed the field from the one you mentioned above.

PR #1118, opened 2019-02-08
[#163741697] Add missing $ sign

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/163741697/comments/199228605


software-project:
@akegan it was an easy fix. `$` was missing in front of `ctrl` in the view. Function itself is tested, so I don't think we need any more tests for that.

PR #1120, opened 2019-02-09
Chores/upgrade lodash #163837789

akegan:
https://www.pivotaltracker.com/story/show/163837789

https://github.com/angular-translate/grunt-angular-translate/issues/92

Turns out grunt-angular-translate looks like it's no longer being maintained. Someone opened a PR a couple of days ago to update lodash but it's unclear whether or not anyone is going to look at it. 

Thankfully it's just a dev dependency so hopefully won't actually impact our security (just our github alerts 😦 )


Gruntfile.js
```diff
@@ -100,7 +100,7 @@ module.exports = function(grunt) {
   // load tasks
   grunt.loadNpmTasks('grunt-contrib-clean');
   grunt.loadNpmTasks('grunt-contrib-copy');
-  grunt.loadNpmTasks('grunt-replace');
+  grunt.loadNpmTasks('grunt-replace-regex');
```
akegan: this is the same repo, just someone forked it so they could update the dependencies

callaghanc: ok cool, looks like the forker seems to be a lead at Esri, seems legit, we don't have any bitcoin to steal anyway so this should be fine 😛

PR #1121, opened 2019-02-10
Bugs/filename too long #163095454

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/163095454


app/assets/javascripts/short-form/components/preferenceProofUploader.js.coffee
```diff
@@ -75,6 +75,12 @@ angular.module('dahlia.components')
           proofOption = 'Lease'
         @application.preferences[@preference + '_proofOption'] = proofOption
 
+      @validateFileNameLength = ($file) ->
+        if $file.name.length > 85
```
akegan: I think the max length is actually 80

software-project: Updated to 80

akegan:
QAing this now (once the new deploy is complete)

akegan:
I just tested it out and I think it's a bit confusing as is. The validation message comes up before you've selected a file, then there's no change when a file with a filename that's too long is selected. 

Can we 1. Stop the error message from showing up before the file is selected and 2. Show an error message that's specific to the error (filename is too long)?
upload file length validation

jrubenoff:
Hey, very sorry for chiming in so late, but: why not truncate the file name on the front-end before submitting to Salesforce?

Then there would be no need to show an error.

akegan:
I'm not sure what we use the filename for on salesforce. I'm slightly worried that it might affect us being able to tell what the filetype is. I can look into how we validate file type and if salesforce cares about the file extension

jrubenoff:
Couldn't we use a regex to only truncate the text before the extension?

software-project:
@jrubenoff I'm not sure if it's possible due to file read access but I can check. For now I made the error messages better.

jrubenoff:
Andrea asked over Slack:

> what's the benefit of truncating the filename vs showing an error?

One principle of interaction design is that the computer should help you save time. (See [Law #2 here.](http://uxdiogenes.com/blog/the-three-laws-of-interaction-design))

The filename length doesn't affect an applicant's chances for a listing. So, making the user shorten it is more work than is strictly necessary.

Also, we show errors as bright red text. This grabs the user's attention, which is good: they'll understand that they did something wrong, and hopefully understand how to fix it. But from an emotional perspective, if the user sees lots of errors, the overall user experience can come off as negative.

In general, if we give users the option to violate our technical constraints in some way, we should try to hide or remove those constraints before resorting to an error message.

software-project:
@jrubenoff I checked yesterday and we can't change file name on the javascript side. It is possible on the Ruby side. I understand your angle here and I agree that it's always good to make the experience for the user better. So I would says it's your call guys here if you want to improve this edge case. We can always put it into icebox for later if there are more pressing tasks.

jrubenoff:
Alright. Showing an error is definitely better than nothing at all! I was just explaining my rationale above.

app/assets/javascripts/short-form/components/preferenceProofUploader.js.coffee
```diff
@@ -75,6 +75,18 @@ angular.module('dahlia.components')
           proofOption = 'Lease'
         @application.preferences[@preference + '_proofOption'] = proofOption
 
+      @validateFileNameLength = ($file) ->
+        if $file
+          if $file.name.length > 80
+            @proofDocument.error = 'ERROR.FILE_NAME_TOO_LONG'
+          else
+            if $file.size > 5 * 1000 * 1000 # 5MB
```
akegan: nice

akegan:
I'm still hung up on making a decision on this because I don't think I have all of the information I need. My outstanding questions are:

1. Do we use the filename for anything on Salesforce?
1. What would be the impact of shortening the filename on the ruby side? Does it have the possibility of causing issues if a user saves a draft then continues the application later?

Where I'm at now:  I see the value of not showing the user an error, and instead just letting them through, but I also want to avoid manipulating anything that a user inputs in unexpected ways.

software-project:
@akegan I don't know about point 1. but regarding point 2. in theory it should be fine. We can change file name with no problem. It just adds a little more complexity. Shouldnt be more than 3 lines of code.

akegan:
I was able to test out the error validation:
image

I was also able to confirm that if you click the button again you can upload a file with a shorter name.

PR #1122, opened 2019-02-11
[#163574689] Remove stubbed salesforce data for listings

software-project:
https://www.pivotaltracker.com/story/show/163574689

### DIFF
--- a/app/services/force/listing_service.rb
+++ b/app/services/force/listing_service.rb
@@ -23,19 +23,13 @@ module Force
       Does_Match
       LastModifiedDate
       imageURL
-      Tenure
     ].freeze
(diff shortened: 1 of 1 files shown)


app/services/force/listing_service.rb
```diff
@@ -23,19 +23,13 @@ class ListingService
       Does_Match
       LastModifiedDate
       imageURL
-      Tenure
```
software-project: Tenure was duplicated, see line 9

software-project:
To fully test it I think we need to merge additional info section in: https://github.com/Exygy/sf-dahlia-web/pull/1116

PR #1123, opened 2019-02-11
[#163766486] Show only rental listings for eligibility

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/163766486


software-project:
@akegan I just pushed a fix. So now then pulling sale listing eligibility filters are reset. Added a small test for it as well.

spec/services/listing_service_spec.rb
```diff
@@ -59,4 +58,22 @@
...
+  describe '.eligible_listings' do
+    let(:filters) do
+      {
+        householdsize: 1,
+        incomelevel: 18_000,
+        childrenUnder6: 0,
+      }
+    end
+
+    it 'returns only rental listing' do
```
akegan: typo listing -> listings

software-project: Fixed

PR #1124, opened 2019-02-11
[#163669090] Replace leasing agent with property agent

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/163669090


jrubenoff:
Looks great! 🙌

akegan:
my only concern here is that we have some manual translations for some of these values. 

It looks like in filipino leasing agent was translated to leasing agent, but in spanish it looks like it's "El agente de arrendamiento" (arrendamiento = lease), and I don't know what it is in chinese.

jrubenoff:
We don't really have a process. We should definitely create one, but avoid tying it to our ownership deadlines.

In the meantime, would you like to create separate stories to update translations as they become outdated?

akegan:
here's a ticket to follow-up with translations https://www.pivotaltracker.com/story/show/163958269. I also chatted with chris about translations this morning and added a ticket to do a proof of concept around using the POEditor API instead of downloading and committing the translations https://www.pivotaltracker.com/story/show/163953668

PR #1127, opened 2019-02-13
[#163917081] Add MLS link to listings process section

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/163917081


akegan:
looks pretty good, only thing that would be nice is to add a target _blank so the site opens in a new tab

software-project:
@akegan added. I was thinking about it when I was reading the story, but forgot to execute.

PR #1130, opened 2019-02-13
[#162565298] Add "Before / after applying" content to sale listings page

jrubenoff:
[See my latest comment](https://www.pivotaltracker.com/story/show/162565298/comments/199530021) for a list of changes.


app/assets/javascripts/listings/components/before-applying-list.html.slim
```diff
@@ -0,0 +1,6 @@
+ol.numbered-list
```
akegan: Since there isn't a component associated with this, this file is a partial and should go under 

app/javascripts/listings/templates/_before-applying-list.html.slim

app/assets/json/translations/locale-en.json
```diff
@@ -732,6 +732,7 @@
         "PRIORITY_UNITS_DESCRIPTION": "This building has units set aside if any of the following apply to you or someone in your household:",
         "REMAINING_UNITS_AFTER_PREFERENCE_CONSIDERATION": "After all preference holders have been considered, any remaining units will be available to qualified applicants in lottery order.",
         "REQUIRED_DOCUMENTS": "Required Documents",
+        "REQUIRED_DOCUMENTS_AFTER_APPLYING": "If you are selected in the lottery, we will ask you to <a href='{{url}}'>provide up to date financial documents</a> on short notice.",
```
akegan: I think we generally try to have links to external pages open in a new tab. You can do that by adding target="_blank"  after the href

jrubenoff: It looks like we haven't done this for any ownership-related links yet in `locale-en.json`, so I'll fix that in this PR.

jrubenoff:
@akegan Cool, just made both changes.

PR #1132, opened 2019-02-15
Features/add prerequisites page #162769857

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/162769857


app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -669,6 +669,27 @@
...
+        ($stateParams, ShortFormApplicationService, AccountService, AutosaveService) ->
+          ShortFormApplicationService.completeSection('Intro')
+          if AccountService.loggedIn()
+            ShortFormApplicationService.importUserData(AccountService.loggedInUser)
+            ShortFormApplicationService.infoChanged = $stateParams.infoChanged
+            # always autosave when you start a new application
+            # TODO: remove hotfix for marking initial autosaves that come from the Name page
+            unless ShortFormApplicationService.application.id
+              ShortFormApplicationService.submitApplication({autosave: true, initialSave: true})
+      ]
```
software-project: @callaghanc I'm not sure if all of those things here are needed. Could you have a look?

callaghanc: @software-project this onEnter block makes sense. for ownership applications, this is the first page of the application where the user begins to enter information. so, marking the Intro section as complete makes sense because the intro is over at this point and the user is now filling out the app. the autosave also could make sense, because the current intention with autosave is to save the user's application when they start it. however, it's a little redundant with the autosave on the Name page. for ownership applications, this will make us do two autosaves. we'll still just be doing one for rental apps. this isn't a big problem, but it would be a little nicer to only have one autosave-at-start per app. so we could:

a) add some logic to not autosave on the Name page if we already autosaved on the Prerequisites page, or to not autosave on the Name page if it's an ownership application, or
b) remove the autosave from the Prerequisites page and only autosave on the Name page.

keeping autosave on the Name page only is less complicated, but we'd have to decide whether we are fine with not autosaving at the technical first-filled-out page of the ownership app. @akegan do you have a preference from a business rules/UX perspective?

akegan: It's a really interesting question because I believe autosave is actually totally disabled for the entire app currently, and we already had to disable it further in (this ticket)[https://www.pivotaltracker.com/story/show/159252180].

In terms of the logic of where to save if we have it saving at the start of the app, I'm not sure. If it's just the checkboxes, then the user isn't really losing anything if it's not saved, but if they've uploaded multiple documents then that's significant time already spent.

tl;dr, I'm not sure! I can follow up with josh

akegan: I think we actually disabled it completely in this PR https://github.com/Exygy/sf-dahlia-web/commit/0a1b65a8609db124cc342cff0ed38ea80ee4757b#diff-bf948082a3b5ee08694d1e3427c28a7fL721

callaghanc: hmm then do we want to keep an autosave call in any onEnters at all? what's the roadmap looking like for re-introducing use of autosave?

akegan: let's just leave it on the name page for now. It's not on the roadmap at all yet, but my hope is that it might be possible once 8cloud comes on board and starts working on making the API better

software-project: 👍

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
```diff
@@ -272,6 +275,8 @@ ShortFormNavigationService = (
           'preferences-programs'
       when 'review-submitted'
         'confirmation'
+      when 'name'
```
akegan: why don't we have to put the check for isSale here?

spec/e2e/features/short_form/filling_sale_application.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Short Form Application
+    As a web user
+    I should be able to fill out the short form application
+    In order to apply online to a sale listing
+
+    Scenario: Submitting a basic application, creating an account on the confirmation page
+        Given I go to the first page of the "Sale Test Listing" application
+        When I hit the Next button "1" time
+        And I wait "5" seconds
```
akegan: Did you add this in because without the wait it fails? have we figured out any other way to wait for a page to load?

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
software-project: Because on name page on rental listing, back button is not displayed.

spec/e2e/features/short_form/filling_sale_application.feature
software-project: I put it when I was trying to make it work. Now it works without it as well, so will remove it.

akegan: cool. hopefully it won't end up being super flaky 😕

software-project: Removed. yeah, I would hate to see it as well

PR #1142, opened 2019-02-22
Features/add prerequisites dropdowns #162770324

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/162770324


akegan:
@software-project is this branched off of the checkboxes PR? Should we change the base for now to make the changes in this PR clearer?

software-project:
@akegan done. It's work in progress as SF part is not ready.

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -551,7 +551,7 @@ ShortFormApplicationController = (
       return
     ShortFormApplicationService.checkHouseholdEligiblity($scope.listing)
       .then( (response) ->
-        # TODO remove this when SF household API is fixed for sale listing
+        # TODO remove this then SF household API is fixed for sale listing
```
akegan: looks like a merge went wrong here

app/assets/javascripts/short-form/components/lendingInstitution.js.coffee
```diff
@@ -0,0 +1,46 @@
...
+          field.$invalid && (field.$touched || form.$submitted)
+        else
+          false
+
+      @lendingInstitutionsNames = ->
+        _.keys(@lendingInstitutions)
+
+      @onChangeLendingInstitution = ->
+        @application.lendingAgent = null
+        @officers = @lendingInstitutions[@selectedInstitution]
```
akegan: I think we should settle on just using lending agents/ agents for all of the variable names instead of loan officer. My understanding is that we're going to update the content to avoid the work "loan officer" at some point too.

app/assets/javascripts/short-form/components/lendingInstitution.js.coffee
```diff
@@ -0,0 +1,46 @@
...
+
+      @officesSelectActive = ->
```
akegan: looks like a typo in the name of this function. I'm not super clear on what its purpose is either

app/services/force/short_form_service.rb
```diff
@@ -120,6 +120,25 @@ def self.submitted?(application)
...
 
+    def self.lending_institutions
+      endpoint = '/services/apexrest/agents/'
+      institutions = {}
+
+      Request.new.cached_get(endpoint).each do |institution|
+        next unless institution['Contacts'] && institution['Contacts']['records']
+        agents = institution['Contacts']['records'].each_with_object([]) do |array, agent|
+          if agent['BMR_Certified__c']
+            status = agent['Lending_Agent_Status__c'].present? &&
```
akegan: Can you clarify to me what you're doing with the status here and what the goal is?

app/assets/javascripts/short-form/components/lendingInstitution.js.coffee
software-project: Sure, that makes sense.

akegan:
changes look good. just doing some QA now
These error messages are missing an "a" in the middle of them. I think we could probably get away with just using the standard "this field is required" message, right @jrubenoff?
image

akegan:
Other than that QA looks good. 

I've confirmed that I'm seeing the lending agents we should be seeing in the dropdowns, and that if i fill out an application the lending agent name shows up on the submitted application!

jrubenoff:
@akegan @software-project Let's stay consistent with [our current error message i18n methods](https://github.com/Exygy/sf-dahlia-web/blob/master/app/assets/json/translations/locale-en.json#L345), and use a unique error message for both fields:

> Please select a lending institution
> Please select a loan officer

software-project:
@akegan @jrubenoff lables are updated, so we should be good now.

akegan:
@software-project I think that josh was suggesting that we should just have 2 separate messages that are 
`Please select a lending institution`
and
`Please select a loan officer`
rather than having "please select a" and then adding the field name. My guess is that it's because when we translate something like that the word order can change and the sentence might not make sense anymore in other languages if we separate them

akegan:
should be good to merge once tests pass

PR #1144, opened 2019-02-23
Features/6 mint content changes #164181844

akegan:
https://www.pivotaltracker.com/story/show/164181844

### DIFF
--- a/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
+++ b/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
@@ -2,7 +2,8 @@ accordion-heading.lead
   img.accordion-icon ng-src="{{::$ctrl.parent.assetPaths['listing-eligibility.svg']}}" alt="eligibility-notebook"
   h3.accordion-name
     | Eligibility
-  span.accordion-subheader Income, occupancy, preferences, and subsidies
+  span.accordion-subheader ng-show="::!$ctrl.parent.isSale($ctrl.parent.listing)"
+    | Income, occupancy, preferences, and subsidies
   span.ui-control.ui-icon.ui-small
     svg
       use xlink:href="#i-arrow-down"
--- a/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
+++ b/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
@@ -78,7 +78,7 @@ accordion-heading.lead
           p.custom-counter_text ng-if="::!$ctrl.listingHasOnlySROUnits($ctrl.parent.listing) && !$ctrl.parent.listingHasSROUnits($ctrl.parent.listing)"
             span ng-if="::$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
               | {{'LISTINGS.OCCUPANCY_SEE_UNIT_INFO' | translate}}
-            span ng-if="::!$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
+            span ng-if="::!$ctrl.parent.isSale($ctrl.parent.listing) && !$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
               | {{'LISTINGS.OCCUPANCY_DESCRIPTION_NO_SRO' | translate}}
 
         .scrollable-table-container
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
```diff
@@ -1,18 +1,17 @@
...
-  ' Applicants will be contacted by the {{$ctrl.parent.isSale($ctrl.parent.listing) ? 'sales agent' : 'leasing agent'}}
+  ' Applicants will be contacted by the property agent
   span ng-if="!$ctrl.parent.isFirstComeFirstServe($ctrl.parent.listing)"
     | in lottery rank order until vacancies are filled.
   span ng-if="$ctrl.parent.isFirstComeFirstServe($ctrl.parent.listing)"
     | on a first-come, first-serve basis until the vacancy is filled.
 .show-for-print ng-show="$ctrl.parent.toggleStates[$ctrl.parent.listing.Id].whatToExpectOpen"
-  p.c-steel.t-small
-    | All of the information that you have provided must be verified, and your eligibility confirmed.  Any fraudulent statements will cause your application to be removed from the lottery.  You will also be removed from the lottery if you cannot prove a housing preference, or if a household member appears on more than one application.
+  p.c-steel.t-small translate="F2_REVIEW_TERMS.P3"
```
akegan: What are our thoughts on re-using text from application stuff elsewhere? I can't decide if it's worse to have the same language copy and pasted in two places, or if it's worse to have someone update the content assuming it's just for the application then have it have unforeseen repercussions.

app/assets/json/translations/locale-en.json
```diff
@@ -653,7 +653,7 @@
             "READ_FULL_LIST": "Read <a href='{{url}}' target='_blank'>our full list of requirements</a> for more details.",
             "STEP_1": "Haven't owned property in the past 3 years",
             "STEP_2": "Meet our <a href='{{url}}' target='_blank'>income requirements</a>",
-            "STEP_3": "Complete a <a href='{{url}}' target='_blank'>homebuyers’ education workshop</a>",
+            "STEP_3": "Complete <a href='{{url}}' target='_blank'>homebuyer education</a>",
```
akegan: I wonder if this would look better if we call it "First-time homebuyer education" instead of just "homebuyer education"

app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
```diff
@@ -77,7 +78,7 @@ accordion-heading.lead
           p.custom-counter_text ng-if="::!$ctrl.listingHasOnlySROUnits($ctrl.parent.listing) && !$ctrl.parent.listingHasSROUnits($ctrl.parent.listing)"
             span ng-if="::$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
               | {{'LISTINGS.OCCUPANCY_SEE_UNIT_INFO' | translate}}
-            span ng-if="::!$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
+            span ng-if="::!$ctrl.parent.isSale($ctrl.parent.listing) && !$ctrl.parent.listingIs('Transbay Block 7', $ctrl.parent.listing)"
```
akegan: This is a hacky way of not showing any text for sales in the occupancy section. We may choose to change the text on Monday and then I can add it to locale-en.json

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
callaghanc: if it's the same piece of text conceptually, then it makes sense to use the same translation term. i.e., if the intention is that we should use the same block of text to convey the concept of "information veracity requirements" on both the listing and the application, then we should use the same translation term. but if the two pieces of text are not meant to be the same conceptually, then it would be better to use separate terms. could product/design help clarify the intention in this particular case?

callaghanc:
@akegan what's the status of this PR? should we reach out to product/design to answer any of the questions above?

akegan:
@callaghanc emily is looking through it now to make sure it all makes sense

app/assets/javascripts/listings/templates/listing/_what_to_expect.html.slim
akegan: Ok. checked with Emily. I'm going to copy and paste because 3rd paragraph of Review terms doesn't communicate the concept so there's a high risk of unintended consequences.

app/assets/javascripts/listings/components/listing/eligibility-section.html.slim
akegan: We're going to keep this text as is

akegan:
@callaghanc this is ready for review now

akegan:
Emily is going to QA as part of release testing

PR #1145, opened 2019-02-25
[#164174922] Show full AMI table in Maximum occupancy

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/164174922

### DIFF
--- a/app/assets/javascripts/listings/ListingEligibilityService.js.coffee
+++ b/app/assets/javascripts/listings/ListingEligibilityService.js.coffee
@@ -32,17 +32,20 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
     else
       parseFloat(Service.eligibility_filters.income_total)
 
-  Service.occupancyMinMax = (listing) ->
+  Service.occupancyMinMax = (listing, amiLevel) ->
     minMax = [1, 1]
     if listing.unitSummary
       min = _.min(_.map(listing.unitSummary, 'minOccupancy')) || 1
-      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || 2
+      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || Service.defaultMaxOccupancy(amiLevel)
       minMax = [min, max]
     return minMax
 
+  Service.defaultMaxOccupancy = (amiLevel) ->
+    _.max(_.map(amiLevel.values, (v) -> v.numOfHousehold))
+
   Service.occupancyIncomeLevels = (listing, amiLevel) ->
     return [] unless amiLevel
-    occupancyMinMax = Service.occupancyMinMax(listing)
+    occupancyMinMax = Service.occupancyMinMax(listing, amiLevel)
     min = occupancyMinMax[0]
     # We add '+ 2' for 2 children under 6 as part of householdsize but not occupancy
     max = occupancyMinMax[1] + 2
@@ -55,6 +58,8 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
       max = 2
     else if ListingUnitService.listingHasOnlySROUnits(listing)
       max = 1
+    # if ListingIdentityService.isSale(listing)
+    #   max = _.max(_.map(amiLevel.values, (v) -> v.numOfHousehold))
     _.filter amiLevel.values, (value) ->
       # where numOfHousehold >= min && <= max
       value.numOfHousehold >= min && value.numOfHousehold <= max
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/listings/ListingEligibilityService.js.coffee
```diff
@@ -32,17 +32,20 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
...
+  Service.occupancyMinMax = (listing, amiLevel) ->
     minMax = [1, 1]
     if listing.unitSummary
       min = _.min(_.map(listing.unitSummary, 'minOccupancy')) || 1
-      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || 2
+      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || Service.defaultMaxOccupancy(amiLevel)
       minMax = [min, max]
     return minMax
 
+  Service.defaultMaxOccupancy = (amiLevel) ->
```
akegan: not super clear on what this is. It's saying that the max is whatever the max value is in the AMI table?

app/assets/javascripts/listings/ListingEligibilityService.js.coffee
```diff
@@ -32,17 +32,20 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
...
+    _.max(_.map(amiLevel.values, (v) -> v.numOfHousehold))
+
   Service.occupancyIncomeLevels = (listing, amiLevel) ->
     return [] unless amiLevel
-    occupancyMinMax = Service.occupancyMinMax(listing)
+    occupancyMinMax = Service.occupancyMinMax(listing, amiLevel)
     min = occupancyMinMax[0]
     # We add '+ 2' for 2 children under 6 as part of householdsize but not occupancy
     max = occupancyMinMax[1] + 2
```
akegan: seems like this adding 2 might not work in this case right?

software-project:
@akegan let me maybe first address your notes.

software-project:
@akegan I renamed the function and fixed the issue with '+2'

app/assets/javascripts/listings/ListingEligibilityService.js.coffee
```diff
@@ -36,19 +36,19 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
     minMax = [1, 1]
     if listing.unitSummary
       min = _.min(_.map(listing.unitSummary, 'minOccupancy')) || 1
-      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || Service.defaultMaxOccupancy(amiLevel)
+      max = _.max(_.map(listing.unitSummary, 'maxOccupancy')) || Service.maxNumOfHousehold(amiLevel)
       minMax = [min, max]
     return minMax
 
-  Service.defaultMaxOccupancy = (amiLevel) ->
+  Service.maxNumOfHousehold = (amiLevel) ->
```
akegan: I think what this function is checking for, is what the max household number that's available on the AMI chart. 

maybe this would be clear if it's like maxAmiNumHousehold or something like that? I can't think of anything better at the moment 🙁

app/assets/javascripts/listings/ListingEligibilityService.js.coffee
```diff
@@ -36,19 +36,19 @@ ListingEligibilityService = ($localStorage, ListingIdentityService, ListingUnitS
...
     _.max(_.map(amiLevel.values, (v) -> v.numOfHousehold))
 
   Service.occupancyIncomeLevels = (listing, amiLevel) ->
     return [] unless amiLevel
     occupancyMinMax = Service.occupancyMinMax(listing, amiLevel)
     min = occupancyMinMax[0]
-    # We add '+ 2' for 2 children under 6 as part of householdsize but not occupancy
-    max = occupancyMinMax[1] + 2
+    # We add '+ 2' for 2 children under 6 as part of householdsize but not occupancy. Unless it's max
```
akegan: it seems like these two functions are both taking different weird responsibilities for figuring out this number. I wonder if occupancyMinMax should just return null if maxOccupancy is null, then we can deal with the logic of figuring out the max based on the AMI level and whether or not to add 2 in this function

akegan:
code looks good. working on QA now

akegan:
caught a small bug for homeownership acres
Since there are 2 different AMI charts defined for this listing, one has up to 9 household members, and one has only up to 8.

I think the solution is that if there isn't a value available, we should just not show the "/year"
image

software-project:
@akegan good spot, on it

PR #1148, opened 2019-02-26
Chores/upgrade ruby #159945905

software-project:
https://www.pivotaltracker.com/story/show/159945905


spec/javascripts/fixtures/json/gis-api-valid-address-boundary-match.json
```diff
@@ -1 +1 @@
-{"gis_data":{"address":"1222 HARRISON ST, San Francisco, CA","location":{"x":-13626517.2840026,"y":4547486.54839139},"score":100,"attributes":{"Loc_name":"eas"},"extent":{"xmin":-13626717.2840026,"ymin":4547286.54839139,"xmax":-13626317.2840026,"ymax":4547686.54839139},"boundary_match":true}}
```
akegan: did float precision change in ruby 2.5 or something?

software-project: Seems like it: https://nithinbekal.com/posts/ruby-2-4-features/

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.7'
+ruby '2.5.1'
```
akegan: only comment is wondering if there's a higher patch version for ruby. I think 2.5.1 was the latest stable when we made the ticket. Seems like 2.6 is a little too new to be supported, but maybe if there's a higher 2.5.x we could use that version

software-project:
There might be issues with SSL. Will check on heroku how is it looking

Gemfile
akegan: nevermind, looks like 2.5.1 is the latest

akegan:
I updated the semaphore job to use 2.5.1 to test if the tests pass

Gemfile
callaghanc: @akegan there's a 2.5.3

callaghanc: https://www.ruby-lang.org/en/news/2018/10/18/ruby-2-5-3-released/

software-project: I'm using 2.5.3 locally for couple project and it's fine. I upgraded to 2.5.1 as it was specified in the task.

callaghanc: cool, well then @software-project go ahead and upgrade to 2.5.3 in this PR and check that everything still works. i believe 2.5.1 was specified in the task because that was the latest 2.5.x at the time

akegan: wait, it looks like semaphore doesn't have 2.5.3 
image

callaghanc: oh that's a bit annoying. we can still install it ourselves in our setup steps though

callaghanc: hmm rbenv doesn't seem to offer 2.5.3 for some reason, that would have been my plan for installing manually on semaphore, i'm going to look into that a bit to see if there is some reason why

software-project: hmmm seems like heroku doesnt like 2.5.1. Deploy is successful but app is down

akegan: may need to upgrade the buildpack

callaghanc: ok looks like i just had to update my ruby-build, i now see 2.5.3 available through rbenv. we'll have to check that semaphore's ruby-build is also up-to-date

software-project: Ok I see why

callaghanc: haha i think we all commented at the exact same time

callaghanc: @software-project we can use the instructions at the bottom of these [semaphore docs about ruby versions](https://semaphoreci.com/docs/languages/ruby/ruby-support-on-semaphore.html) to add installation of ruby 2.5.3 to our semaphore set up

software-project:
@akegan @callaghanc good news. Tests are green, semaphore set and running. I made couple fixes related to an issue with render_error (same method Faraday was using) and Devise. So far everything loos fine.

akegan:
@software-project did you end up having to update heroku at all?
interesting, looks like code climate doesn't like that your changes aren't covered by tests
image
image

What should we do?

software-project:
@akegan, nope. No changes on heroku side. Regarding code climate I'm not sure. But truth be told we dont have good coverage out there, so maybe we can use this opportunity to start adding tests in places we 'touch'? @callaghanc @chrisdolendo what do you think?

akegan:
I could go either way. If you have the bandwidth to add a couple tests for these functions that would be great!

callaghanc:
@software-project @akegan i think it would be great to start adding coverage to those areas! @software-project you can keep it limited to as much as you have bandwidth for and as much as will still let us move forward in a timely fashion

PR #1152, opened 2019-03-01
Chores/comment out sale listing e2e tests #164323713

akegan:
https://www.pivotaltracker.com/story/show/164323713

We'll follow-up in https://www.pivotaltracker.com/story/show/164323644 to uncomment the tests when we have a suitable prod listing to test against


akegan:
@chrisdolendo it looks like the issues found by code climate in this PR are totally unrelated to the diff and might be an issue because I initially opened up the PR against master instead of production. I'll try closing the PR and re-opening to see if that fixes it.

PR #1153, opened 2019-03-01
Chores/Comment out sale listing E2E tests #164323713

akegan:
https://www.pivotaltracker.com/story/show/164323644

Opened up a second PR to see if I could get codeclimate to pass but that doesn't seem to have worked.


akegan:
Code climate still seems to be comparing my branch to master. My vote would be to ignore the codeclimate check on this PR.

callaghanc:
Closing because we managed to get a test sale listing created for us on Prod, and so we will be using the sale listing application e2e tests after all.

PR #1158, opened 2019-03-14
Remove ADA page from sale app and do small refactor of short form navigation #164252126

callaghanc:
https://www.pivotaltracker.com/story/show/164252126


callaghanc:
@akegan This still needs some more test work, but you can start taking a look at this now.

spec/javascripts/controllers/short_form_application_controller_spec.coffee
```diff
@@ -482,40 +545,40 @@ do ->
...
-        it 'routes user to live work preferences page', ->
-          fakeShortFormApplicationService.eligibleForRentBurden = jasmine.createSpy().and.returnValue(false)
-          fakeShortFormApplicationService.eligibleForNRHP = jasmine.createSpy().and.returnValue(false)
-          fakeShortFormApplicationService.eligibleForLiveWork = jasmine.createSpy().and.returnValue(true)
-          scope.checkIfPreferencesApply()
-          path = 'dahlia.short-form-application.live-work-preference'
-          expect(state.go).toHaveBeenCalledWith(path)
+          expect(fakeShortFormNavigationService.goToAndTrackFormSuccess).toHaveBeenCalledWith(path)
+
+      # describe 'preferences do not apply to household',->
```
callaghanc: working on updating these commented out ones

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -332,30 +328,26 @@ ShortFormApplicationController = (
   $scope.getLandingPage = (section) ->
     ShortFormNavigationService.getLandingPage(section)
 
-  $scope.goToLandingPage = (section) ->
```
akegan: I wonder if this would be clearer if we called it goToSection, and then have a docstring that says "go to first page in section".

Looking at it it took me a while to figure out what a landing page is

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -144,18 +144,14 @@ ShortFormApplicationController = (
...
+      options.callbacks.forEach (callback) ->
+        callback()
+    if options.scopedCallbacks
+      options.scopedCallbacks.forEach (scopedCallback) ->
+        $scope[scopedCallback.func](scopedCallback.param) if $scope[scopedCallback.func]
     if options.path
-      $scope.goToAndTrackFormSuccess(options.path)
-
-  $scope.goToAndTrackFormSuccess = (path, params) ->
-    AnalyticsService.trackFormSuccess('Application')
```
akegan: Seems like the analytics service here is a bit of a side-effect of this function, not actually part of the goal. Would it be clearer if we just called it goToPage? I wasn't sure what "TrackFormSuccess" was just looking at where we're calling the function.

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
```diff
@@ -1,61 +1,147 @@
...
+    'household-member-form-edit':
+      scopedCallbacks: [
+        {func: 'addHouseholdMember'}
+        {func: 'checkPreferenceEligibility'}
+      ]
+    'household-member-verify-address':
+      scopedCallbacks: [{func: 'checkPreferenceEligibility'}]
+      path: 'household-members'
+    'household-public-housing': {scopedCallbacks: [{func: 'checkIfPublicHousing'}]}
+    'household-monthly-rent': {scopedCallbacks: [{func: 'checkIfReservedUnits'}]}
```
akegan: isn't HH monthly rent just if rent burdened is a pref? why is it checking for reserved units?

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
```diff
@@ -1,61 +1,147 @@
...
+    'household-reserved-units-veteran':
+      scopedCallbacks: [{
+        func: 'checkIfReservedUnits'
+        param: RESERVED_TYPES.DISABLED
+      }]
+    'household-reserved-units-disabled':
+      path: Service.getPostReservedPage(ShortFormApplicationService.listing)
```
akegan: I think I might need another pairing session to understand the navigation service. In this case, I think it might actually be clearer to put in a ternary statement rather than calling this function, but I am open to feedback!

I also think I don't fully understand what the difference between callbacks and paths are wonder if it'd be clearer to call these "nextPage" instead of "path".

callaghanc: @akegan and i discussed this over slack call and decided that this is as far as we have time to take the navigation refactor for now. further refactor will be done in the future as we continue to work on the short form.

app/assets/javascripts/short-form/ShortFormNavigationService.js.coffee
callaghanc: @akegan and i discussed this over slack call and went over how the short form navigation checks whether it needs to display pages for reserved units

PR #1162, opened 2019-03-19
Fix request chaining when switching pages quickly #163744941

chrisdolendo:


chrisdolendo:
@software-project Agree not seeing how we can test - an e2e test won't work, I thought of unit testing the params fed into http calls (to include a timeout) but I don't see a way to do that from what I've found online doesn't seem to be possible. Not easy to test the $q cycle either.. Will merge

PR #1168, opened 2019-03-27
Fix failing jasmine tests for ExternalTranslateService #164937511

akegan:
https://www.pivotaltracker.com/story/show/164937511

Our Prod semaphore job doesn't run jasmine tests, which is why I missed this issue.


akegan:
We should follow-up later with tests for my fix, but I'm not sure it's worth investing the time in it right now

PR #1176, opened 2019-04-01
Prevent access to other sections of Sale application before filling out prerequisites #164807951

chrisdolendo:
Story: https://www.pivotaltracker.com/story/show/164807951

Because of the structure of the sections and the mechanism of checking the authorization to proceed, I had to pull out the prerequisites page into its own section -- which I called the "Qualify" section -- as opposed to including it in the same section as the Name page. We may need to check with @jrubenoff if he is fine with the naming of this section.

I also had to create measures to ensure Rental applications still do not have access to the prerequisites section and do not have Qualify as a section in the process.

To keep our tests lean, I chose to focus on unit tests.


akegan:
@chrisdolendo not sure if this is ready for review yet, but I was still able to skip to the name page via updating my url and submit an application even though I didn't fill out the "Qualify section"

jrubenoff:
@chrisdolendo So this is inconsistent with how we currently handle qualifications for [senior listings.](https://dahlia-full-pr-1176.herokuapp.com/listings/a0W0P00000GwGl3UAF/apply-welcome/community-screening) Would it be possible to use the same design pattern?

If we can't, the new section makes the progress bar [overflow on mobile](https://user-images.githubusercontent.com/1328849/55346440-566f2500-5467-11e9-8489-ba2190f9bd5c.png), so we'd need to redesign that component before shipping this.

chrisdolendo:
@jrubenoff Right, I was initially implementing that solution (to use the same pattern for seniors) but the reason why it has to be within the `short-form-application` namespace is because there is data on that form that needs to be saved to the application (e.g. we save files on this page) - unlike the senior check - that, and autofill logic and loading draft logic lives within that  `short-form-application` namespace. 

Let me see if there is an alternative solution and get back to you before considering a design update.

jrubenoff:
Makes sense!

chrisdolendo:
@akegan I fixed the issue with this commit: https://github.com/Exygy/sf-dahlia-web/pull/1176/commits/8bc5b68be682625eebcc6e94f0ef07681d920561

Now on design:
@jrubenoff I hid the qualify section on the nav to make it look like it's part of the intro, and so the nav has its original elements in it. The only small issue with this is that if the user wants to go back and change those prerequisite, they'll have to click on "You" on the nav and then click "Back" button on the name page. Not the worst thing honestly, but not the best flow. Check it out on `/listings/a0W0P00000GlKfBUAV/apply/name`

Other options is to:
1. Update the UI to accommodate the Qualify section because to me it does seem like it belongs in its own section and not the You section
1. Refactor the whole access logic -- there are several ways we can refactor but it will take a full day or two to implement and QA

Let me know your thoughts.

chrisdolendo:
@akegan @jrubenoff 

Tests should be fixed now, overflow on short form nav incorporated.

akegan:
ok @chrisdolendo I'll start testing now

PR #1178, opened 2019-04-01
Fix devise password change error #165028596

software-project:
https://www.pivotaltracker.com/story/show/165028596

### DIFF
--- a/spec/controllers/overrides/passwords_controller_spec.rb
+++ b/spec/controllers/overrides/passwords_controller_spec.rb
@@ -1,0 +1,38 @@
+require 'rails_helper'
+
+describe Overrides::PasswordsController do
+  before(:each) do
+    @request.env['devise.mapping'] = Devise.mappings[:user]
+  end
+
+  let!(:user) do
+    @user ||= User.create(
+      email: 'jane@doe.com',
+      password: 'somepassword',
+      password_confirmation: 'somepassword',
+    )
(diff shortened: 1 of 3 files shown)


spec/controllers/overrides/passwords_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  describe '#edit' do
+    let(:token) do
+      allow_any_instance_of(User).to receive(:send_devise_notification).and_return(true)
+      @token ||= user.send_reset_password_instructions
+    end
+
+    it 'change allow_password_change to true' do
```
akegan: i'm a little confused about the logic around allow_password_change, why do we set it to true on edit?

akegan:
are there any steps we need to do on deploy to handle the DB update?

spec/controllers/overrides/passwords_controller_spec.rb
software-project: This is devise behavior. We are expecting it to be set to true. I presume it has something to do with safety. This functionality is available for unsigned users, so without token unuathorized user is not able to change password for some other account.

software-project:
@akegan just run the db migration. I believe heroku doest it for us.

akegan:
got QA sign-off from emily: merging and deploying tonight

PR #1189, opened 2019-04-16
Limit eligibity estimator results per listing type #163100134

software-project:
https://www.pivotaltracker.com/story/show/163100134


akegan:
looks like this text needs to be updated to match the screenshot in the comments of the ticket 
image

It also looks like the titles should be "Showing matching units for rent" / "Showing matching units for sale" instead of Showing matches for sale/rental

I also wasn't able to get our test sale listing to show up as a match for me. Max income is 45k and there's no min income so I should be able to get a match. Can you investigate what's going on? It might be that salesforce eligibility isn't working for sale listings?

software-project:
@akegan I fixed the labels, sorry I didn't notice the comment. Regarding results, it's an interesting way they work, cause we get all the listings there are (or at least 52). If they match the criteria, they have variable doesMatch set to true. For sale listings it's false, thats why they does't show, so it must we something with the way SF marks them as a match.

akegan:
Ok @software-project I'll reach out to Larry to ask about the salesforce eligibility thing. Do you happen to have the request we send to salesforce handy?

software-project:
@akegan endpoint: `/services/apexrest/ListingDetails` params: `{:householdsize=>2, :incomelevel=>72000.0, :childrenUnder6=>0}`

akegan:
added "blocked" label to this PR since it's blocked by salesforce work

akegan:
Larry updated salesforce code and this now looks ready to go as soon as tests pass!

akegan:
@software-project looks like rspec test is breaking on this

PR #1192, opened 2019-04-18
Ownership content changes #165075209

software-project:
https://www.pivotaltracker.com/story/show/165075209

### DIFF
--- a/app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
+++ b/app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
@@ -20,10 +20,7 @@ angular.module('dahlia.components')
       _.has(docSectionMap, @preference.preferenceName)
 
     @docSection = ->
-      if ListingIdentityService.isSale(@listing)
-        'homebuyers'
-      else
-        docSectionMap[@preference.preferenceName]
+      docSectionMap[@preference.preferenceName]
 
     return ctrl
   ]
--- a/app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
+++ b/app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
@@ -5,7 +5,7 @@ angular.module('dahlia.components')
     listing: '<'
   templateUrl: 'listings/components/listing/lottery-preference.html'
 
-  controller: ['ListingDataService', 'ListingIdentityService', (ListingDataService, ListingIdentityService) ->
+  controller: ['ListingDataService', (ListingDataService) ->
     ctrl = @
 
     docSectionMap = {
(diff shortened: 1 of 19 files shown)


app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
```diff
@@ -19,7 +20,10 @@ angular.module('dahlia.components')
       _.has(docSectionMap, @preference.preferenceName)
 
     @docSection = ->
-      docSectionMap[@preference.preferenceName]
+      if ListingIdentityService.isSale(@listing)
```
akegan: what's going on with this function? I'm not super clear on what the goal is

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -658,6 +658,10 @@ ShortFormApplicationController = (
   $scope.listingIsSale = ->
     ShortFormApplicationService.listingIsSale()
 
+  $scope.checklistSection = ->
```
akegan: same here, just wondering what the purpose of this function is.

app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
software-project: This is for anchors. The users for to checklist page, we want to direct them to related sections. For rental preferences or Homebuyers guide for sale listing. Same thing does the thing below, just the link is on the application form.

akegan: hm. if the listing is sale they might still be looking for information on a preference. would this work on that case?

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
software-project: It would should homebuyers guide. I don't think we are doing it now though and in the story it was said that there will be two different pages for sale and rental listing later on, so it's a temporary thing I believe. I can also modify it to work for both, default 'sale' and preference 'sale'

akegan: can we make it a bit clearer what the function is doing?

software-project:
@akegan Emily commented on the ticket that linking to homeownership in one place was put there by mistake and in the other it's not worth an effort so I deleted both functions.

app/assets/javascripts/listings/components/listing/lotteryPreference.js.coffee
```diff
@@ -2,9 +2,10 @@ angular.module('dahlia.components')
 .component 'lotteryPreference',
   bindings:
     preference: '<'
+    listing: '<'
   templateUrl: 'listings/components/listing/lottery-preference.html'
 
-  controller: ['ListingDataService', (ListingDataService) ->
+  controller: ['ListingDataService', 'ListingIdentityService', (ListingDataService, ListingIdentityService) ->
```
akegan: @software-project we can get rid of ListingIdentityService in here now too right?
```suggestion
  controller: ['ListingDataService', (ListingDataService) ->
```

akegan:
@software-project I ended up just making the change. Will merge once tests pass

PR #1193, opened 2019-04-19
Better password change validation and message #165075965

software-project:
https://www.pivotaltracker.com/story/show/165075965


akegan:
@software-project looks like we need to update messaging for password change when user is logged into account 
image

akegan:
Met with Emily to clarify AC for this PR and we need an updated design and a bit more work. Will fill in details in a bit

app/assets/javascripts/account/AccountController.js.coffee
```diff
@@ -272,6 +273,20 @@ AccountController = (
     else
       false
 
+  $scope.validatePasswordConfirmationMatch = (value) ->
```
akegan: @software-project are you saying that we can't check for password match if the first password is invalid because the value doesn't actually get set?

akegan:
@software-project I pushed visual updates after talking to emily. There might be another small tweak when we hear back from josh/another designer.

Outstanding items that need to be fixed:
1. We'd like validation to happen on loss-of-focus rather than just on submit. This is happening on e.g. create-account, but inputInvalid isn't getting called on account-settings until form submit for some reason.
1. I'd like to check in with you re: comment about ng-model not being saved until input is valid. Ideally password confirmation would only show an error if there's a mismatch, and would show an error on mismatch even if first password is invalid (which doesn't seem to be the case with current inputs)

akegan:
Found the source of the tabbing issue. It looks like the issue is that we have multiple forms on the account settings page, so currentForm is only getting set on submit of a given form https://github.com/Exygy/sf-dahlia-web/blob/master/app/assets/javascripts/account/AccountController.js.coffee#L41

app/assets/javascripts/account/AccountController.js.coffee
```diff
@@ -37,14 +37,14 @@ AccountController = (
...
   $scope.showSaleListings = $window.env.showSaleListings == 'true'
-  $scope.passwordConfirmation
 
   $scope.accountForm = ->
     # pick up which ever one is defined (the other will be undefined)
     $scope.form.signIn ||
     $scope.form.createAccount ||
     $scope.form.updatePassword ||
     $scope.form.passwordReset ||
+    $scope.form.accountPassword ||
```
akegan: @software-project this then messes with updating values in other forms on the account settings page. I tried it yesterday evening

app/assets/javascripts/account/templates/account-settings.html.slim
```diff
@@ -52,15 +52,15 @@
                   | {{'SIGN_IN.FORGOT_PASSWORD' | translate}}
               label.label-secondary for="auth_password"
                 | {{'LABEL.NEW_PASSWORD' | translate}}
-              input#auth_password ng-model="userAuth.user.password" name="password" placeholder="{{'CREATE_ACCOUNT.PASSWORD_SHORT_INSTRUCTIONS' | translate}}" type="password" ng-required="true" ng-class="{ error: inputInvalid('password'), 'margin-bottom': !inputInvalid('password') }" ng-pattern="passwordRegex" aria-describedby="auth_password_error"
+              input#auth_password ng-model="userAuth.user.password" name="password" placeholder="{{'CREATE_ACCOUNT.PASSWORD_SHORT_INSTRUCTIONS' | translate}}" type="password" ng-required="true" ng-class="{ error: inputInvalid('password'), 'margin-bottom': !inputInvalid('password') }" ng-pattern="passwordRegex" aria-describedby="auth_password_error" ng-model-options="{ allowInvalid: true }"
```
akegan: awesome! I'll update on other pages where we use password reset so the messaging improves there too

app/assets/javascripts/account/AccountController.js.coffee
software-project: @akegan I tested all other forms on account settings and they are working as expected. Reset password as well. It didnt work for me when I put that line as the last element. Maybe that was the case?

akegan: Looks like just need to remove this line and otherwise looks good!
```suggestion
```

PR #1194, opened 2019-04-19
Allow password reset link to be clicked more than one time #163441487

akegan:
https://www.pivotaltracker.com/story/show/163441487

If anyone in Eng wants to test reset token expiration, you can update config [here](https://github.com/Exygy/sf-dahlia-web/blob/master/config/initializers/devise.rb#L200)

I put this in the review instructions too, but we should test all of the different account creation/password update paths to make sure they still work


akegan:
I'm not super clear on the logic around this:
```
 def save_resource_for_edit
      # ensure that user is confirmed
      @resource.skip_confirmation! if user_is_confirmed
      # allow user to change password once without current_password
      @resource.allow_password_change = true
      @resource.save!
```

What was the goal of this logic? I want to make sure that I'm not removing any functionality. cc @chrisdolendo in particular

PR #1183, opened 2019-04-03
Explain 2018 income calculation rules #162877862

jrubenoff:
https://www.pivotaltracker.com/n/projects/1405352/stories/162877862

I utilized Foundation's existing tooltip component to avoid creating a new pattern.

Here's what it looks like on the listings page:

tooltip

@emilyshore: After my last comment in Pivotal, I realized that SMEs approved a version of the content that links to the PDF manual at the end. Ideally, we'd have a page on SFMOHCD.org that explains these income exceptions in a more readable way.

Would you like me to work on writing this page and getting that content approved by SMEs, or does this suffice for now?

PR #1195, opened 2019-04-24
Explain 2018 income calculation rules #162877862

jrubenoff:
This PR, unlike #1183, links to [a page on SFMOHCD.org](https://sfmohcd.org/special-calculations-household-income) instead of the inclusionary manual. Cissy and Maria have approved the page's content.

Here's what this looks like on the listings page:

Screen Shot 2019-04-23 at 5 39 52 PM

cc: @emilyshore


akegan:
looks good! only comment is that we should make the links to mohcd target=blank so they open in a new page so user doesn't lose their place in income calculation

jrubenoff:
@akegan: Alrighty, fixed that.

Also fixed a `ul` tag on the wrong line that I just noticed.

PR #1196, opened 2019-04-24
[Hotfix] Change page after autofill based on type of listing #165574739

akegan:
https://www.pivotaltracker.com/story/show/165574739

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
@@ -75,8 +75,16 @@ ShortFormApplicationController = (
 
   $scope.emailRegex = SharedService.emailRegex
 
-  $scope.trackAutofill = ->
+  $scope.startAutofilledApp = ->
     AnalyticsService.trackFormSuccess('Application', 'Start with these details')
+    # Start page after autofill preview
+    $scope.go($scope.getFirstFillableAppPage())
+
+  $scope.getFirstFillableAppPage = ->
+    if $scope.listingIsSale()
+      'dahlia.short-form-application.prerequisites'
+    else
+      'dahlia.short-form-application.name'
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/short-form/templates/b0-autofill-preview.html.slim
```diff
@@ -9,7 +9,7 @@
 
   .button-pager
     .button-pager_row.primary
-      button.button.primary ui-sref="dahlia.short-form-application.name" ng-click="trackAutofill()"
+      button.button.primary ng-click="trackAutofill()"
```
chrisdolendo: This should be `startAutofilledApp` yeah?

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -75,8 +75,16 @@ ShortFormApplicationController = (
 
   $scope.emailRegex = SharedService.emailRegex
 
-  $scope.trackAutofill = ->
+  $scope.startAutofilledApp = ->
     AnalyticsService.trackFormSuccess('Application', 'Start with these details')
+    # Start page after autofill preview
+    $scope.go($scope.getFirstFillableAppPage())
+
+  $scope.getFirstFillableAppPage = ->
```
chrisdolendo: I noticed that this function already exists in `ShortFormNavigationService` which is already injected in this controller and you can just call to return the right statename...

```
  Service.initialState = () ->
    if ListingIdentityService.isSale(ShortFormApplicationService.listing)
      'dahlia.short-form-application.prerequisites'
    else
      'dahlia.short-form-application.name'
```

akegan: fixed!

akegan:
it's calling web-prod to check for semaphore failure. web-full is passing so I'm going to merge

PR #1198, opened 2019-04-25
[#161892754] Upgrade Node to 10.15.3 (LTS)

software-project:
https://www.pivotaltracker.com/story/show/161892754


akegan:
@software-project one other thing to do on this PR, since there are a lot of dependencies that change around in yarn.lock, it might be a good idea to see if there's some sort of migration/upgrade guide that might detail breaking changes that our tests don't catch. Would you be up for looking for that guide and seeing if there's anything we should worry about?

software-project:
@akegan sure, I'll have a look

software-project:
@akegan TL;DR; if tests are passing we should be good.
I had a look on many pages regarding upgrading node. Two that make the most sense:
https://www.joyent.com/blog/upgrading-nodejs-to-8
https://auth0.com/blog/nodejs-10-new-changes-deprecations
Long story short we are not using node js that much. Async/ await in tests should be more stable, which should be really good for us. It seems like frontend should be affected at all, as rails views are generated on the rails side with it's own coffeescript version.
Most libraries that were changed in the lock file are related to grunt, but there no braking changes and actually yarn.lock didn't reflect versions that were set in the package.json a while back and this commit just fixes that.

akegan:
ok will test locally then pass!

PR #1205, opened 2019-05-09
 Upgrade lodash and devise #164739712

software-project:
https://www.pivotaltracker.com/story/show/164739712


app/assets/javascripts/shared/SharedService.js.coffee
```diff
@@ -27,7 +27,7 @@ SharedService = ($http, $state, $window, $document) ->
   Service.getLanguageCode = (langName) ->
     # will take "English" and return "en", for example
     if langName
-      _.invert(Service.languageMap)[_.startCase(langName)]
+      _.invert(Service.languageMap)[_.startCase(_.lowerCase(langName))]
```
akegan: @software-project why did this update need to be made? seems weird to lower case then start case?

software-project: It seems like it used to work differently with the past version, as tests were failing for upper case scenarios. That fixed it.

software-project:
@akegan if I remember correctly we were pairing on it, if I remember correctly with Catherine, and images were not working. Thats why we decided to update toolkit and it fixed the styling/ asset issue.

app/assets/javascripts/shared/SharedService.js.coffee
akegan: oh I see. There was a change to startCase to just capitalize the first char rather than changing to capitalization case. 

e.g. 
used to do EN -> En, now just does EN -> EN.

Maybe we could replace with [Capitalize](https://lodash.com/docs/4.17.11#capitalize) Instead and see if the tests pass? Seems to be what we're interested in

software-project: Yes, exactly. That should also work, will give it a try.

app/assets/javascripts/shared/SharedService.js.coffee
```diff
@@ -27,7 +27,7 @@ SharedService = ($http, $state, $window, $document) ->
   Service.getLanguageCode = (langName) ->
     # will take "English" and return "en", for example
     if langName
-      _.invert(Service.languageMap)[_.startCase(_.lowerCase(langName))]
+      _.invert(Service.languageMap)[_.startCase(_.capitalize(langName))]
```
akegan: @software-project if you switch to capitalize then you shouldn't need the _.startCase

app/assets/javascripts/shared/SharedService.js.coffee
```diff
@@ -27,7 +27,7 @@ SharedService = ($http, $state, $window, $document) ->
   Service.getLanguageCode = (langName) ->
     # will take "English" and return "en", for example
     if langName
-      _.invert(Service.languageMap)[_.startCase(langName)]
+      _.invert(Service.languageMap)[_.startCase(_.capitalize(langName))]
```
akegan:
```suggestion
      _.invert(Service.languageMap)[_.capitalize(langName)]
```

Should just need capitalize and not start case in this case, but we'll see how it does on the tests

PR #1209, opened 2019-05-23
Upgrade Jasmine and fix failing tests #165609939

callaghanc:
https://www.pivotaltracker.com/story/show/165609939

In Jasmine 3, it is now the default to run tests in random order. Previously, we were not running tests in random order. Running them randomly has exposed many tests where the test setup was not isolated enough and so the test failed if certain other tests were run before. In this story, we'll be fixing all those newly failing tests after the Jasmine upgrade.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,7 +117,7 @@ group :development, :test do
   gem 'awesome_print'
   gem 'dotenv-rails'
   gem 'jquery-rails' # only needed for jasmine-jquery
-  gem 'jasmine', github: 'pivotal/jasmine-gem'
+  gem 'jasmine'
   gem 'jasmine-jquery-rails' # used for functions like `getJSONFixture`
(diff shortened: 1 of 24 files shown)


Gemfile
```diff
@@ -117,7 +117,7 @@ group :development, :test do
   gem 'awesome_print'
   gem 'dotenv-rails'
   gem 'jquery-rails' # only needed for jasmine-jquery
-  gem 'jasmine', github: 'pivotal/jasmine-gem'
+  gem 'jasmine'
```
callaghanc: @akegan jasmine is developed by github. back in the day they had the jasmine gem at https://github.com/pivotal/jasmine-gem, but it's since moved to its own proper rubygems gem at https://rubygems.org/gems/jasmine/, hence this change to draw the gem from rubygems instead.

PR #1211, opened 2019-06-05
Separate caching of rental and sale listing calls #165862948

chrisdolendo:
https://www.pivotaltracker.com/story/show/165862948

**Approach** 

The goal of the story is to reduce the size of the cache for Listings. Prior to this PR, we were calling for and storing all the listings (both rental and sale) under one key `/ListingDetails`, _and then_ filtering all the data by Tenure type (rental or sale) to return to the client. This PR separates the caching (and call) of rental and sale listing calls into their own respective keys.

Please see code comment for more details.


app/services/force/listing_service.rb
```diff
@@ -32,7 +32,8 @@ class ListingService
     # `ids` is a comma-separated list of ids
     # returns cached and cleaned listings
     def self.listings(attrs = {})
-      params = attrs[:ids].present? ? { ids: attrs[:ids] } : nil
+      params = { Tenure: attrs[:Tenure] }
+      params[:ids] = { ids: attrs[:ids] } if attrs[:ids].present?
       results = get_listings(params)
```
chrisdolendo: `get_listings` calls `cached_get` which is contains this code:

```
      @cache.fetch(key, force: force_refresh, expires_in: expires_in) do
        get(endpoint, params)
      end
```

It asks Rails cache for the corresponding key, which is defined as:

```
      key = "#{endpoint}#{params ? '?' + params.to_query : ''}"
```

The code change above allows the `Tenure` attrs to be part of the key params so that the cache can return data saved to the key `/ListingDetails?Tenure=sale` and `/ListingDetails?Tenure=rental`.

In line 45, `ListingService#raw_listings`, is the method that's called in the caching rake task, so I updated that to cache for rental listings and sale listings _separately_.

If you access the rails console for this PR review app and run `Rails.cache.fetch('/ListingDetails?Tenure=sale')` and `Rails.cache.fetch('/ListingDetails?Tenure=rental')`, you'll see the cached data saved to each key respectively :)

To access the rails console for this review app, you can locally run `heroku run rails c --app dahlia-full-pr-1211`.

app/services/force/listing_service.rb
```diff
@@ -41,7 +42,10 @@ def self.listings(attrs = {})
 
     def self.raw_listings(opts = {})
       force = opts[:refresh_cache] || false
-      Request.new(parse_response: true).cached_get('/ListingDetails', nil, force)
+      Request.new(parse_response: true)
+             .cached_get('/ListingDetails', { Tenure: 'rental' }, force)
+      Request.new(parse_response: true)
+             .cached_get('/ListingDetails', { Tenure: 'sale' }, force)
```
chrisdolendo: Getting a red line here from code climate, but not 100% sure how test write a test for this function as it's only used by the rake task 🤔

akegan:
@chrisdolendo looks like test coverage decreased a bit. any way to add a bit more testing.

Also curious why the listings.yml file was 19k lines and was replaced by 2 files with only 102 lines.

chrisdolendo:
Tests finally pass. We added 4 additional tests and it still didn't budge the total-coverage percentage which is very weird. Code climate chrome plugin also doesn't indicate where the untested sections are.

lib/tasks/cache.rake
```diff
@@ -4,11 +4,19 @@ namespace :cache do
     Rails.cache.clear
   end
 
-  desc 'Pre-fetch all listings for caching'
+  desc 'Pre-fetch all listings, rental listings, and ownerhsip listings, for caching'
```
akegan:
```suggestion
  desc 'Pre-fetch all listings, rental listings, and ownership listings, for caching'
```

akegan:
@chrisdolendo @software-project  I logged into console and caches were nil, then I tried running `rake cache:prefetch` and got this error

```
rake aborted!
NameError: uninitialized constant Request
/app/lib/tasks/cache.rake:9:in `block (2 levels) in <top (required)>'
Tasks: TOP => cache:prefetch
(See full trace by running task with --trace)
```

not sure if I did something wrong or if it's just not working. I accessed by running `heroku run bash` then `rake cache:prefetch`

chrisdolendo:
@akegan Sorry about that typo - you can run again, just re-tested and everything works. You can follow the review instructions on the story :+1:

chrisdolendo:
actually @software-project looks like @akegan andrea asked you to merge if you confirm QA :) So can you review and test?

PR #1214, opened 2019-06-06
Add year to past listings #164999194

software-project:
https://www.pivotaltracker.com/story/show/164999194


app/assets/javascripts/customFilters.js.coffee
```diff
@@ -10,6 +10,15 @@ angular.module('customFilters', [])
     suffix = if relevantDigits <= 3 then suffixes[relevantDigits] else suffixes[0]
     dtfilter + suffix
   ]
+.filter 'dateSuffixOrShort', ['$filter', ($filter) ->
```
akegan: is there a better name we could have for this that describes what it's doing?

akegan: might also be helpful to add some comments to explain the business rules around this, as per @callaghanc's suggestion recently

software-project: I've added a comment and slightly improved the name. I dont have a better idea, but I'm open for suggestions.

akegan:
I'll take this one over and add a test so we can get it merged

akegan:
@chrisdolendo are you up for reviewing the tests that I added to this PR?

PR #1222, opened 2019-06-12
COP/DTHP Update - Content only #165472486

akegan:
This is an alternative implementation that doesn't use radio buttons and just updates the content to make it less confusing


chrisdolendo:
@emilyshore approved the content changes -- will merge

PR #1223, opened 2019-06-24
Translations without master

akegan:
Translations updates without pulling in latest master (so we have consistent experience for asian inc testing


akegan:
no longer needed

PR #1227, opened 2019-06-26
Update Pricing Table styles #165887950

akegan:
# Open questions/ issues

## Pricing hero - Rental
Reserved units doesn't quite work (using `table.table-pricing.table-pricing-hero.table-reserved`).

Will follow-up with fix in future ticket [#167024851](https://www.pivotaltracker.com/story/show/167024851)
image

## Pricing hero - Sale
For HOA dues, do we want to say per month? if yes, do we add "or more" to it too? (Initial mock ups just have "or more" if there's a range of prices, otherwise no subtitle)
image


emilyshore:
@akegan In the Chinesetrans Rent veterans listing the horizontal orange bar looks odd in mobile.

LoganArnett:
Nothing specific code wise that I can see as an issue, looks clean and I am able to run through it locally

akegan:
ok @LoganArnett is the PR good to go then? @emilyshore did QA but it might be helpful if you could do a bit of QA too

akegan:
I replied to Emily in slack, but the tables for reserved units in listing directory is a known issue that won't be fixed in this PR

PR #1231, opened 2019-07-10
Include checks for any senior buildings vs all senior building for household members #162426017

chrisdolendo:
https://www.pivotaltracker.com/story/show/162426017

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
@@ -526,7 +526,7 @@ ShortFormApplicationController = (
   ###### Household Section ########
   $scope.addHouseholdMember = ->
     noAddress = $scope.householdMember.hasSameAddressAsApplicant == 'Yes'
-    if $scope.applicantDoesNotMeetSeniorRequirements('householdMember')
+    if $scope.applicantDoesNotmeetAllSeniorBuildingRequirements('householdMember')
       age = { minAge: $scope.listing.Reserved_community_minimum_age }
       $scope.eligibilityErrors = [$translate.instant('ERROR.SENIOR_EVERYONE', age)]
       $scope.handleErrorState()
@@ -557,10 +557,18 @@ ShortFormApplicationController = (
   $scope.validateHouseholdEligibility = (match) ->
     $scope.clearEligibilityErrors()
     form = $scope.form.applicationForm
+
+    if ShortFormApplicationService.householdDoesNotMeetAtLeastOneSeniorRequirement()
+      age = { minAge: $scope.listing.Reserved_community_minimum_age }
+      $scope.eligibilityErrors = [$translate.instant('ERROR.SENIOR_ANYONE', age)]
+      $scope.handleErrorState()
+      return
+
     # skip the check if we're doing an incomeMatch and the applicant has vouchers
     if match == 'incomeMatch' && $scope.application.householdVouchersSubsidies == 'Yes'
       ShortFormNavigationService.goToSection('Preferences')
       return
+
     ShortFormApplicationService.checkHouseholdEligibility($scope.listing)
       .then( (response) ->
         eligibility = response.data
@@ -824,7 +832,7 @@ ShortFormApplicationController = (
 
   $scope.afterSignInWhileApplying = ->
     if (ShortFormApplicationService
-        .applicantDoesNotMeetSeniorRequirements(AccountService.loggedInUser)
+        .applicantDoesNotmeetAllSeniorBuildingRequirements(AccountService.loggedInUser)
     )
       # log user out to either create account or continue anonymously
       AccountService.signOut({ preserveAppData: true })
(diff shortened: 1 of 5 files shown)


akegan:
@chrisdolendo I just did another test with the marcy adelman listing https://dahlia-full.herokuapp.com/listings/a0W0P00000HNfNTUA1/apply/household-members, and the message is wrong but the behavior is correct right now as is. If you have at least one HH member who is old enough, it validates on the "done adding HH members" and lets you through. If you don't have anyone who is over the age, it gives a warning and has the wrong message. 

I'm confused about what the changes are in this PR and what the issue was. Can we hop on a call for you to explain when you have a sec?

akegan:
@chrisdolendo reviewing now, but seems like the two e2e errors are definitely related to your changes

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -578,15 +586,14 @@ ShortFormApplicationController = (
...
-    if eligibility.householdMatch && !seniorReqError
+    if eligibility.householdMatch
       # determine next page of household section
       if ShortFormApplicationService.hasHouseholdPublicHousingQuestion()
         ShortFormNavigationService.goToApplicationPage('dahlia.short-form-application.household-public-housing')
       else
         $scope.checkIfReservedUnits()
     else
-      $scope._determineHouseholdEligibilityErrors(error, seniorReqError)
+      $scope._determineHouseholdEligibilityErrors(error)
```
akegan: I think with all of these householdEligibilityErrors functions it'd be great to clarify which household eligibility errors? 

I'm confused about the check if reserved units check above, for example.

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -557,10 +557,18 @@ ShortFormApplicationController = (
...
+      $scope.eligibilityErrors = [$translate.instant('ERROR.SENIOR_ANYONE', age)]
+      $scope.handleErrorState()
+      return
+
     # skip the check if we're doing an incomeMatch and the applicant has vouchers
     if match == 'incomeMatch' && $scope.application.householdVouchersSubsidies == 'Yes'
       ShortFormNavigationService.goToSection('Preferences')
       return
+
     ShortFormApplicationService.checkHouseholdEligibility($scope.listing)
```
akegan: also confused what this function does because it checks income and household eligibility?

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
```diff
@@ -977,24 +977,26 @@ ShortFormApplicationService = (
...
-    reservedType = listing.Reserved_community_type || ''
+    return false unless !!requirement.match(/entire household/i)
+
     if _.isString(member)
       # are we evaluating a form value
       age = Service.memberAgeOnForm(member)
     else
       # or evaluating an appMember object
       age = Service.memberAge(member)
-    !!reservedType.match(/senior/i) && !!requirement.match(/entire household/i) &&
```
akegan: do we still need this check for senior (vs e.g. veteran or developmental disability?)

akegan:
OK @chrisdolendo I just tested locally to see what was going on with the failing tests and I couldn't reproduce. I think it looks as expected. 

image

on semaphore it's the last line of this snippet that's failing. 
```
Scenario: Signing in on the welcome back page when my account DOB disqualifies me and creating a new account
        Given I go to the first page of the "Senior Test Listing" application
        When I answer "No" to the community screening question
        Then I should see a form notice that says "Everyone in your household must be a Senior"

        When I wait "1" seconds
        And I answer "Yes" to the community screening question
        # Continue past welcome page
        And I hit the Next button "1" time
        # Email for existing account, but a different birth date than in account settings to qualify for senior listing
        And I fill out the Name page as "Alice Walker" with birth date "1/1/1900"
        And I sign in as "Alice Walker" with my email pre-filled
        Then I should see a form notice that says "Everyone in your household must be a Senior"
```

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
chrisdolendo: sure I noticed that too, will clarify the function a bit more because there are different HH elig checks

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
chrisdolendo: Actually yes this is a good point, it could be veteran for entire household or at least one - forgot that that was a possibility. Will update.

chrisdolendo:
It does, so I'm not sure why it's failing, I will see if the description matches the actual check.

app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
chrisdolendo: Yes basically it refers to the callback after the income page:

```
    'income':
      scopedCallbacks: [{
        func: 'validateHouseholdEligibility'
        param: 'incomeMatch'
      }]
```

When the applicant puts in the income, we ask Salesforce if the householdsize-to-income ratio is valid for the listing so we call `validateHouseholdEligibility` again. The param 'incomeMatch' here tells the function that it's coming from the income page... BUT if the user said "Yes" on having a voucher subsidy then we skip it the income check.

chrisdolendo:
Tests pass yipee!

PR #1233, opened 2019-07-16
Include latest Chinese translations from Nicole #166910180

akegan:
https://www.pivotaltracker.com/story/show/166910180


chrisdolendo:
@akegan I'm seeing a few conflicts with the Chinese translations on here, and it seems like it actually makes sense to overwrite master with this head.

akegan:
@chrisdolendo ready to review again

PR #1236, opened 2019-07-17
Remove background color from matched listings #167360135

akegan:
https://www.pivotaltracker.com/story/show/167360135


chrisdolendo:
Looks good to me!! 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1239, opened 2019-07-19
Update github vulnerabilities #166825104

software-project:
https://www.pivotaltracker.com/story/show/166825104


akegan:
@software-project there's a github alert for eslint but I can't find it in the code, can you?

akegan:
@software-project I confirmed that we are OK with grunt-angular-translate because it's a dev dependency and is not installed on production

PR #1241, opened 2019-07-19
[Hotfix] Use browse listings fields for individual listing cache #167409242

akegan:
https://www.pivotaltracker.com/story/show/167409242


app/services/force/listing_service.rb
```diff
@@ -104,8 +99,9 @@ def self.array_sort!(listing)
       end
     end
 
-    private_class_method def self.get_listings(params = nil)
-      results = Request.new(parse_response: true).cached_get('/ListingDetails', params)
+    private_class_method def self.get_listings(params = nil, force = false)
```
akegan: very open to feedback on how to handle passing "force" down to cache listings, I'm not super clear on what best practices are for ruby. also open to different naming of the var (e.g. force_recache/refresh instead of just force)

callaghanc: @akegan what you have here is fine. let's get into other options in https://www.pivotaltracker.com/story/show/167411060, since that is more focused on changing the get listings path

callaghanc:
@akegan i tested this locally and confirmed that the cache prefetch tasks are now calling the listings endpoint with subset set to browse, instead of fetching all the listings with the full fieldset. i added a few small changes to the CacheService wording, take a look and edit wording/naming as you see fit, then feel free to merge if tests have passed.

PR #1242, opened 2019-07-22
Make subset: 'browse' default for ListingService #167411060

software-project:
https://www.pivotaltracker.com/story/show/167411060


akegan:
@callaghanc your changes look good to me! just did a quick test to see that only browse fields show up when I got to `/api/v1/listings.json` I should probably do some caching testing though right?

callaghanc:
@akegan You can test that caching still works as expected, you can also test that we are using the browse subset when fetching listings for the eligibility checker and on the My Favorites page

PR #1246, opened 2019-07-25
Add text to Chinese translation for 'Here's what to expect' #143618787

chrisdolendo:
https://www.pivotaltracker.com/story/show/143618787


app/assets/json/translations/locale-zh.json
```diff
@@ -24,7 +24,7 @@
         "P2": "<a ui-sref='{{sref}}' target='_blank'>請查閱我們的檢查表</a>，以了解有關所需文件和如何宣稱擁有優先權利的資訊。",
         "P3": "您也可以現在開始填寫申請表並儲存草稿，日後再上傳文檔。",
         "P4": "請注意, 每個家庭成員的名字只限在房屋上市名單的申請表格出現一次. 如果在同一房屋上市名單上您的申請表格包含有其他家庭成員已遞交申請表格, 則所有申請資格都將會被取消.",
-        "TITLE": "針對此申請，您需要完成以下作業。",
```
akegan: @chrisdolendo i think we settled on adding it as a new string so it's clearer to the translators what's going on

PR #1249, opened 2019-08-02
Do not translate numerical values for AMI charts #166808219

chrisdolendo:


chrisdolendo:
I'll look into it now!

akegan:
@chrisdolendo just figured out what it is. Some listings have max income /month, and /year, and some with multi ami have the different % AMIs as the columns. They are 2 separate components that we pull in. I think you updated the single AMI table, and not the multi-ami table

akegan:
approved! thanks @chrisdolendo !

PR #1252, opened 2019-08-12
 Upgrade to factorybot #166936249

chrisdolendo:
https://www.pivotaltracker.com/story/show/166936249


spec/support/jasmine.rb
```diff
@@ -4,7 +4,7 @@
   # this resets factory girl sequences so that the jasmine
```
akegan:
```suggestion
  # this resets factory bot sequences so that the jasmine
```

PR #1256, opened 2019-08-24
Set locales for devise emails #167773665

akegan:
https://www.pivotaltracker.com/story/show/167773665

This PR:
- Adds locale as a param to all front-end requests that lead to email creation
- Adds jasmine tests to verify that that locale is being added to those params
- Adds switch_locale wrapper to all API requests, that sets the I18n locale setting for nested requests based on the provided param using with_locale
- Adds tests for update_account mailer
- Moves content previously being picked up by default devise strings to our own locale yml files for consistency across languages
  - Previously, we were getting the spanish translations for some of the strings for free, but there were no translations available for chinese and filipino
- Updates emailer_preview to be able to easily set the locale and preview emails in different languages. 
  - Created [folllow-up ticket](https://www.pivotaltracker.com/story/show/168210747) to later be able to set locale via param in email previews

Still left to do?
- [ ] Consolidate params translate.use() usage?
- [ ] Testing scheme for switch_locale


akegan:
@callaghanc heads up, I got this working but on monday I'm going to be looking on your input on the two items in the description above. ( Consolidate params translate.use() usage and 
  Testing scheme)

akegan:
@callaghanc I think I've reached my limit of what I can figure out how to test on this PR. I would still like to test the with_locale somehow, but I'm not sure how. We don't have any actions on the base api_controller to test against, and we don't have any tests at all on the account_controller and the short_form_controller has no tests at all on submit application. 

Let me know if you can pair on this tomorrow morning and maybe add another test or just review this PR and get it merged. 

I've also updated the description to give a better idea of the changes that I've made

akegan:
@callaghanc I added a test for account requests to check if with_locale is called

app/assets/javascripts/account/AccountService.js.coffee
```diff
@@ -234,14 +237,14 @@ AccountService = (
 
   Service.userDataForSalesforce = ->
     contact = Service.userDataForContact()
-    contact.DOB = ShortFormDataService.formatUserDOB(contact)
-    contact = ShortFormDataService.removeDOBFields(contact)
-    contact
+    contactWithDOB = _.merge({}, contact, {'DOB': ShortFormDataService.formatUserDOB(contact)})
+    ShortFormDataService.removeDOBFields(contactWithDOB)
 
   Service._createAccountParams = ->
```
callaghanc: @akegan while we're in here could we remove the underscore from in front of this function name? so just `createAccountParams` instead of `_createAccountParams`

akegan:
@callaghanc change made!

PR #1257, opened 2019-08-30
[Hotfix] Add open house addresses to listing #168224482

akegan:
https://www.pivotaltracker.com/story/show/168224482


exygy-dev:
Code looks good to me and everything looks to be rendering well in test environments for mira. My only quibbles would be about labelling the location or adding a tiny more whitespace padding between the address and the hours above, but neither are a big deal.

akegan:
thanks @bk3c I think we'll have to revisit with a designer to find a longer-term solution so we don't repeat addresses over and over again.

PR #1264, opened 2019-09-19
Moving untranslated strings to locale #167847832

software-project:
https://www.pivotaltracker.com/n/projects/1405352/stories/167847832


app/assets/javascripts/listings/components/browse-listings.html.slim
```diff
@@ -9,17 +9,17 @@ header.lead-header ng-if="$ctrl.parent.hasEligibilityFilters()"
...
       .results-header
         p.results-header_values
-          | For
+          | {{'t.for' | translate}}
           strong
             '  {{$ctrl.eligibilityFilters.household_size}}
           ' {{$ctrl.eligibilityFilters.household_size == 1 ? 'person ' : 'people '}}
           span ng-if="$ctrl.eligibilityFilters.children_under_6 && $ctrl.eligibilityFilters.children_under_6 > 0"
-            ' (including
+            ' ({{'t.including' | translate}}
```
akegan: This part might be a bit complicated because the order of words and phrases don't always match up in different languages. I'll take a look at this bit in particular and ask for advice from a chines-language speaker

app/assets/javascripts/listings/components/listing/features-section.html.slim
```diff
@@ -72,26 +71,22 @@ accordion-heading
...
+          span.t-small translate="listings.features.deposit"
           p.t-delta.no-margin.t-bold
             | {{::$ctrl.parent.listing.Deposit_Min | currency:"$":0}}
             span ng-if="$ctrl.parent.listing.Deposit_Max && $ctrl.parent.listing.Deposit_Max != $ctrl.parent.listing.Deposit_Min"
               | &ndash;{{ $ctrl.parent.listing.Deposit_Max | currency:"$":0 }}
-          span.content-label.t-small
-            | or one month's rent
-          span.t-tiny.d-block ng-if="$ctrl.listingIsBMR()"
-            | May be higher for lower credit scores
+          span.content-label.t-small translate="listings.features.or_one_months_rent"
```
akegan: this bit might be tough too

app/assets/javascripts/listings/templates/listing/_income-table.html.slim
```diff
@@ -2,9 +2,9 @@
   table role="grid"
     thead
       tr
-        th scope="col"  household size
-        th scope="col"  maximum income/month
-        th scope="col"  maximum income/year
+        th scope="col"  {{'t.household_size' | translate}}
+        th scope="col"  {{'t.maximum' | translate}} {{'t.income_month' | translate}}
```
akegan: these two probably need to be combined into one string, again because of word order differences between languages.

app/assets/javascripts/listings/components/browse-listings.html.slim
```diff
@@ -63,7 +63,7 @@ section.results-section#listing-results ng-if="!$ctrl.parent.openListings.length
           p.margin-bottom--half
             | {{ $ctrl.noMatchesLabel() }}
           p.t-small.c-steel.no-margin
-            | <a class="lined" ui-sref="dahlia.additional-resources">Click here</a> for other rental and ownership affordable housing opportunities.
+            | <a class="lined" ui-sref="dahlia.additional-resources">{{'label.click_here' | translate}}</a> {{'listings.for_other_lisitings' | translate}}
```
akegan: This should be combined into 1 string with the html embedded if possible, to help with word order or phrasing differences. We have a couple of examples of including links in the locale-en.json file if you want to check it out. You can pass the ui-sref in as a variable if it makes sense.

app/assets/javascripts/listings/components/listing/features-section.html.slim
software-project: Do you want the hole thing including deposit in a translation?

app/assets/javascripts/listings/components/browseListings.js.coffee
```diff
@@ -28,5 +29,19 @@ angular.module('dahlia.components')
       else if @tenureType == 'rental'
         $translate.instant('listings.you_dont_match_any_rent')
 
+    @resultsHeader = () ->
```
akegan: I'm having a hard time parsing all of these different pieces of logic. I think we still want to combine it into one big sentence for all of the different potential outputs. 

Can you list out what all of the different final possibilities are?

software-project: Sure, there are 12 different options here. There are always one or more household members with income calculated per month or year. That are 4 options. Of each of those we can have no child, one or multiple.

app/assets/javascripts/listings/components/listing/features-section.html.slim
akegan: This is probably fine as is. We'll see if we hear any issues from the translators

app/assets/json/translations/locale-en.json
```diff
@@ -1154,7 +1165,8 @@
             "leave_your_application": "Leave your application?",
             "listing_updated": "Listing Updated",
             "lottery_results_posted": "Lottery Results Posted",
-            "maximum": "maximum",
+            "maximum_income_month": "maximum income/month",
```
akegan: Let's actually update these to use "per" instead of "/". That's easier to translate ("/"'s aren't super common in chinese, and then we have the added benefit of it making more sense for screenreaders

app/assets/json/translations/locale-en.json
```diff
@@ -763,6 +771,7 @@
                 "visit_a_housing_counselor": "Visit a local housing counselor for help with your application.",
                 "visit_homeownership_sf": "Visit HomeownershipSF"
             },
+            "at_total_income": "at <strong>{{income}}</strong>/{{per}}",
```
akegan: same here, let's ditch the "/" while we're at it and have the per be "per month" or "per year" then we can reuse the "per_month" and "per_year" keys we already have too.

app/assets/javascripts/listings/components/browseListings.js.coffee
akegan: OK I asked a couple of folks who speak chinese here in the office and they think that it can be OK as you have it (pending update to move from "/month" to "per month").

I'll work with them directly to make sure that the translations end up making sense in the end.

app/assets/json/translations/locale-en.json
```diff
@@ -739,29 +771,76 @@
                 "visit_a_housing_counselor": "Visit a local housing counselor for help with your application.",
                 "visit_homeownership_sf": "Visit HomeownershipSF"
             },
+            "at_total_income": "at <strong>{{income}}</strong> per {{per}}",
```
akegan: @software-project I actually think you should include the per in the variable, and use the per_month and per_year strings that we already have translated as the timeframe variable instead of just year vs month

akegan:
OK. I think this is almost good to go, just have one more comment on the per_month piece. . My only other concern is whether or not we should add a bit more structure to the way we store the content (add more categories basically). 

I also want to confirm @software-project that you checked as you were going through to see if there were strings that were reusable from locale-en already to avoid duplicating content to translate if possible.

akegan:
Just started some QA, posting some screenshots here:

Rental directory page:
image

Sale directory page:
image

Sale listing page, the pipes ("|") are new :
image

software-project:
@akegan missed that, sorry. Updated.

akegan:
@software-project  I found a few more QA issues, this time doing side-by-side comparison against master (should be the final check!). I added checkboxes next to the pieces to make it easier to track

Full is on the left, PR changes are on the right.

# Sale listing details
- [ ] What to expect content on the sale listing is off:
image

- [ ] Sales pricing table missing colons between sales price and how dues and the price:
image

- [ ] Unit features table missing space between number and "square feet"
image

# Rental directory page
- [ ] Pricing table has "income" instead of "per month" in minimum income column
image

# Rental listing details
- [ ] Same issue on what to expect, although I think the content is supposed to be slightly different for rental vs sale
image

app/assets/javascripts/listings/templates/property-card/_rental-stats-row.html.slim
```diff
@@ -1,40 +1,40 @@
 tr ng-repeat-start="unit_summary in ::summary" ng-if="::unit_summary.minRentalMinIncome || unit_summary.maxRentalMinIncome"
   / Dollar based rent and min income
-  td.is-subtitled data-th="Unit Type"
+  td.is-subtitled data-th="{{'t.unit_type' | translate}}"
     | {{::unit_summary.unitType}}
-  td.is-subtitled data-th="Min. Income"
+  td.is-subtitled data-th="{{'t.min_income' | translate}}"
```
akegan: might be nice to update this variable name to make it clear that it's the shortened mobile view title?

e.g. min_income_short?

akegan:
One more piece  as I check mobile view too:

# Sale listing details page

- [ ] Mobile title says prices instead of price
image

software-project:
@akegan I addressed issues you found. Good job finding them. Thank you.

app/assets/javascripts/listings/components/browseListings.js.coffee
```diff
@@ -37,7 +37,7 @@ angular.module('dahlia.components')
         children = $translate.instant(if @eligibilityFilters.children_under_6 == '1' then 't.child' else 't.children')
         header.push($translate.instant('listings.including_children', {number: @eligibilityFilters.children_under_6, children: children}))
 
-      timeframe = $translate.instant('t.' + $filter('incomeTimeframe')(@eligibilityFilters.income_timeframe))
+      timeframe = $translate.instant('t.per_' + $filter('incomeTimeframe')(@eligibilityFilters.income_timeframe))
```
akegan: This one is going to be hard for the grunt translations script to pick up and detect. It's slightly less clean but it'd be easier to maintain if you explicitly say `$translate.instant('t.per_month')` or `$translate.instant('t.per_year')`

akegan:
did another quick QA and things look good, just want that last per_month/per_year update, and then make sure you run grunt translations to get everything sorted and make sure that all of the keys are recognized. 

Let me know if you have any questions on changes after running grunt translations

PR #1266, opened 2019-09-23
Fix lottery results caching #167523476

software-project:
https://www.pivotaltracker.com/story/show/167523476


spec/services/listing_service_spec.rb
```diff
@@ -103,4 +103,13 @@
       end
     end
   end
+
+  describe '.lottery_buckets' do
+    it 'should call right enpoint with forced recache' do
```
akegan:
```suggestion
    it 'should call right endpoint with forced recache' do
```

akegan:
@software-project found the issue, CACHE_SALESFORCE_REQUESTS isn't set on semaphore and I don't think we want it to be set? / we don't have the capability of running memcached on semaphore. 

I'm going to look into it locally and see if I can figure out a plan that makes sense!

PR #1275, opened 2019-10-25
Include directional prefix in cleaned street address #169187139

akegan:
Review instructions on [the ticket](https://www.pivotaltracker.com/story/show/169187139)


akegan:
set back to WIP because I need to fix and add tests

akegan:
@LoganArnett tests should be passing now so this is ready for review

PR #1306, opened 2020-01-18
Property card a11y fixes #169413027

cpalmatier:
Fix https://www.pivotaltracker.com/story/show/169413027 via changing all property card h1 elements to h2, because a11y screen reader behavior.

### DIFF
--- a/app/assets/javascripts/listings/components/property-card.html.slim
+++ b/app/assets/javascripts/listings/components/property-card.html.slim
@@ -1,5 +1,5 @@
 .property-card ng-class="{last: $last, first: $first}"
-  h2.property-card_title.text-center.hide-for-medium-up.skiptranslate
+  h3.property-card_title.text-center.hide-for-medium-up.skiptranslate
     a ui-sref="dahlia.listing({id: $ctrl.listing.Id})" {{::$ctrl.listing.Name}}
   p.property-card_address.text-center.hide-for-medium-up.skiptranslate
     | {{::$ctrl.listingContainer.formattedBuildingAddress($ctrl.listing, 'street')}}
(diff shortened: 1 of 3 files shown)


akegan:
does it make sense for the section titles to upcoming lotteries to have the same header level as the listings? not that it makes sense that in the past the listings were h1, just curious since you're in there thinking about it

cpalmatier:
I think it's fine to leave them all as h2 (they were all h1 before) since the 3 sections in the listings page themselves serve as the semantic "buckets" if you will, so the markup doesn't necessarily need to be quite that exacting in terms of meaning. The bigger problem here was that there were multiple h1s on the page, which is a no-no.

cpalmatier:
if what you meant was "do the section titles themselves need to be h2s, and the card titles are h3s" i guess we could do that, but i think if _that_ were a problem, Christina would tell us.

akegan:
Yeah pretty much. I think the biggest issue is if there is more than one h1 attribute on a page, but otherwise it's also nice to have a "logical reading order" that makes sense if the person using the screen reader is scanning through the page from header to header at a given level, and/or if they're reading through a summary of the whole page.

cpalmatier:
poking at the surrounding code for browsing listings, it looks like having section headers as h2 and the listing titles as h3 would be straightforward enough. that said, the sale-header also has a couple of rogue h2 elements (where rental-header does not), and it looks like the application intro (so a1-intro, which uses similar code to the property cards although not that exact partial) has similarly wack markup.

anyway if we want to expand scope i can get it into a somewhat better state here and in pattern-library and modify the ticket to reflect that. alternately we could do another ticket for a 2nd, more comprehensive pass. either way ok to me—what do you think?

akegan:
up to you! sounds like we planned for this one thinking that it would just be a small change, so I don't want to derail your sprint by adding more scope. I'm not sure what the overhead would be to pick this up again in a couple of sprints vs just finishing it here either. 

If you decide to do a follow-up ticket can you create it and link it on this PR?

cpalmatier:
i am inclined to just go ahead and sort it out now; i may regret saying this later but it sure _looks_ like it's still purely a markup thing and so should be pretty quick. further commits incoming here and on pattern-library this afternoon—i'll ping you again when they are ready to review.

akegan:
might be worth asking christina to do a quick check if she's up for it to make sure updates help? she has a compliance sheriff tool that should be able to automatically test it

cpalmatier:
Cool—I assume she can use the review app link? If so I'll ping her and ask her to take a look.

akegan:
yep ping her with the link, and point her to the specific pages you updated too (sale directory page, rental directory page, application intro page). thanks!

PR #1307, opened 2020-01-28
Bugs/spanish pricing table formatting #170459019

cpalmatier:
Fix an issue with the formatting of the pricing table when the user selects Spanish as the site language: https://www.pivotaltracker.com/story/show/170459019

This required altering a style that depended on a data attribute that itself is translated, so when a non-English language is selected, the style was not applied. This PR contains compiled CSS from the pattern library in commit 10c5666, part of PR https://github.com/Exygy/sf-dahlia-pattern-library/pull/114 as well as changes to pricing table markup where there is a data-th="Availability" attribute. These changes remove this dependency and applies the correct styling regardless of language.


akegan:
Same issue looks like it's happening on rental listings, probably worth updating both herein this PR 
image

akegan:
2nd row right-most column shifts over in chinese and spanish, not sure if that's something you want to fix in this PR 
image

image

akegan:
otherwise looks good, just let me know if you want to try to fix the right-hand column offset thing

cpalmatier:
Will merge but wanted to leave record re: the column-shift thing in Spanish and Chinese. Based on some investigation, it appears to be caused by Google Translate adding markup wrapped around translated strings. If that extra markup isn't there, the layout works properly. Alas there's no way for us to fix what GT is doing, though, so it will have to stand as it is.

PR #1311, opened 2020-01-31
Add logging for cases where application get for logged in user fails #159802520

akegan:
https://www.pivotaltracker.com/story/show/159802520

Just added sentry logging to see if we can get a better sense of what's going on with this error. see the ticket for more details


app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -662,6 +662,8 @@
                 ShortFormApplicationService.application.answeredCommunityScreening != 'Yes'
                   $state.go('dahlia.short-form-welcome.community-screening', {id: listing.Id, skipConfirm: true, lang: lang})
             ).catch( (response) ->
+              # Verify source of errors in https://www.pivotaltracker.com/story/show/159802520
+              console.error('Error getting my application for listing (angularRoutes, application)', $stateParams.id)
```
LoganArnett: Is there a reason we are not using Raven.captureMessage here or vice versa with console.error below?

akegan: When I was testing it out it looked like we don't have access to Raven here for some reason, so I left it as a console error instead hoping Sentry would pick it up by default.

PR #1316, opened 2020-02-12
Chores/fix flickering e2e tests #163066582

chrisdolendo:
https://www.pivotaltracker.com/story/show/163066582

Approach that made a huge difference for all of the flickering tests was waiting for the element to appear before :

```
EC = protractor.ExpectedConditions
elem = element(SELECTORHERE)
browser.wait(EC.presenceOf(elem), 5000)
```

I also increased the timeout universally on `conf.js`

For future tests, we get to apply this if there is a pattern of it flickering.

Tests passed cleanly several times locally and semaphore had a clean pass right at the first attempt!

### DIFF
--- a/spec/e2e/conf.js
+++ b/spec/e2e/conf.js
@@ -3,6 +3,8 @@ var config = require('./conf-shared')
 config.baseUrl = 'http://localhost:3000/'
 config.capabilities = {
   browserName: 'chrome',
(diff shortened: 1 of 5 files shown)


akegan:
looks good, just going to run e2e tests a few times in a row on semaphore to see if they keep passing 😄

PR #1321, opened 2020-03-13
Hotfix - Allow links in alert message

akegan:
This PR allows for links in the alert message so we can link out to a MOHCD page for more information about changes to the affordable housing application 

image

My main concern for this is that we could be opening ourselves up to some sort of vulnerability just allowing any html content from an env var, but I think that we're pretty safe

Review instructions:
1. Go to the heroku page for the review app, update the config settings to add a sample message with a link with the name ALERT_MESSAGE, e.g. `"Test some text and a link and that's about all <a href='https://google.com'> test link</a>"`
1. Go to the [review app](https://dahlia-full-hotfixes-al-h4xbaj.herokuapp.com/), and verify that the alert message shows up as expected and the link works.

### DIFF
--- a/app/assets/javascripts/shared/templates/alert-message.html.slim
+++ b/app/assets/javascripts/shared/templates/alert-message.html.slim
@@ -5,5 +5,4 @@
         span.ui-icon.ui-medium.i-white
           svg
             use xlink:href="#i-warning"
-        p.offset-icon.c-white.t-small.t-semi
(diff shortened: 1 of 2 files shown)


akegan:
Not sure why integration with heroku isn't working, but this has been deployed here: https://dahlia-full-hotfixes-al-h4xbaj.herokuapp.com/

PR #1322, opened 2020-03-19
Reset Alternative Contact ID on application #171706913

software-project:
Fixes issue with alternative contact being the same between applications.
https://www.pivotaltracker.com/story/show/171706913

### DIFF
--- a/app/services/force/short_form_service.rb
+++ b/app/services/force/short_form_service.rb
@@ -61,6 +61,7 @@ module Force
         name: nil,
         agreeToTerms: nil,
         shortFormPreferences: [],
+        alternateContact: { appMemberId: nil },
       }
       # reset income fields on apps > 30 days old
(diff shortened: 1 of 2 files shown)


app/services/force/short_form_service.rb
```diff
@@ -61,6 +61,7 @@ def self.autofill_reset(application, listing_id)
         name: nil,
         agreeToTerms: nil,
         shortFormPreferences: [],
+        alternateContact: { appMemberId: nil },
```
akegan: can we do the same thing with other application members? I think we want new ones for every application

akegan:
Tests weren't connected because of the switch to SFDigitalServices repo, just started a new job now that should be running.

app/services/force/short_form_service.rb
akegan: nevermind, seems like we're already OK.

PR #1325, opened 2020-04-16
Fix language selection on application #171215415

software-project:
Fixes: https://www.pivotaltracker.com/story/show/171215415


software-project:
Hmmm, tests are green. I must say I'm surprised @akegan. But I dont see multilingual.feature on semaphore, will double check that.

app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -653,7 +653,7 @@
                 # send them to their review page if the application is already submitted
                 $state.go('dahlia.short-form-review', {id: ShortFormApplicationService.application.id})
               else if ShortFormApplicationService.application.autofill == true
-                $state.go('dahlia.short-form-application.autofill-preview', {id: listing.Id, lang: lang})
+                $state.go('dahlia.short-form-application.autofill-preview', {id: listing.Id, lang: $stateParams.lang})
```
akegan, line 656: wow, surprised that this was enough!

akegan:
@software-project looks like something is happening where the new fixture is affecting the one that comes after it. I wonder if it it keeping the language or something? 
```
Feature: Short Form Application - Sale Listing
Scenario: Navigating through a basic application
```
is what comes after it on semaphore and it's getting hung up. 

Command to run to try to reproduce locally is `npm run protractor -- --specs './spec/e2e/features/short_form/!(submitting_application|sign_in_while_applying).feature'`

software-project:
@akegan this is the test I said I need a second pair of eyes on. Locally it's passing and I run Semaphore ~5 times and it's failing usually on the same thing. I run it once again now and got same results again :/

akegan:
yep it's on my list of things to do today

akegan:
found the issue. user is still logged in after the multilingual one. I'll try to fix quickly

akegan:
Wasn't able to fix it quickly @software-project you can add "And I sign out" to the end but it fails because the page is in filipino ☹️. tried to add step to switch back to english but ran out of time today. I can keep working on it tomorrow if you don't get to it overnight.

software-project:
@akegan kudos for noticing that. It seems like using same user name was the issue. I've also added sign out as it may also cause issues. Tests passed, I've merged changes from master now.

PR #1330, opened 2020-04-27
Fix user proof not uploading #172114799

software-project:
Fixes https://www.pivotaltracker.com/story/show/172114799


app/controllers/api/v1/short_form_controller.rb
```diff
@@ -160,18 +160,11 @@ def initial_submission?
   end
 
   def send_attached_files(application_id)
-    if user_signed_in?
-      files = UploadedFile.where(
```
akegan, line 164: I think that we need to also update where we're saving the files, right? otherwise the user_id will be set to not-nil and we won't be able to find it.

software-project: We have two options. We can remove `user_id: nil` from where statement here and leave it where we are setting it. Then in case we would need to debug sth that might me handy. 
Other option would be to remove it completely, including DB. Which version to you prefer @akegan

PR #1336, opened 2020-05-13
Fix screen reader for nav bar #169027064

james-wills-sf:
[#169027064](https://www.pivotaltracker.com/n/projects/1405352/stories/169027064)

## Summary
Previously, tab changes weren't being read by the screen reader. This change adds a hidden aria-live div that reads changes to the page title.

### Why did you put the div in the application.html, why not make each component header an aria-live region?
aria-live regions only read when content is changed, they don't register when the span with aria-live appears or disappears

### Why'd you use aria-live="polite"
Polite makes it not interrupt the current reading. Maybe it should interrupt it?

### What if we want to read something more customized than the page title?
I spent probably 6 hours researching how to plug in custom data to `application.html.slim` but every solution was fragile and I couldn't quite get it working. I think this is better than nothing as a first pass and if we want custom readouts I can come back to this.

## Screencast: note the titles are read after the new tab loads
tabs screen reader

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -72,14 +72,15 @@ html lang="en" ng-app="dahlia" ng-strict-di="" ng-controller="SharedController"
         section(ui-view="version")
         section(ui-view="navigation")
       .loading-overlay(bs-loading-overlay)
-        section#main-content.main-content(ui-view="container" aria-live="polite")
+        section#main-content.main-content(ui-view="container")
     .footer-section.skiptranslate
       .loading-overlay(bs-loading-overlay)
         section(ui-view="footer")
 
+    div.sr-only id="announce-text" aria-live="polite" {{$title || 'DAHLIA San Francisco Housing Portal'}}
+
     / icon collection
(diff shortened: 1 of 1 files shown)


akegan:
Seems like it works to at least register that something has happened on the page.

james-wills-sf:
@akegan I'm trying to get the latest review app to Christina, I found the review app on [heroku](https://dashboard.heroku.com/pipelines/9b135df1-372c-402b-a8d2-3db5f37c646b) but it said the build failed. I retried the build but it doesn't really look like it's doing anything. Do you know how I can check the progress of the build or force kick off a new build?

james-wills-sf:
@akegan Just letting you know that my "fix" to read out the page content apparently didn't work for Christina with JAWs/windows I went ahead and reverted it, and we can look into finding a fix for the content readout later.

I think it may have something to do with doubling up on `aria-live="polite"` and `role="alert"` but I'm not sure and it's hard to test without being able to reproduce locally

akegan:
OK we can follow-up again another time. thanks!

PR #1337, opened 2020-05-14
Remove text regarding optional attendance #172655306

software-project:
Remove language about "attendance is optional" from upcoming lottery listings.
https://www.pivotaltracker.com/n/projects/1405352/stories/172655306


.env.sample
```diff
@@ -7,6 +7,7 @@ GOOGLE_TAG_MANAGER_KEY=''
 S3_ACCESS_KEY=''
 S3_ACCESS_KEY_ID=''
 SHOW_SALE_LISTINGS=true
+SHOW_PRE_LOTTERY_INFO=false
```
james-wills-sf, line 10: Does this also need to be added to the `app.json` file? I'm not really sure what `app.json` is used for.

software-project: Not sure what is it for, but good catch, thanks.

software-project:
Thanks @james-wills-sf, should be good now.

james-wills-sf:
@software-project @akegan lgtm! For reference I don't need to be a blocker if I've already approved a PR, even if I've requested minor changes. I'll make that more explicit in my approval messages

PR #1342, opened 2020-05-26
Update listing details to accommodate AMI tiers  #171215384

akegan:
https://www.pivotaltracker.com/story/show/171215384


app/assets/javascripts/listings/ListingUnitService.js.coffee
```diff
@@ -29,49 +34,126 @@ ListingUnitService = ($http, ListingConstantsService, ListingIdentityService) ->
...
       summaries.push(summary)
     summaries
 
-  # Group units by AMI % and Unit type
-  # Returns an object of the form:
-  #  {'50 (percent ami)':
-  #     {'1br': [{rent/price, # of units, etc}, ...], '2br': [...]},
-  #   '40 (percent ami)': {...}}
+  Service._getAMIAmount = (incomeList, occupancy) ->
+    if incomeList && incomeList.values?
```
akegan, line 61: Not sure how best to handle cases where we can't find the AMI chart associated with the unit. This happens fairly rarely. Console.error should help it show up in Sentry, but I'm wondering if we should do something a bit more drastic so that it looks wrong for SMEs when they're previewing the listing.

app/assets/javascripts/listings/ListingUnitService.js.coffee
```diff
@@ -29,49 +34,126 @@ ListingUnitService = ($http, ListingConstantsService, ListingIdentityService) ->
...
+      console.error('Missing AMIs for occupancy', incomeList, occupancy)
+
+  Service._getIncomeRangesByOccupancy = (unitGroup) ->
+    # Given a row of units grouped by size, AMI, price, etc, determine the
+    # min and max incomes for that range for every available occupancy size.
+    # If a max occupancy isn't defined, as in the ownership case, default to showing up to 3
+    # occupancy options.
+    maxOccupancy = unitGroup.Max_Occupancy || unitGroup.Min_Occupancy + 2
+    occupancyRange = [unitGroup.Min_Occupancy..maxOccupancy]
+    # FIXME: This does not handle cases where listings have multiple AMI charts with the same percent.
```
akegan, line 73: right now I think what we do in the max income table is just pick one of the charts if there is > 1 kind of chart with the same year + percent. It's pretty rare (I couldn't find any that are live now). 

Wondering if it makes sense to fix in this PR or do in a follow-up ticket.

app/assets/javascripts/listings/ListingUnitService.js.coffee
james-wills-sf: maybe we can check in on the sentry logs in a couple weeks to see if this ever happens? If it has happened we can think about error handling.

Unless you know for a fact that this will happen

app/assets/javascripts/listings/ListingUnitService.js.coffee
james-wills-sf: This PR is pretty big, I'd be happy with splitting it into a separate PR if that's not too much of a hassle

james-wills-sf:
@akegan just finished QA and I found the following issues:

### 1. No rent price is listed for % income rents
This is probably the same issue you already found.

**reproducible listing id:** a0W0P00000Gs3V3UAJ


### 2. Borders don't go all the way across the view on small screen sizes
Only able to reproduce on some of the borders for certain listings
**reproducible listing ID:** a0W1D000001IJ6SUAW


### 3. On small screens, not enough spacing between value and labels
Could add a [non-breaking space](https://www.computerhope.com/jargon/n/nbsp.htm) between the spans.

akegan:
@cpalmatier  for the case of % income I'm also wondering if the minimum income should be $0 or should be something like "None", did you have a design for that?

app/assets/javascripts/config/angularRoutes.js.coffee
```diff
@@ -123,8 +123,9 @@
                 return deferred.promise
 
               # trigger this asynchronously, allowing the listing page to load first
-              setTimeout(ListingDataService.getListingAMI.bind(null, ListingDataService.listing))
-              setTimeout(ListingUnitService.getListingUnits.bind(null, ListingDataService.listing, forceRecache))
+              ListingUnitService.getListingAMI(ListingDataService.listing).then( ->
```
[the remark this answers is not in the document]
akegan: we need the AMIs to be able to calculate the units now, so I wanted to make sure that amis were present before we started grouping units. Could probably be done in a more sophisticated way in the service instead.

For the timeout, I was honestly just copying what was out there for the listing call. I wasn't sure if we could wrap both in setTimeout.

I'd like to maybe follow up with a ticket that does a better job of optimizing the calculations so there's less of a delay when the page loads

akegan:
OK I've fixed the issue with % income

I tried adding `&nbsp;`s everywhere to space out the #s from the "small" labels but realized that this is an issue on the listing directory too so we'd have to update it in ~15 places. If we can update the pattern lib that might be a better long-term solution.

cpalmatier:
@akegan answers re: a few of the above comments...

- $0 minimum income requirements: those should just be "None" with no "per month" added.
- the <small> spacing thing: i thought we ran into (and fixed) that before? in the pattern library at the "small" (confusing!) breakpoint, there was a rule added for margin-left: 0.125rem for <small> elements in tables with class .pricing-table. happy to take a look at the review app and diagnose.
- the border thing in @james-wills-sf 's #2 above is something i _think_ is fairly straightforward and would be a fine fast-follow if y'all don't think it's a showstopper.

akegan:
Thanks @cpalmatier  I will make the change to None (i think the code is in there but isn't quite working)

I thought we had solved the spacing thing too 🤔. Here is the review app if you're up for checking it out https://dahlia-web-full-pr-1342.herokuapp.com/listings/a0W0P00000Gs3V3UAJ

We can wait + follow-up on the border thing, maybe smoosh it together with the "moderate income" overlap fix.

akegan:
OK @james-wills-sf @cpalmatier I think I fixed all the issues we were going to fix in this pr:
1. Show rent if it's % income
1. Show "none" for min income if it's $0 on rental and sales
1. Added spacing on mobile view (thanks to Chris!)

let me know if it looks good and I'll merge!

PR #1349, opened 2020-06-11
Add .nvmrc with recommended node version

james-wills-sf:
Partners already has [an .nvmrc file](https://github.com/SFDigitalServices/sf-dahlia-lap/blob/master/.nvmrc), this just follows that pattern by adding an .nvmrc file with the recommended node version from our repo [README](https://github.com/SFDigitalServices/sf-dahlia-web#dependencies).

If we feel we don't want this file checked in, that's fine and I can update this PR to add .nvmrc to our .gitignore instead.


james-wills-sf:
Closing this pull request because it messes with the CircleCi npm install script. There is probably a better solution for this but it's not worth the investment now.

PR #1340, opened 2020-05-22
Fix duplicate translations on listing #172950053

software-project:
Fixes duplicate translations on values: https://www.pivotaltracker.com/n/projects/1405352/stories/172950053

PR #1351, opened 2020-06-15
Do not translate number of units available or household size

akegan:
This fixes #1340 again.

To review:
1. Check out e.g. Ed lee apartments in spanish and verify that numbers aren't duplicated


akegan:
Don't think we can fix those with this PR unfortunately. Google translate kind sucks. 

For the AMI percent one we have a translation string in there, just not translated yet, so that should be fixed in the next round of translations

For the bedrooms one, we pull it as e.g. "1 BR" from salesforce and don't have human translations for it so it's up to google translate's crazy antics. It's a known issue and we do have a ticket in the icebox for it [#167847946](https://www.pivotaltracker.com/story/show/167847946), and a separate ticket [#169413366](https://www.pivotaltracker.com/story/show/169413366) to spell out BR as bedroom to make it more accessible for folks who use screen readers.

PR #1359, opened 2020-06-23
[Webapp] Add circleci prod e2e tests #173302542

james-wills-sf:
[#173302542](https://www.pivotaltracker.com/n/projects/1405352/stories/173302542)

## Summary
Add tests that will run only on the "production" branch with prod env
variables.

## Steps taken to verify
- [x] added this pr name to the e2e-production conditional to test that it runs the production tests (and skips the rest)

- [x] removed this pr name to the e2e-conditional to test that regular branches are unchanged.

- [x] sshed into e2e-production box and verified that it's using prod credentials.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,6 @@ prod_and_release_jobs: &prod_and_release_jobs
     branches:
       only:
         - /^release.*$/
-        - chores/circle-ci-on-production-#173302542
         - production
 aliases:
   - &attach_workspace
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,7 +183,6 @@ workflows:
             branches:
               ignore:
                 - production
-                - chores/circle-ci-on-production-#173302542
       - e2e-submit-application:
           requires:
             - build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -151,6 +157,20 @@ jobs:
     steps:
       - run-e2e-tests-for-path:
           path: "./spec/e2e/features/short_form/!(submitting_application|sign_in_while_applying).feature"
+  e2e-production:
+    <<: *defaults
+    steps:
+      - setup
+      - start-server
+      - run-e2e-tests-for-path:
```
akegan, line 165: is the idea to run prod tests all serially? or just waiting til marcin's pr to parallelize

james-wills-sf: @akegan You know, this was just out of wanting to keep the config simple. I wasn't thinking at all about serial vs parallel. I think it would be better to make these parallel, changing it now!

james-wills-sf:
Ran a test with this in "release branch mode" -- where prod and dev tests are run. All tests passed!

- 2020-06-25 at 12 12 12 PM

PR #1362, opened 2020-06-25
Fix missing listing directory unit rows #173504337

akegan:
This PR fixes an issue where unit groups with min % AMI instead of min income were not showing in the listing directory page
[ticket](https://www.pivotaltracker.com/story/show/173504337)

Before
image

After
image


app/assets/javascripts/listings/templates/property-card/_rental-stats-row.html.slim
```diff
@@ -1,4 +1,4 @@
-tr ng-repeat-start="unit_summary in ::summary" ng-if="::unit_summary.minRentalMinIncome || unit_summary.maxRentalMinIncome || unit_summary.minPercentIncome"
```
james-wills-sf, line 1: Why replace these two instead of adding an additional `|| minMonthlyRent`?

akegan: honestly I'm not sure why we need any of them, but for now we're not displaying anything about the income on here, so there's no need to filter on it. minPercentIncome is really the min rent as percent income, e.g. 30% on the alexander residence.

PR #1365, opened 2020-06-26
Upgrade devise_token_auth to fix emails #173504864

james-wills-sf:
[#173504864](https://www.pivotaltracker.com/n/projects/1405352/stories/173504864)

### The bug
Users dont get a confirmation email when they change their emails

### What was really happening
devise_token_auth overrides some methods that devise uses to send emails. See this [issue](https://github.com/heartcombo/devise/issues/3654) for more information. We had a workaround for that where we overrode `has_email?` in our user object, but after we upgraded rails to 5.1 that override stopped working. That's because `has_*` methods were changed to `saved_change_to_*` and we didn't update the name of our override method.

### The fix
The devise_token_auth issue was fixed by this pr: https://github.com/lynndylanhurley/devise_token_auth/pull/1343 which is in [release 1.1.4 of devise_token_auth](https://github.com/lynndylanhurley/devise_token_auth/releases). This change updates that library to include the fix in webapp.

Now that devise_token_auth is working properly, we can remove the `has_email?` and `update` overrides, along with the manual boolean we were storing.

Additionally, the devise library appears to have an issue in 4.7.2, so I explicitly downgraded us to 4.7.1 and added a comment, we should be careful to update this until the issue is fixed and the release is verified.


akegan:
great investigative work! were you able to test what effect this has on users who currently have un-confirmed emails? Does this change make the unfconfirmed email column work again?

james-wills-sf:
@akegan not yet, I can spend the rest of my time today trying to figure all of that out.

I'm  99% sure this will fix the unconfirmed email column but haven't verified yet, will do and get back to you

akegan:
I added a simple rspec test to make sure that the email is being sent for reconfirmation. I was able to run the same test against main and it failed, so I think this is ready to go if tests pass.

PR #1370, opened 2020-07-08
[Webapp] Fix squished sale listings #173618329

james-wills-sf:
[#173618329](https://www.pivotaltracker.com/n/projects/1405352/stories/173618329)

Related Pattern Library PR: https://github.com/Exygy/sf-dahlia-pattern-library/pull/127/files

The AMI tiers table is too squished at medium screen sizes, this change adds the new Pattern Library changes to force the table into mobile viewat tablet screen sizes.

## Question for Nicole
it looks like our page actually switches to "tablet" mode at the large breakpoint, not medium. If you look closely in the gif just before the table switches to the mobile view, the rest of the page snaps to tablet view.

This definitely fixes the spacing bug but would it be better for us to show the mobile table for large and below instead of medium and below to stop the extra snapping on page resize?

### Screencast: Highlighting extra snapping between medium and large viewports
hosting this screencast here so we can embed it in pivotal. **Update: we've updated this PR to set the breakpoint to large so this extra snapping doesn't happen.** 
webapp-highlight-snapping

## Update: spanish and filipino
Even with the large breakpoint change this table doesn't look right for spanish and Filipino.
### Spanish
- 2020-07-10 at 11 17 22 AM
### Filipino
- 2020-07-10 at 11 19 21 AM


james-wills-sf:
@akegan since we're splitting this fix into two tickets, this is ready for review, along with the [pattern library change](https://github.com/Exygy/sf-dahlia-pattern-library/pull/127)

akegan:
ok! does "mobile for tablet" still make sense with the switch from medium to large breakpoints? that's my only bit of feedback on both

james-wills-sf:
@akegan ah sorry I forgot about that comment. Let me think a little about the naming

james-wills-sf:
@akegan i think using the word "tablet" in the class name is a mistake since it's a little ambiguous whether tablet size is medium + large or just medium, so I changed the class name to `with-mobile-styling-at-large-breakpoint`, let me know what you think!

akegan:
@james-wills-sf OK thanks for making the quick change. I'll do QA on the review app then approve

PR #1372, opened 2020-07-09
Remove usage of SHOW_SALE_LISTINGS env var #172845330

akegan:
Picked up from the small-wins pile. [Ticket](https://www.pivotaltracker.com/story/show/172845330)
After we release we'll need to remove the env var from the prod app. 

# Review instructions
Goal is just to check that al the places where we were using the flag still look OK (the same as before)

1. Top nav bar
image

1. Mobile nav
image

2. Empty My-favorites page
image

3. Empty My-applications page
image

### DIFF
--- a/spec/javascripts/controllers/nav_controller_spec.coffee
+++ b/spec/javascripts/controllers/nav_controller_spec.coffee
@@ -1,8 +1,7 @@
 do ->
   'use strict'
   describe 'NavController', ->
-    fakeWindow = {}
-    fakeWindow['env'] = {showSaleListings: 'true'}
+
     fakeDocument = {}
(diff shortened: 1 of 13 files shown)


spec/javascripts/controllers/nav_controller_spec.coffee
```diff
@@ -1,8 +1,7 @@
 do ->
   'use strict'
   describe 'NavController', ->
-    fakeWindow = {}
-    fakeWindow['env'] = {showSaleListings: 'true'}
+
```
akegan, line 4:
```suggestion
```
Remove unnecessary whitespace

PR #1384, opened 2020-07-20
File upload test improvements #173163945

software-project:
Fixes: https://www.pivotaltracker.com/story/show/173163945


software-project:
@akegan @james-wills-sf I suppose we don't want to have that `file test` endpoint active for production, right? Have a look on other things and let me know what should I change. Locally it worked, so fingers crossed ;)

software-project:
I've refactored js logic and added constrain to routes. So it's looking better now. Tests seems to be successful, so thats good. Only thing that is still bugging me is if we should test testing logic ;) besides that ready for review.

config/routes.rb
```diff
@@ -1,3 +1,10 @@
+# Constrain class that rejects production request
+class RestrictProduction
```
james-wills-sf, line 2: nice!

software-project:
@james-wills-sf I did a production test as requested: https://circleci.com/workflow-run/3ed166d7-ac34-4253-9d52-6e4bb9c19af3 it's green, so all good. Also the test is running in the suite, see below.
Zrzut ekranu 2020-07-28 o 14 31 06
Also the typo is fixed so we should be good to go.

PR #1386, opened 2020-07-28
Update design for listings directory table #173346911

james-wills-sf:
Ticket [#173346911](https://www.pivotaltracker.com/n/projects/1405352/stories/173346911)

### Screenshot: sale desktop view -- very long text edge case

todo:
- [ ] remove the mock data
- [x] fix sale listings mobile view
- [x] fix rental listings mobile view
- [x] tests
- [x] a11y pass


akegan:
@cpalmatier we have a question on how to make these changes work for mobile view:

## What we have currently
for rental:
image
for sale:
image

## What it looks like just making the changes in the mock-ups
for rental:
 image
and sale:
image

## The issues
The text overlaps on mobile view for sales and on rental, the 1BR plus availability is a bit cramped. 

## Potential solutions
1. Maybe it would look better if we go back to having the small text (like per month) go below the values? 
1. Or maybe we have to drop some of the info on mobile view or on the directory view for now, until we can revisit?
1. James has offered to break these tables up into separate tables for mobile and desktop view so we don't have to rely on css magic to make the mobile tables. We could do that in the scope of this ticket if the changes are small, or tackle a bigger refactor when we tackle the mobile view for the listing details tables?

cpalmatier:
@akegan @james-wills-sf I think the idea of separating out mobile and desktop components for pricing tables is one whose time has come! Let's do it.

Given separate mobile components here, I think the idea of "break[ing] the HOA dues onto their own line in addition to the 'Sales price' line" is the correct solution. I would start with that, and then consider pulling out some info from the mobile view if we're still having trouble fitting everything in a reasonable way (lmk if we need to do that, and I can offer suggestions about what to pull out). I'm also fine with moving the small text onto a 2nd line (as it was before) but _only_ if there's no alternative—keeping the label and value text on the same baseline makes it a little easier to parse, and these views are complex enough that every little bit helps in terms of readability.

james-wills-sf:
@cpalmatier I just updated the screenshots in the PR description with the new mobile views. I moved availability and HOA to their own lines. Does this seem okay to you? We can always skip lines entirely for mobile if we feel it's too long

james-wills-sf:
@akegan I went ahead and marked this PR as ready for review. It's still using mock data but everything else is done. Up to you if you want to wait to review til after the salesforce change is in, but I don;t anticipate anything but the mock code to change.

app/assets/javascripts/listings/components/rental-stats.html.slim
```diff
@@ -1,11 +1,33 @@
...
+    / Reserved
+  table.table-pricing.table-pricing-hero.table-reserved ng-if="::$ctrl.listing.unitSummaries.reserved"
+    caption ng-if="::$ctrl.listingContainer.listingIsReservedCommunity($ctrl.listing)"
+      span translate="listings.reserved_units_for_who_are"  translate-value-community-type="{{::$ctrl.listingContainer.reservedLabel($ctrl.listing, $ctrl.listing.Reserved_community_type, 'reservedFor')}}" translate-value-reserved-type="{{::$ctrl.reservedForLabels($ctrl.listing)}}"
+    caption ng-if="::!$ctrl.listingContainer.listingIsReservedCommunity($ctrl.listing)"
+      span translate="listings.reserved_for" translate-value-type="{{::$ctrl.reservedForLabels($ctrl.listing)}}"
+
+    tbody ng-include="'listings/templates/property-card/_rental-stats-row-mobile.html'" ng-init="summary = $ctrl.listing.unitSummaries.reserved"
+
+/ Destop table
```
akegan, line 22:
```suggestion
/ Desktop table
```
minor typo

app/assets/javascripts/listings/components/sale-stats.html.slim
```diff
@@ -1,12 +1,35 @@
...
+    / Reserved
+  table.table-pricing.table-pricing-hero.table-reserved ng-if="::$ctrl.listing.unitSummaries.reserved"
+    caption ng-if="::$ctrl.listingContainer.listingIsReservedCommunity($ctrl.listing)"
+      span translate="listings.reserved_units_for_who_are"  translate-value-community-type="{{::$ctrl.listingContainer.reservedLabel($ctrl.listing, $ctrl.listing.Reserved_community_type, 'reservedFor')}}" translate-value-reserved-type="{{::$ctrl.reservedForLabels($ctrl.listing)}}"
+    caption ng-if="::!$ctrl.listingContainer.listingIsReservedCommunity($ctrl.listing)"
+      span translate="listings.reserved_for" translate-value-type="{{::$ctrl.reservedForLabels($ctrl.listing)}}"
+
+    tbody ng-include="'listings/templates/property-card/_sale-stats-row-mobile.html'" ng-init="summary = $ctrl.listing.unitSummaries.reserved"
+
+/ Destop table
```
akegan, line 23:
```suggestion
/ Desktop table
```
same typo

app/assets/javascripts/listings/templates/property-card/_rental-stats-row.html.slim
```diff
@@ -1,15 +1,17 @@
 tr ng-repeat-start="unit_summary in ::summary" ng-if="::unit_summary.minMonthlyRent || unit_summary.minPercentIncome"
-  td.is-subtitled data-th="{{'t.unit_type' | translate}}"
+  td.is-subtitled data-th="{{'t.units' | translate}}"
```
akegan, line 2: should we rename these files to -desktop just as a hint that there's another file out there?

james-wills-sf:
@akegan updated this to have the new sales table layout, and to handle null absoluteMinIncome values for rentals. Even though there is still the salesforce bug for AMI tiers listings, I still think this is ready for re-review.

### Desktop sales
- 2020-08-12 at 1 47 46 PM

### 320px sales
- 2020-08-12 at 1 48 08 PM

### Tagalog desktop sales
- 2020-08-12 at 1 50 43 PM

### Tagalog 320px sales
Note that there is some text cutoff, we decided in the related pattern library PR that we're not going to fix this for now.
- 2020-08-12 at 1 51 06 PM

### Rental with $0 minimums
Technically this is still a bug, these particular minimums shouldn't be $0 and are currently being fixed on the backend, but they show what this looks like when absoluteMinIncome is null.
- 2020-08-12 at 1 55 49 PM

james-wills-sf:
@akegan sounds good. I'll look into the missing rent.

Missing income range is expected, there are only 3 populated rental listings on full for us, a0W0P00000HbryHUAR, a0W0P00000HE9sfUAD, anda0W1F000002DHG3UAO, so I wouldn't expect Alexander to have the right income info. It should have rent info though...

akegan:
@james-wills-sf can we add alexander residence to the listings Larry is testing with so we have one of the % income listings as a test? That's the only case where we'd expect there to be no min income for rentals so would be nice to check that case.

akegan:
@james-wills-sf income ranges on listing directory look great. Looks like rent is missing on alexander residence still

james-wills-sf:
@akegan looks like I caused a regression with units where rent is calculated by % income. This is fixed now:

- 2020-08-17 at 12 07 58 PM

- 2020-08-17 at 12 08 16 PM

PR #1388, opened 2020-07-29
Add missing to ListingUnitService #173695074

software-project:
Fixes https://www.pivotaltracker.com/story/show/173695074

Adds a missing reference to `$q` to `ListingUnitService`. Bug occurs when listing is missing units (eg. listing `a0W0P00000DZDKZUA5`)


akegan:
OK thanks! I wonder if it was MOHCD staff looking at listings that don't have units yet 🤔. Thanks for figuring this out

PR #1394, opened 2020-08-01
Pull in latest phrase translations

akegan:
Done:
1. Pull in latest translations
1. Fix overwritten callout button 
1. Stop pulling english from phrase

Todo:

- [x]  Figure out how to get push to delete unused keys - **added to README**
- [x]  Improve workflow documentation (e.g we should maybe only push english by default and pull other languages. Make updates in phrase if we want to update non-english strings) - **Updated readme and script to only pull non-english and only push english**
- [x]  Get filipino translations reviewed (Not sure how to do this) - **Sent to msol for review**


akegan:
@james this is ready for review now. Let me know what you think about the changed process in particular.

README.md
```diff
@@ -93,12 +93,19 @@ To run stress testing against the Salesforce instance, refer to the documentatio
 To get started working with our Phrase translations, you will need to:
 
 1. Install the Phrase CLI with `brew tap phrase/brewed && brew install phrase`
-1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in e.g. LastPass
+1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in Lastpass and as a local env var.
 
-**To upload new strings or translations to Phrase**
+- Each time you add new strings to locale-en.json, push to Phrase to keep it up-to-date
+- If you delete keys from locale-en.json, run phrase push, then run phrase cleanup to delete unused keys
```
james-wills-sf, line 99:
```suggestion
   - Each time you add new strings to locale-en.json, push to Phrase to keep it up-to-date
   - If you delete keys from locale-en.json, run phrase push, then run phrase cleanup to delete unused keys
```

nit: 3 spaces before the `-` will indent this like a sub-bullet

README.md
```diff
@@ -93,12 +93,19 @@ To run stress testing against the Salesforce instance, refer to the documentatio
...
+```
+phrase uploads cleanup --id=[upload id from the response of phrase push]  --access_token=[your access token]
+```
+- If you want to make updates directly to non-english languages, you need to update them in phrase, then pull in the latest values.
```
james-wills-sf, line 103:
```suggestion
   - If you want to make updates directly to non-english languages, you need to update them in phrase, then pull in the latest values.
```

nit: 3 spaces before the `-` will indent this like a sub-bullet. Posting this as a separate comment because the ticks on line 100/102 break up the suggestions

app/assets/json/translations/locale-es.json
```diff
@@ -1004,7 +1032,6 @@
                 "title": "Antes de presentar la solicitud"
             },
             "callout": {
-                "button": "Vea los anuncios",
```
james-wills-sf, line 1007: 👍  thanks for removing these

README.md
```diff
@@ -93,12 +93,19 @@ To run stress testing against the Salesforce instance, refer to the documentatio
...
+
+**To upload new strings to Phrase**
```
james-wills-sf, line 105: are these backwards? Seems like this is for downloading, and the `grunt phrasePush --phraseAccessToken=[your access token]` command is for uploading.

README.md
```diff
@@ -93,12 +93,19 @@ To run stress testing against the Salesforce instance, refer to the documentatio
...
 
 1. Install the Phrase CLI with `brew tap phrase/brewed && brew install phrase`
-1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in e.g. LastPass
+1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in Lastpass and as a local env var.
 
-**To upload new strings or translations to Phrase**
+- Each time you add new strings to locale-en.json, push to Phrase to keep it up-to-date
+- If you delete keys from locale-en.json, run phrase push, then run phrase cleanup to delete unused keys
+```
+phrase uploads cleanup --id=[upload id from the response of phrase push]  --access_token=[your access token]
```
james-wills-sf, line 101: Can we make this part of the `phrasePush` script? As long as there's no downside to running cleanup when nothing is deleted.

README.md
```diff
@@ -93,12 +93,19 @@ To run stress testing against the Salesforce instance, refer to the documentatio
 To get started working with our Phrase translations, you will need to:
 
 1. Install the Phrase CLI with `brew tap phrase/brewed && brew install phrase`
-1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in e.g. LastPass
+1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in Lastpass and as a local env var.
 
-**To upload new strings or translations to Phrase**
+- Each time you add new strings to locale-en.json, push to Phrase to keep it up-to-date
```
james-wills-sf, line 98: I just realized these aren't really sub-bullets of the steps, can we make this section look something like:
```
## Translation processes
**When strings are added to or deleted from locale-en.json**
run phrasePush after merging your PR // contingent on if you can always run cleanup from phrasePush script.

**When you need to directly update non-english phrases**
   1. Update the phrases in Phrase
   2. run phrasePull 
```

Then you can delete the  **To upload new strings to Phrase** and  **To download new translations from Phrase** sections since those commands are covered above

README.md
akegan: 🤦‍♀️ . will fix

README.md
akegan: I'm not sure how to get the upload id from the output, but I'll spend a bit of time looking into it

akegan: looks like it's not supported with grunt-exec unfortunately https://github.com/jharding/grunt-exec/pull/31. I'm going to open an issue in the phrase_cli repo to ask for a option on push to delete unused keys instead if that's alright with you!

Other option would be to spin our a separate shell file, but this seems like something other folks would want to do whenever they push too?

james-wills-sf: @akegan yep seems like something others would want, I'm fine with just opening an issue and following up if they resolve it.

akegan: Ok cool. Here's the issue I created: https://github.com/phrase/phrase-cli/issues/22

akegan:
@james-wills-sf updates made! Plus a bonus update thanks to Shawn's issue (https://github.com/phrase/phrase-cli/issues/18) where it turns out we don't need to specify the access token if we have it set as an env var. 


Unfortunately, I don't think we can use any of the phrase integration with github because of the way our files are structured with the top-level locale keys. 

As a part of the rewrite we'll probably have to pull out the ruby (email) translations again and then maybe we can put them back into a yml format with the top-level language keys and see if phrase can handle that. The frontend locale files shouldn't need the top-level language key so then we'd be able to use the api and/or github integration. Then we'd have to deal with 2 different files, or 2 different projects(?) tracking ruby vs react translations, but it might still be better!

README.md
akegan: updated! let me know if changes make sense. We have to pull for the case that we have new translations from our partners, and also for the case where we rename keys and have to update the translations in phrase

README.md
```diff
@@ -89,17 +89,27 @@ In order to test caching locally,
...
 To get started working with our Phrase translations, you will need to:
 
 1. Install the Phrase CLI with `brew tap phrase/brewed && brew install phrase`
-1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in e.g. LastPass
+1. [Create an access token for Phrase](https://app.phrase.com/settings/oauth_access_tokens). Save it for future use in Lastpass and as a local env var.
+1. Save the access token as an env var so you don't have to pass it to the phrase commands `export PHRASE_ACCESS_TOKEN="YOUR_ACCESS_TOKEN"`
 
-**To upload new strings or translations to Phrase**
-Run `grunt phrasePull --phraseAccessToken=[your access token]`
+### When strings are added to or deleted from locale-en.json
```
james-wills-sf, line 99: Can you split this into the following cases:

When locale-en.json is edited (new keys added only)
When locale-en.json is edited (when some keys are added and some are deleted)
When locale-en.json is edited (when keys are renamed)
When locale-en.json is edited (when keys keep the same name but translation strings change)

Even if some of them have the same processes it would be helpful, since right now I don't know what to do for the bottom two.

akegan:
@james-wills-sf let me know if this is clearer tomorrow! I agree that the process is a bit confusing. The rule is really changing to no manual updates to non-english locale files. And phrase doesn't really have the concept of renaming keys, that just counts as deleting the old key and making a new key.

akegan:
ok @james-wills-sf just pushed again. Let me know if that's any clearer 🙈 

I found a param to set to get the english strings to update when pushed. I just reached out to our contact at Phrase to ask about unverifying translations if English strings are updated. It doesn't seem to be workign right now, so for now, I added a note to the readme about manually unverifying them. 

My guess is that as we procure Phrase and iterate on our process we'll be able to make things a bit clearer and easier.

README.md
```diff
@@ -89,17 +89,31 @@ In order to test caching locally,
...
-**To upload new strings or translations to Phrase**
-Run `grunt phrasePull --phraseAccessToken=[your access token]`
+### Push your changes to Phrase every time you update locale-en.json
 
-**To download new strings or translations from Phrase**
-Run `grunt phrasePush --phraseAccessToken=[your access token]`
+Run `grunt phrasePush` to push to Phrase each time you update locale-en.json to keep Phrase up-to-date
+
+Special cases:
+- If you changed the meaning or intent of an existing English string, you should delete the existing translations in Phrase and run `grunt phrasePull` so users don't get incorrect info.
```
james-wills-sf, line 104: should this be `delete the existing non-English translations`?

PR #1397, opened 2020-08-06
Display rich text for Venue #174197523

software-project:
Fixes: https://www.pivotaltracker.com/story/show/174197523
Adds rich text option for venue description.


james-wills-sf:
@software-project change looks good to me!

cc @akegan @cschlessinger looks like the text in salesforce should be updated to include a space. Just something to keep in mind when we start getting html content in prod

## Screenshot: html before this change
Screen Shot 2020-08-06 at 9 48 35 AM


## Screenshot: after this change
- 2020-08-06 at 12 13 29 PM

james-wills-sf:
Actually, it looks like this change also adds markdown parsing to the process section. What section is that @software-project ? I don't see it mentioned in the ticket

software-project:
@james-wills-sf I wasn't mentioned in the ticket, but I assume it should be also updated as it seems to me like it's using same field, so I thought it should look consistent. Was I wrong?

james-wills-sf:
I guess if it's using the same field we have to do the same parsing. @akegan do you know what the process section is so I can check it in qa?

app/assets/javascripts/listings/components/listing/process-section.html.slim
```diff
@@ -24,8 +24,7 @@ accordion-heading
             | {{::session.Date | dateSuffix}}
           span.content-group_time
             | {{::session.Start_Time}}
-        p.content-group_address.c-steel.no-margin.skiptranslate ng-if="::session.Venue"
-          | {{::session.Venue}}
+        p.content-group_address.c-steel.no-margin.skiptranslate ng-if="::session.Venue" ng-bind-html="::session.Venue | stripMostTags | nl2br"
```
akegan, line 27: I think this is for info session venues vs open house venues. I'll ask if we want to update those too, but for now they are not being updated in salesforce

akegan:
@james-wills-sf re: the missing space. In salesforce it actually looks just like a jira ticket or something with formatting options, so the SMEs don't have to generate the html themselves. 
image

app/assets/javascripts/listings/components/listing/process-section.html.slim
akegan: asked in slack here: https://sfinnovation.slack.com/archives/CPZHGET3R/p1596818833032800

akegan:
ok @software-project @james-wills-sf Msol updated info sessions to have those be rich text too. Let's test those out and then should be good to merge with these changes 😄

akegan:
@james-wills-sf @software-project is this PR ready to merge? I think the last piece was to do QA for the updated info session field that should now support rich text too

app/assets/javascripts/listings/components/listing/process-section.html.slim
james-wills-sf: Decision was to go ahead and support rich text in this as well as we might want links here in the future.

james-wills-sf:
@akegan thanks for reminding me about this. You're right that's all we need to do. I don't even think we need to QA the new field with rich text since we don't have any yet, just to make sure there is no regression for regular text. Where on the page is the info session field displayed so I can check?

akegan:
@james-wills-sf it should be on the sidebar too. Looks like someone already added one for [ed lee apartments](https://dahlia-full.herokuapp.com/listings/a0W0P00000HbryHUAR)? 
image

akegan:
alright so I just did some QA and it turns out that we don't allow for most of the features of rich text, but we do allow for links!

because of our "strip most tags" modifier we don't render: bold, italics, colors, new lines, bulleted lists. 
image

PR #1398, opened 2020-08-11
Stop sending gis error notifications #173779062

akegan:
Original Pivotal ticket: https://www.pivotaltracker.com/story/show/173779062

Instead of adding more GIS notifications, this PR removes the email notifications for geocoding errors completely. We are instead replacing these emails with Papertrail logging alerts, to better match the system we have for monitoring and responding to other errors. This means that we also do not need to create or maintain an email alias for these alerts. 

To Review:
1. Verify that there is no remaining unused code around these notifications
1. Verify that the [alert we setup on Papertrail](https://papertrailapp.com/alerts/2186782/edit?return_to=%2Falerts) will cover the use cases of errors we see here.


akegan:
@james-wills-sf NeighborhoodBoundaryService should definitely still be in use! Just trying to remove the email sending, not the checking of boundaries. I'll look at the other 2 pieces too.

akegan:
@james-wills-sf just pushed up removal of the emailer preview pieces, let me know if there was code somewhere else that I am missing. I'm not sure where else ArcGISService::GeocodingService::NAME  is tbh. Have to hop on a meeting now, but hoping we can maybe merge before I head offline today

james-wills-sf:
@akegan so GeocodingService::NAME is just the NAME constant in GeocodingService: https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/services/arc_gis_service/geocoding_service.rb#L11

This LGTM as is if the NeighborhoodBoundaryService is being used though, so you can go ahead and merge!

akegan:
OK thanks! totally just missed it. I removed NAME from geocoding and NRHP boundary service check. 

NeighborhoodBoundaryService is still used here: https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/controllers/api/v1/gis_controller.rb#L42

PR #1400, opened 2020-08-11
Add hardcoded solution to not show SRO box for Antonia Manor listing #174281476

akegan:
https://www.pivotaltracker.com/story/show/174281476


akegan:
we're actually not going to do this after all. closing

PR #1401, opened 2020-08-12
Fix failing review apps and fix failing e2e tests

akegan:
This PR tries to fix 2 things at once:
1. Fix flaky e2e tests: I added a wait as Marçin suggested and it tests keep passing so I think that solves our issue
1. I updated the post-deploy script to see if it'll stop causing the error below on review apps. I removed `rake db:setup` because of the message from Heroku support (below) and `rake listing_images:process_images` because we have been OK without it running for a long time and because it'll run automatically if we turn on caching. 

image

Suggestion from Heroku support:
> db:setup tries to also create the database, but it doesn't actually have permission to do so. db:schema:load, db:structure:load or db:migrate are the only rake commands we formally support for scenarios that involve feeding starter data. It's possible that in the past db:setup either didn't explicitly try to create the database or didn't bother to try if the database already existed.

> In the two-step process you have, swapping db:setup with db:schema:load should do the trick.

I tried a couple of variations and this one seems to work best. I tried rake db:schema:load instead of db:setup but it caused issues with "destroying a production database". This script matches what we have in Partners.

### DIFF
--- a/app.json
+++ b/app.json
@@ -31,7 +31,7 @@
         "heroku-postgresql"
       ],
       "scripts": {
-        "postdeploy": "rake db:schema:load && rake db:migrate && rake listing_images:process_images"
+        "postdeploy": "rake db:migrate && rake listing_images:process_images"
       }
     }
   },
--- a/app.json
+++ b/app.json
@@ -24,12 +24,15 @@
   ],
   "environments": {
     "review": {
-    "addons": [
-      "newrelic",
-      "papertrail",
-      "memcachier",
-      "heroku-postgresql"
-      ]
+      "addons": [
+        "newrelic",
+        "papertrail",
+        "memcachier",
+        "heroku-postgresql"
+      ],
+      "scripts": {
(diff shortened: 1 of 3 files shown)


akegan:
Alright we've got a review app that successfully redeployed after a change plus a lot of green checks on CircleCI:
image
image

app.json
```diff
@@ -24,12 +24,15 @@
...
-      "heroku-postgresql"
-      ]
+      "addons": [
+        "newrelic",
+        "papertrail",
+        "memcachier",
+        "heroku-postgresql"
+      ],
+      "scripts": {
+        "postdeploy": "rake db:migrate"
```
james-wills-sf, line 34: So if we ever add an environment we should expect that to fail too, right? I guess we can cross that bridge when we get to it

akegan: yeah I think it might but it might also have to do with review apps and their dbs specifically, I don't know. Kinda was planning to just cross it when we get to it

PR #1411, opened 2020-09-08
Add lighthouse workflow

akegan:


.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - id: deployment
+        uses: SFDigitalServices/wait-for-deployment-action@v2
+        with:
+          github-token: ${{ github.token }}
+          interval: 300
+          timeout: 5000
```
shawnbot, line 15: I think we could set this to 10 minutes, or 600 seconds. That should prevent you from hitting API rate limits if the deployment fails for some reason.

```suggestion
          timeout: 600
```

lighthouserc.js
```diff
@@ -0,0 +1,38 @@
...
+  LHCI_COLLECT_BASE_URL: baseUrl = 'http://localhost:3000',
+} = process.env
+
+const testListingId = 'a0W0P00000F8YG4UAN'
+
+const urlsToTest = [
+  `${baseUrl}/`,
+  `${baseUrl}/listings/for-rent`,
+  `${baseUrl}/listings/${testListingId}`,
+  `${baseUrl}/listings/${testListingId}/apply/name`
```
shawnbot, line 18: It looks from the statuses like these URLs have a double slash (`//`) at the beginning. I think we could fix this with [url-join](https://github.com/jfromaniello/url-join):

```suggestion
  join(baseUrl, '/'),
  join(baseUrl, '/listings/for-rent'),
  join(baseUrl, `/listings/${testListingId}`),
  join(baseUrl, `/listings/${testListingId}/apply/name`)
```

(assuming `const join = require('url-join')` at the top and `npm install -D url-join`)

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,23 @@
+name: Lighthouse
+on: [push]
+
+jobs:
+  lighthouse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
```
shawnbot, line 9: @akegan I think this is what we would need to do to give us the option of importing node modules in the lighthouse config:

```suggestion
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v1
      - run: npm ci
```

akegan: looks like it worked 🎉

akegan:
Performance #s are truly abysmal. I think a lot of it is the lack of caching + the slow initial page load time on review apps only. A11y numbers should be super useful though

PR #1412, opened 2020-09-23
Run Lighthouse CI on successful deployments

shawnbot:
This is an update to the Lighthouse workflow that should only run the Lighthouse CI workflow when any deployment is successful, by responding to the [`deployment_status` event](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#deployment_status).

I had to write [another action](https://github.com/SFDigitalServices/git-the-branch) to figure out which branch the commit SHA belongs to, since only `push` events set `GITHUB_REF`, and actions/checkout doesn't appear to bother doing the "reverse" lookup (with `git name-rev`).

The other thing I changed to speed up these runs is that the `npm install` step only installs `url-join` explicitly, rather than all of the dev dependencies. 🚀

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -7,12 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.deployment.sha }}
+
+      - name: determine branch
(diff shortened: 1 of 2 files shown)


akegan (approved):
looks good. we'll see how it goes post-merge. 

I'll also re-run the e2e tests to get them to pass so we can merge this btw

akegan (approved):
ready to go if we think it'll work!

PR #1413, opened 2020-09-24
Fix max income table missing row issue DAH-530

akegan:
DAH-530

I did a few things in this PR:
1. Fix the issue where showAMItoggler thought that the max value was 9 because it was looking at the "last" AMI chart
1. Generalized the way that we decide how many rows to show in the max income table to look across all AMI chart values for the max available value instead of just looking at the first value
1. Explicitly only look for the household size range in _income-table-multiple.slim.html instead of passing around irrelevant income values
1. Add more tests/improve tests for related functions. 
1. Fixed a bug where we were showing "/year" for missing AMI values 
image
1. Switched from "/month" and "/year" to "per month" and "per year" for better experience for screen readers, and started using human translations. 

Review instructions are in the ticket.

### DIFF
--- a/app/assets/javascripts/customFilters.js.coffee
+++ b/app/assets/javascripts/customFilters.js.coffee
@@ -82,9 +82,6 @@ angular.module('customFilters', [])
     s = ['th', 'st', 'nd', 'rd']
     v = n % 100
     (s[(v - 20) % 10] or s[v] or s[0])
-.filter 'divideAndRoundDown', ->
-  (input, divider) ->
-    Math.floor(input / divider)
 .filter 'listify', ->
   # will turn an array of items into a comma separated list, ending with "and" e.g.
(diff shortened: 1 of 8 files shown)


james-wills-sf:
QA'd:

- Listings look good on mobile
- Chart looks good in tagalog on mobile
- Max income chart looks correct for dahlia gardens
- empty cell /year issue is fixed

Brought this up in slack but the only thing I noticed was that some sale listings have only one row in the max income table, but the AMI tables show up to 3 hh max

akegan:
@james-wills-sf the bug on sale listings should be fixed now.

james-wills-sf (commented):
@akegan sale listings look good

Code also looks good but honestly a lot of the math was complicated so I mostly relied on what it looked like in the review app

james-wills-sf (approved):

PR #1417, opened 2020-11-17
Add dry run functionality to reupload script [DAH-155]

james-wills-sf:
I forgot to commit a small piece of code in my last pr, this is what actually makes the dry run functionality work.

I also had to move print_results into its own method because RuboCop took issue with the number of lines in the block.

[DAH-155]

[DAH-155]: https://sfgovdt.jira.com/browse/DAH-155


akegan (approved):
looks good

PR #1419, opened 2020-11-17
Fix rake script after running in prod [DAH-155]

james-wills-sf:
Final fix for [DAH-155]

two small fixes:
- Somehow my method name change wasn't commited, change method name to "for_each_id_in_file"
- print more detailed error message when the salesforce request fails. This is just something I realized would be nice when running the one-off.

[DAH-155]: https://sfgovdt.jira.com/browse/DAH-155


akegan (approved):
looks good!

PR #1422, opened 2020-12-28
Do not show SRO box on 1335 Folsom DAH-683

akegan:
DAH-683

Since we already had a listing for this building that made some changes, just adding this ID impacts the result in 3 places:
- https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/assets/javascripts/listings/ListingEligibilityService.js.coffee#L59
- https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/assets/javascripts/listings/components/listing/property-hero.html.slim#L93
- https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim#L75-L77
This is simpler than adding code to each of those places saying that it's technically 1335 Folsom Unit 403. 

Would be nice to check based on the building instead, but that is a bigger fix


james-wills-sf (approved):
@akegan lgtm

PR #1423, opened 2020-12-28
Run e2e production jobs on hotfix branches

akegan:
No ticket here, just would be nice to run prod e2e tests on hotfix branches too since we don't generally create release branches for them


james-wills-sf (approved):
@akegan this lgtm, interesting that it seems like this branch ran e2e-production tests, even though I thought those regexes only matched if the branch name starts with `hotfix`. 

Do you think it's matching if hotfix is anywhere in the name? Fine if so, I'm just curious

akegan:
yeah that's what I expected too, which is why I opened up the other sample branch. not sure why it's matching on this branch

PR #1428, opened 2021-01-14
Update Dalli config timeouts [DAH-694]

james-wills-sf:
[DAH-694]

[Recommended Dalli config](https://devcenter.heroku.com/articles/memcachier#ruby) has more relaxed timeout settings, this updates ours to use the recommended config.

[DAH-694]: https://sfgovdt.jira.com/browse/DAH-694


akegan (approved):
Turned on caching for the review app and it looks like it's still working as expected. 

One minor comment about adding the info link in a comment but otherwise LGTM

PR #1429, opened 2021-01-15
Update config for DALP env vars DAH-668

akegan:
DAH-668

This PR:
1. Adds info to README about these env vars
2. Makes the env vars not required on Heroku so we won't get an error on deploy is they're not present.

I've confirmed locally that if they aren't provided we get the same text that we have now
image

Prevents errors like:
image

And is best practice since the env vars are not required

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,11 @@ This script will:
 - Merge it with the latest main
 - Open a PR in a browser window
 
+## Environment variable configurations
+### DALP Advertising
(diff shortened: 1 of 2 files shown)


james-wills-sf (approved):
@akegan lgtm, tested with all 4 combinations of env vars (DALP_PROGRAM_INFO: null/not null, ADVERTISE_DALP: true/false) and everything looks good

PR #1436, opened 2021-02-22
Initial bloom/angular monorepo rewrite skeleton [DAH-691][DAH-727]

james-wills-sf:
Feasibility testing: DAH-691
Rewrite starting point: DAH-727

Create a starting point for the bloom rewrite code.


james-wills-sf:
@akegan it looks like circleCI is still having issues, but this change is ready for a qa/first pass code review.

app/controllers/home_controller.rb
```diff
@@ -1,9 +1,6 @@
-# Handles static pages
+# Controller for the DAHLIA homepage.
 class HomeController < ApplicationController
-  def index
-  end
-
-  def plus_housing
```
akegan, line 6: I tested this to see if it causes a problem, but it looks like we still redirect as expected if you go to /mohcd-plus-housing

app/views/layouts/application-react.html.slim
```diff
@@ -0,0 +1,41 @@
...
+      noscript
+        iframe height="0" src="//www.googletagmanager.com/ns.html?id=#{ENV['GOOGLE_TAG_MANAGER_KEY']}" style="display:none;visibility:hidden" width="0"
+    #google_translate_element
+
+    #root
+      == yield
+
+    / MAINTENANCE_LISTINGS can be set to a comma-separated list of listing Ids
+    / to force those listings to display a maintenance page, e.g.
+    / ENV['MAINTENANCE_LISTINGS'] = 'a0W4B0000050b5gUAA,a0W4B0000050b5gUBB'
```
akegan, line 35: never new that was an option! is that something that we have on angular too?

bookmarklet.js
```diff
@@ -1,32 +0,0 @@
-// This code is used to create the tooltip bookmarklet to show the mapping of
-// salesforce fields => fields displayed on the listing page
-// http://ted.mielczarek.org/code/mozilla/bookmarklet.html
-
-addTooltipsToChildren = function(children) { angular.forEach(children, function(e) {
```
akegan, line 5: weird. no idea what this is. I guess some firefox extension?

config/initializers/new_framework_defaults.rb
```diff
@@ -17,6 +15,3 @@
 # Make Ruby 2.4 preserve the timezone of the receiver when calling `to_time`.
 # Previous versions had false.
 ActiveSupport.to_time_preserves_timezone = false
-
-# Do not halt callback chains when a callback returns false. Previous versions had true.
-ActiveSupport.halt_callback_chains_on_return_false = true
```
akegan, line 22: was this change and the one above necessary? both of the settings that were here seem reasonable?

config/webpack/development.js
```diff
@@ -0,0 +1,5 @@
+process.env.NODE_ENV = process.env.NODE_ENV || 'development'
```
akegan, line 1: at some point I might ask what the actual differences are between environments

config/webpacker.yml
```diff
@@ -0,0 +1,95 @@
...
+    - .png
+    - .gif
+    - .tiff
+    - .ico
+    - .svg
+    - .eot
+    - .otf
+    - .ttf
+    - .woff
+    - .woff2
```
akegan, line 33: not a big deal, but wondering where this list came from. I have never heard of .woff files!

spec/controllers/overrides/registrations_controller_spec.rb
```diff
@@ -47,34 +55,38 @@
...
 
   describe '#update' do
-    # We no longer override the update method, but we want to confirm that email
-    # change confirmation emails are being sent.
     let(:user_update_params) do
       {
         user: {
-          id: 1,
-          email: 'jane2@doe.com',
+          email: 'jack2@doe.com',
```
akegan, line 61: wondering what the id is for jack2doe and how the update finds the id?/user?

spec/e2e/pages/short-form/you-name.coffee
```diff
@@ -2,6 +2,7 @@ AngularPage = require('../angular-page').AngularPage
 
 class Name extends AngularPage
   constructor: ->
+    super()
```
akegan, line 5: what does the super() do in these?

spec/javascripts/fixtures/json/listings-api-ami.json
```diff
@@ -1,187 +1 @@
...
....0,"numOfHousehold":4.0,"chartType":"HCD/TCAC","amount":61500.0},{"year":2016.0,"percent":50.0,"numOfHousehold":5.0,"chartType":"HCD/TCAC","amount":66450.0},{"year":2016.0,"percent":50.0,"numOfHousehold":6.0,"chartType":"HCD/TCAC","amount":71350.0},{"year":2016.0,"percent":50.0,"numOfHousehold":7.0,"chartType":"HCD/TCAC","amount":76300.0},{"year":2016.0,"percent":50.0,"numOfHousehold":8.0,"chartType":"HCD/TCAC","amount":81200.0}]},{"percent":"60","values":[{"year":2016.0,"percent":60.0,"numOfHousehold":1.0,"chartType":"Non-HERA","amount":51660.0},{"year":2016.0,"percent":60.0,"numOfHousehold":2.0,"chartType":"Non-HERA","amount":59040.0},{"year":2016.0,"percent":60.0,"numOfHousehold":3.0,"chartType":"Non-HERA","amount":66420.0},{"year":2016.0,"percent":60.0,"numOfHousehold":4.0,"chartType":"Non-HERA","amount":73800.0},{"year":2016.0,"percent":60.0,"numOfHousehold":5.0,"chartType":"Non-HERA","amount":73800.0},{"year":2016.0,"percent":60.0,"numOfHousehold":6.0,"chartType":"Non-HERA","amount":79740.0},{"year":2016.0,"percent":60.0,"numOfHousehold":7.0,"chartType":"Non-HERA","amount":85620.0},{"year":2016.0,"percent":60.0,"numOfHousehold":8.0,"chartType":"Non-HERA","amount":91560.0}]}]}
```
akegan, line 1: did anything change in here? wondering why this was reformatted

akegan (changes requested):
It looks like the package size thing is OK. Same size as on prod unless you load the actual react page
From incognito load of the angular home page
image

vs for prod
image
seems about the same/slightly less than prod actually. 

Here is the react homepage 
image



Only bug that I found that is a blocker is that if you set the ENV var to false so the default home page is angular, the link back to the home page still doesn't respect the language you're on. So if you're on https://dahlia-web-full-pr-1436.herokuapp.com/es/listings/for-sale and then click the logo, you get sent to the english home page

config/initializers/new_framework_defaults.rb
software-project: @akegan both of those options were removed from Rails. The first one actually didnt have any effect since Rails 5.1. The second one was replaced by explicitly calling `throw :abort` in Rails controller.

bookmarklet.js
software-project: I even wonder what that was even supposed to do. It wasnt referenced anywhere, even on the initial commit that added it.

akegan: strange! good to get rid of it then

config/initializers/new_framework_defaults.rb
akegan: thanks!

config/webpack/development.js
software-project: At this point, no difference at all. But I suppose for production we will want to compress and uglify css as it's 3mb in development ;)

config/webpacker.yml
software-project: This is autogenerated file when you install webpacker. And woff is a font format for dog lovers ;)

spec/controllers/overrides/registrations_controller_spec.rb
james-wills-sf: if you look at the [update method on the base class](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/registrations_controller.rb#L69) it isn't querying by id, it's applying update changes to the existing @resource, which we assigned to the user in the `before` block.

spec/e2e/pages/short-form/you-name.coffee
james-wills-sf: I just tried to remove it and our e2e tests failed, I think it's because we've updated our ES version and now it's a requirement: https://github.com/jashkenas/coffeescript/issues/4547

spec/javascripts/fixtures/json/listings-api-ami.json
james-wills-sf: i don't think any of this should have changed. reverted it

config/webpacker.yml
akegan: 😆  thanks!

james-wills-sf:
@akegan good catch on the language thing. I think it's going to take a little longer than this afternoon to fix. It's probably very related to DAH-739, since if we can handle all routing in one place, that's also where we can handle routing the language url path

akegan (approved):
After talking to James, I am OK with merging this as is and following up quickly w/ the language-missing-in-redirect-issue

james-wills-sf:
@akegan I just updated this with better behavior for translations, and I'll make a ticket about the 404 issue. The fix is slightly hacky, especially in routes.rb, but I think on the angular side it's actually a better system to do the redirects in stateChangeStart, so we can catch it all in one place.

I'm not going to rush and merge this tonight, maybe if you have time you can look at the new commit and re-qa on monday and if @software-project feels it will speed him up he can merge? Otherwise I'll merge first thing tuesday.

akegan:
OK QA looks good to me so up to @software-project if you want to merge before tuesday morning 😄

software-project (approved):
Looks good. Really nice job @james-wills-sf on the details.

PR #1439, opened 2021-03-11
[Webapp] Add heroku stack to app.json [DAH-675]

james-wills-sf:
Webapp: DAH-675
Partners: DAH-665

Following these instructions: https://devcenter.heroku.com/articles/heroku-20-stack

We need to add the new heroku stack version to make sure that review apps are on the correct version. Deploying this code won't upgrade existing apps, though, we have to upgrade them manually.

Partners PR is here: https://github.com/SFDigitalServices/sf-dahlia-lap/pull/493

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,6 +1,7 @@
 {
   "name":"sf-dahlia-web",
   "description":"",
+  "stack": "heroku-20",
   "scripts":{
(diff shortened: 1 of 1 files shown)


software-project:
@james-wills-sf shouldn't ruby version be upgraded to `2.6.x`? `2.5.x` is not supported for stack-20.

PR #1444, opened 2021-03-30
[hotfix][Webapp] Bump rails to 5.2.5 [DAH-751]

james-wills-sf:
Webapp ticket: [DAH-751]
Partners ticket: [DAH-752]

Webapp currently relies on mimemagic 3.5, which has been removed from the gem repository for licensing reasons. Rails 5.2.5 removes the mimemagic dependency.

See the [rails release](https://weblog.rubyonrails.org/2021/3/26/marcel-upgrade-releases/) and these issues [new mimemagic version released under MIT](https://github.com/rails/rails/issues/41757), [dependency on mimemagic no longer valid](https://github.com/rails/rails/issues/41750) for more info

[DAH-751]: https://sfgovdt.jira.com/browse/DAH-751
[DAH-752]: https://sfgovdt.jira.com/browse/DAH-752


james-wills-sf:
@akegan I think this is ready for review. I just did a quick test to submit an application and upload a proof and the proof shows up on salesforce correctly.

I also had to re-fix that same test. I tried using the new [assert_enqued_email_with](https://api.rubyonrails.org/v6.0.2.1/classes/ActionMailer/TestHelper.html#method-i-assert_enqueued_email_with) method but couldn't get it to work and figured it wasn't worth my time.

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 5.2 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
```
akegan, line 5: have you made the flips yet? or do we have to do that after?

akegan (approved):
QA looked good to me. I tested:

- Account creation and confirm email
- change email and confirm
- forgot password
- change something about acct and get an email about it
- submit an application, get a confirmation email
- that the file I uploaded for preference proof made it into salesforce

Just have the one question about the framework_defaults config file then should be ready to go?

config/initializers/new_framework_defaults_5_2.rb
james-wills-sf: @akegan I did them one by one (ish) locally so they're all already flipped except for `Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true` which should only be flipped once you're stable on 5.2, so I'm not in any rush to turn that on in a hotfix.

I did update the comment to remove the "Once upgraded flip defaults one by one" line since it's confusing to leave in after we're already upgraded and the flags are flipped

akegan: ok sounds reasonable

PR #1446, opened 2021-04-02
Add Employment/Permanent Disability preference for 588 mission DAH-744

akegan:
DAH-744

## Things to do

- [x] Get the look and content right
- [x] Individual preference field required for 588 mission
- [x] Allow bullet points for the preference description
- [x] Pull read more url from Salesforce
- [x] Save individual preference value to salesforce
- [x] Load saved values for saved draft
- [x] Make sure that the application summary looks correct

## Notes for QA
- Verify that the Employment/Permanent disability page matches the mock-up and has the right content
- Test submitting applications claiming the 588 mission preference with all of the different individual preference options and check salesforce to make sure that the Application Preference objects were created with the household member and individual preference option as expected. 
- Sign into an account, and fill out a 588 mission application through the custom preference page. Save the application as a draft. Continue the application and verify that values were saved and show up as expected. 
- Verify that the preference summary at the end of the application and in your account after you submitted the application looks ok
- Verify that COP and DTHP and L/W preferences still all work as expected. 
- Make sure that the listing details page looks OK
- Check that values for the custom pref are unset when you check and uncheck the checkbox.


james-wills-sf:
pulled in the pattern library changes, looks good.

- 2021-04-02 at 3 42 54 PM

app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee
```diff
@@ -374,6 +374,7 @@ ShortFormApplicationService = (
     else
       Service.preferences["#{prefType}_household_member"] = null
       Service.preferences["#{prefType}_proofOption"] = null
+      Service.preferences["#{prefType}_preference"] = null
```
akegan, line 377: this is for resetting the value on check + uncheck. I don't think this will hurt live/work or rent burdened/assistive housing but we should probably check to make sure that the behavior for those doesn't change

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
```diff
@@ -468,6 +470,7 @@ ShortFormDataService = (ListingDataService, ListingConstantsService, ListingPref
         unless prefKey
           # must be a customPreference... just identify by ID much like on e7b-custom-preferences
           prefKey = listingPref.listingPreferenceID
+          isCustom = true
```
akegan, line 473: this feels a bit hacky, but keeps adding the inidividualPreference piece limited to custom preferences.

app/assets/javascripts/short-form/components/preferenceCheckbox.js.coffee
```diff
@@ -9,16 +9,20 @@ angular.module('dahlia.components')
...
   controller: ['ListingDataService', 'ListingPreferenceService', (ListingDataService, ListingPreferenceService) ->
     ctrl = @
 
-    @showDescription = false
     listingPreference = ListingPreferenceService.getPreference(@preference, ListingDataService.listing)
     if listingPreference
       @moreInfoLink = listingPreference.readMoreUrl
+    else
+      # If you can't find the moreInfo link, for example, with a custom preference
+      # like 588 Mission, look for a passed readMoreUrl value
```
akegan, line 24: I tried to add 588 mission to comments in wonky places that we had to update but could be a bit more consistent

app/assets/javascripts/short-form/components/preferenceComponent.js.coffee
```diff
@@ -49,6 +50,20 @@ angular.module('dahlia.components')
       @onChange = =>
         @resetPreference()
 
+      # For 588 Mission preference, we need to override some defaults.
+      @is588 = =>
```
akegan, line 54: I might change this to be `isEmploymentDisability` so that it's more tied to the preference than the listing

akegan (commented):
This is all working now 🎉. Just need to clean up a bit and add tests but ready to start thinking about how to QA

akegan:
@james-wills-sf I updated the review instructions in the PR body and in the ticket. I didn't write up every step but tried to cover all the things that need to be covered. It's definitely a bit of work to QA this one!

james-wills-sf:
@akegan okay I QA'd this in a few different ways, all looks good:
- content matches what we have so far
- saving the new preference and other preferences works when saving as a draft or not
- changes show up correctly in salesforce including individual preference field

One question I had, this is what it looks like when you say you want the custom preference (In this case all of the other preferences are also claimed):
- 2021-04-07 at 1 26 32 PM
Do you know what the variances in the pre- and post-lottery validation columns mean?

akegan:
The validation fields should match what we have for something like live/work. COP and DTHP are validated pre-lottery by MOHCD staff so that's what that field looks like. Would be good to click into that and verify that the application member and the individual preference field are populated as expected.

james-wills-sf:
@akegan yeah i clicked into each, all look good

app/assets/javascripts/short-form/components/preference-component.html.slim
```diff
@@ -1,14 +1,73 @@
...
+    custom-description="{{$ctrl.customDescription}}"
+    short-description="{{$ctrl.translatedShortDescription}}"
+    read-more-url="{{::$ctrl.readMoreUrl}}"
+    on-change="$ctrl.onChange()"
+    required="$ctrl.required()"
+  )
+  .form-group.input-inset.margin-bottom-none.padding-bottom-none ng-if="$ctrl.application.preferences[$ctrl.preference]"
     / member selector
-    preference-member-selector(application="$ctrl.application" preference="{{$ctrl.preference}}" eligible-members="$ctrl.eligibleMembers" label="{{$ctrl.memberSelectorLabel}}")
+    preference-member-selector(
```
james-wills-sf, line 16: this was just reformatted without edits, right?

app/assets/javascripts/short-form/components/preferenceCheckbox.js.coffee
```diff
@@ -9,16 +9,20 @@ angular.module('dahlia.components')
...
   templateUrl: 'short-form/components/preference-checkbox.html'
 
   controller: ['ListingDataService', 'ListingPreferenceService', (ListingDataService, ListingPreferenceService) ->
     ctrl = @
 
-    @showDescription = false
     listingPreference = ListingPreferenceService.getPreference(@preference, ListingDataService.listing)
     if listingPreference
       @moreInfoLink = listingPreference.readMoreUrl
+    else
```
james-wills-sf, line 22: nit: On a component API level I kind of feel like this should be flipped, like if you pass an explicit read more link it should override the salesforce version

like 
```
if !!@readMoreUrl
  @moreInfoLink = @readMoreUrl
else if listingPreference
  @moreInfoLink = listingPreference.readMoreUrl
```

just nitpicking though since I know this is hopefully not a component we'll have to use much more.

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
```diff
@@ -447,7 +449,7 @@ ShortFormDataService = (ListingDataService, ListingConstantsService, ListingPref
       return unless listingPref
 
       member = _.find(allHousehold, {appMemberId: shortFormPref.appMemberID})
-
+      isCustom = false
```
james-wills-sf, line 452: I know it's less efficient but it might be more readable to to

```
isCustom = !Object.values(ListingDataService.preferenceMap).includes(listingPref.preferenceName)
```

and then you can get rid of line 473

james-wills-sf (approved):
@akegan the code lgtm! I have nit comments but nothing blocking

app/assets/javascripts/short-form/components/preference-component.html.slim
```diff
@@ -1,14 +1,73 @@
...
+      application="$ctrl.application"
+      preference="{{::$ctrl.preference}}"
```
akegan, line 18: looks like I also added :: to this line. I'm going to remove it because there's a possibility that it could cause issues

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
akegan: I kind of like having the whether or not it's a custom preference just be in one place here because the if/else statements are already really confusing 😕. I think the way I put it is actually less confusing to me?? Idk all of the if elses are just pretty crazy

james-wills-sf: okay, fine to keep it then

app/assets/javascripts/short-form/components/preference-component.html.slim
akegan: answered [below](https://github.com/SFDigitalServices/sf-dahlia-web/pull/1446#discussion_r609104833) then updated to remove the ::'s. Should probably do another quick QA runthrough

app/assets/javascripts/short-form/components/preferenceCheckbox.js.coffee
akegan: updated, should probably QA this again

akegan: actually, i wonder if I still need this w/ the change I made back to use the listing preference ID instead of calling the fields five88. I'll investigate now

akegan: OK I was able to use getPreferenceById() to simplify this a bit. I did a bit of QA but would appreciate another set of eyes too!

akegan:
I also added a wait to the e2e tests to see if I could get one that was being flaky to be less flaky

james-wills-sf:
@akegan re-QA lgtm

PR #1449, opened 2021-04-13
Initial auth setup

software-project:
[DAH-728]

[DAH-728]: https://sfgovdt.jira.com/browse/DAH-728


app/javascript/pages/SignIn.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { SignInForm } from "../authentication/SignInForm"
+import { LoggedInUserIdleTimeout } from "../authentication/timeout"
+import { UserProvider } from "../authentication/UserContext"
+import FormsLayout from "../layouts/Forms"
+
+const SignIn = () => {
+  return (
+    <FormsLayout>
+      <UserProvider>
```
james-wills-sf, line 11: i think we should have a component wrapper function that applies all of the application-y stuff on top of each page.

```
function withAppSetup<T>(Component: React.ComponentType<T>) {
  return (props: T) => (
      <UserProvider>
        <LoggedInUserIdleTimeout />
        <Component {...props} />
      </UserProvider>
}
```

and then in SignIn / Homepage you can do 

`export withAppSetup(SignIn)`

config/webpack/environment.js
```diff
@@ -1,4 +1,18 @@
 const { environment } = require("@rails/webpacker")
+const dotenv = require("dotenv")
+const webpack = require("webpack")
+
+const dotenvFiles = [".env"]
+dotenvFiles.forEach((dotenvFile) => {
```
james-wills-sf, line 6: What do these changes do?

config/routes.rb
```diff
@@ -84,6 +84,8 @@ def matches?(_)
   get '/zh', to: 'home#index'
   get '/tl', to: 'home#index'
 
+  get '/sign-in', to: 'auth#sign_in'
```
james-wills-sf, line 87: so you'll have to add /en/es/zh/tl variations here. If you have ideas on how to simplify it as a regex let me know!

james-wills-sf (commented):
@software-project did a first pass (code only, I haven't QA'd yet). Just have a few mostly nit comments so far

app/javascript/authentication/timeout.tsx
```diff
@@ -0,0 +1,134 @@
...
+    promptAction,
+    promptText,
+    redirectPath,
+    alertMessage,
+    alertType = "alert",
+    onTimeout,
+  } = props
+
+  // 5 minutes
+  const idleTimeout = 5 * 60 * 1000
```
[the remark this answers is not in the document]
software-project: I can't find a setting for it. It seems like it's passed from devise and out there we have 30 minutes. Should I set it to 15 for now? Or 30?

config/webpack/environment.js
software-project: They are loading env variables. It seems like they are not loaded automatically.

config/routes.rb
software-project: can't we add url param? like `/?lang=en`

james-wills-sf: No, we want it to be interoperable with angular, and angular adds the language prefix to the path.

It's also the way bloom does it, so the workarounds wouldn't be worth it.

app/javascript/authentication/timeout.tsx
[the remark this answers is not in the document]
james-wills-sf: I just tested on dahlia prod and the timeout is <10 minutes there. I guess we should keep it at 5 for now.

app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,4 @@
+# Controller for the DAHLIA homepage.
+class AuthController < HomeController
```
james-wills-sf, line 2: I don't think this should extend HomeController, instead it should extend `ApplicationController` and define its own `useReactApp` override like

```
  def use_react_app
    ENV['SIGN_IN_PAGE_REACT'].to_s.casecmp('true').zero?
  end
```

james-wills-sf: Also, we want a separate feature flag for this page since the homepage will be ready for launch before this page is.

app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,4 @@
+# Controller for the DAHLIA homepage.
```
james-wills-sf, line 1: Nit: this comment isn't accurate anymore

app/javascript/pages/HomePage.tsx
```diff
@@ -11,13 +11,14 @@ import {
...
 
+import { withAppSetup } from "../layouts/app_settings"
 import Layout from "../layouts/Layout"
 
 export interface HomePageProps {
   listings?: Listing[]
 }
 
-const HomePage = (props: HomePageProps) => {
+const component = (props: HomePageProps) => {
```
james-wills-sf, line 21: I think we want to keep this named `HomePage`. Question about the `export { HomePage as default, HomePage }` pattern: Why not just prefer `export default HomePage`? At my last job we would prefer a default export when a file only exported one thing, but I could just be missing why it's helpful to also provide a named export

james-wills-sf (commented):
@software-project I added a few more comments, but I'm a little confused how to QA this. Can you add review instructions to the ticket to test this? What I'd like to check is the following:

- sign in on angular and switching to a react page keeps you authenticated
- sign in on react and switching to angular keeps you signed in

app/javascript/pages/HomePage.tsx
software-project: That is my preference as well, but how would you use `withAppSetup` in that case without causing an error and keeping with our linting rules? Cause I couldn't make that happen.

james-wills-sf: @software-project let me try it out now. At my last job we used this pattern a lot and would just do

```
export default withAppSetup(HomePage)
```

but maybe that doesn't work with our linting rules

james-wills-sf: Yeah I'm not getting any linting errors when I do:

```
const HomePage = (props: HomePageProps) => {
... component rendering ...
}

export default withAppSetup(HomePage)
```

app/controllers/auth_controller.rb
software-project: Sure, that make sense. But actually `use_react_app` should be a `protected` method. `public` methods are reserved for actions as of Rails convention. Will make amendments.

app/javascript/pages/HomePage.tsx
software-project: Ok, weird. Maybe there was a bug with my vs code linting, I'm like 90% sure I tried that.

app/controllers/auth_controller.rb
james-wills-sf: sounds good!

app/controllers/auth_controller.rb
```diff
@@ -1,4 +1,10 @@
-# Controller for the DAHLIA homepage.
-class AuthController < HomeController
+# Controller for the authentication pages.
+class AuthController < ApplicationController
   def sign_in; end
+
+  protected
```
james-wills-sf, line 5: do you need to add `protected` to the HomeController, too?

software-project: yes, method types should be consistent, good catch. Will add it.

james-wills-sf (approved):
LGTM, I'm testing the timeout functionality now

james-wills-sf (changes requested):
Actually, I've been QAing this and ran into a weird thing.

1. go to https://dahlia-web-full-pr-1449.herokuapp.com/sign-in?react=true
2. sign in, it will redirect you to https://dahlia-web-full-pr-1449.herokuapp.com/?react=true
3. change the url to react=false
4. Notice that the navbar says "sign in", not "My account" like it should

This only happens to me in the review app, not locally. I'm not sure why, but it seems like on the review app session storage is cleared out when you go from react to angular

james-wills-sf:
QA on the new changes working as expected. merging this now.

james-wills-sf (approved):

PR #1450, opened 2021-04-13
Support localization in rewrite [DAH-749]

james-wills-sf:
[DAH-749]

This PR does a few big things:
- Adds support for localized links and switching languages
  - Adds new i18n phrase files (localization/react/en/es/zh/tl.json) and uses those to populate react strings
  - Updates the navbar to (mostly) work
  - Swapped out translation key placeholders with the real translation keys
- Switched libraries from @bloom-housing/ui-components to @sf-digital-services/ui-components
- Refactored some sign-in stuff (sorry, I missed these during code review)
  - Instead of using Logged in or logged out timeout components, you can just use `<IdleTimeout />` which handles both cases
  - Some renaming and deleting of unused files and code
-  Updates to tests
  - e2e tests added to check translation
  - simple sign in e2e tests added
  - setup-tests (which is run before each unit test) modifications
    - now loads english translations
    - runs cleanup after each test. I'm not totally sure what this does but we had it in our homepage test, might as well do it before each test unless we notice any slowdown
    - Fails on any console warning
      - I added this to catch string not translated errors. 

[DAH-749]: https://sfgovdt.jira.com/browse/DAH-749


akegan:
@james-wills-sf can you convert to draft if it's not ready for review yet? Github keeps sending me passive aggressive DMs about it every morning 😂

Gruntfile.js
```diff
@@ -91,7 +93,7 @@ module.exports = function (grunt) {
               replacement: "//fonts.googleapis.com",
             },
             {
-              match: /\.\.\/images\/([a-zA-Z0-9\-_@]*\.(png|jpg|svg))/g,
+              match: /\.\.\/images\/([\w@-]*\.(png|jpg|svg))/g,
```
james-wills-sf, line 96: Sorry, I added a linter that simplifies regexes ad part of the unicorn/recommended pack. I think the idea of it is nice but I don't necessarily find this any easier to read, what do you think?

app/javascript/layouts/application.tsx
```diff
@@ -1,90 +0,0 @@
-import React, { useContext } from "react"
```
james-wills-sf, line 1: This file was actually unused, we use Layout.tsx instead

app/javascript/util/languageUtil.ts
```diff
@@ -0,0 +1,115 @@
...
+  const config = LANGUAGE_CONFIGS[prefix]
+
+  if (!config.isDefault) {
+    addTranslation(await config.load())
+  }
+}
+
+export const toLanguagePrefix = (routePrefix: string | undefined): LanguagePrefix => {
+  switch (routePrefix) {
+    case LanguagePrefix.Spanish:
```
james-wills-sf, line 71: I know this looks a little weird but it's the only way to get a string enum value

app/javascript/.jest/setup-tests.ts
```diff
@@ -2,11 +2,15 @@ import "@testing-library/jest-dom/extend-expect"
 
 import { configure } from "enzyme"
 import Adapter from "enzyme-adapter-react-16"
-import { addTranslation } from "@bloom-housing/ui-components"
-import * as general from "../page_content/locale_overrides/general.json"
+import { addTranslation } from "@sf-digital-services/ui-components"
+import enPhrases from "../../assets/json/translations/react/en.json"
```
software-project, line 6: Seems like those two are not used.

software-project:
@james-wills-sf I tested it locally and it looks really nice. My only wonder is why then I'm on default language sign-in link is `/en/sign-in` and others are without language prefix. Is it what we expect?

james-wills-sf:
@software-project good catch. No, it should just be /sign-in. I'll fix that bug

james-wills-sf:
@software-project looking at it again, the en/sign-in thing is coming from bloom. We're going to have to re-make or update `<UserNav />` to get it working, we can assume it's not working in this PR

Gruntfile.js
akegan: I don't think it's easier to read but if it still works that's OK. I am just mildly nervous because in the past when I was experimenting with this file the regexes didn't seem to behave as I expected them to.

james-wills-sf: @akegan that's a really good point, that regex parsers all have particular rules that the linting rule might not account for. That's enough of a reason for me to turn this off!

I'll change the linting rule to be a warning (or a "suggestion" if something less than a warning is possible) so we can learn from it if we write our own regexes but it doesn't automatically change our code.

akegan:
@james-wills-sf did you want some extra QA on [DAH-728](https://sfgovdt.jira.com/browse/DAH-728) auth stuff in this PR too? Are the review instructions for that still all covered by the ones in DAH-728?

james-wills-sf:
@akegan I actually think that auth QA makes more sense after we complete the My account dropdown in the navbar (DAH-759).

I added auth specific testing instructions to that ticket.

Gruntfile.js
james-wills-sf: @akegan removed and reverted this change

akegan:
Switched to non-react homepage and got this behavior where spanish and filipino are switching to google translated chinese automatically. It's not happening on full so I'm wondering if something is off with how we're communicating the language code to google translate?
https://user-images.githubusercontent.com/11825994/116919498-35f59400-ac06-11eb-9683-61bf71f640cb.mov

akegan:
I think this looks good and we can merge it in and try to fix the google translate issue as a release bug?

akegan (approved):
Turns out the translation bug is on main too, so we broke it out into a separate ticket([DAH-787](https://sfgovdt.jira.com/browse/DAH-787?atlOrigin=eyJpIjoiYmQxN2FlODQ3YTg0NDdkMThlNjM3YzBiNDc4YmRkOGMiLCJwIjoiaiJ9)). 

LGTM as is now

PR #1451, opened 2021-04-14
db schema + gem update

aekong:
* updates puma to 4.3.6
* updates db schema generated by rake

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ gem 'coffee-rails'
 gem 'sprockets_uglifier_with_source_maps'
 # See https://github.com/rails/execjs#readme for more supported runtimes
 # gem 'therubyracer',  platforms: :ruby
-gem 'puma', '~> 4.3.5'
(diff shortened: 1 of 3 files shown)


james-wills-sf (approved):
@aekong this lgtm.

I did a little investigating on what differences the db_schema changes actually make, and it all seems like it won't change anything and were actually caused by our Rails 5.2 upgrade, not changes to the migrations files:

### Change 1: adding `id: :serial` columns
Before, IDs were added by default and would be integer/int4/serial (all different ways of saying the same thing)

After 5.1, IDs default to bigserial/biginteger/int8, so we need to explcitly specify they are _not_ bigintegers. See [this post for more explanation](https://stackoverflow.com/questions/54598531/what-determines-if-rails-includes-id-serial-in-a-table-definition).

## Change 2: removing `using: :btree`
This is a change from upgrading to Rails 5.2 (see [this](https://github.com/rails/rails/pull/27981)). Sounds like it's another no-op change.

PR #1452, opened 2021-04-15
Pull chinese translations from LionBridge DAH-758

akegan:
This is the first part of DAH-758

Once we finalize the content for the 588 preference we'll kick off another job to get the Spanish and Filipino translations

### DIFF
--- a/app/assets/json/translations/locale-en.json
+++ b/app/assets/json/translations/locale-en.json
@@ -390,13 +390,13 @@
             "title": "There are lottery preferences for people who already have special certificates."
         },
         "e7b_custom_preferences": {
-            "title": "This listing has other special lottery preferences",
             "employment_disability": {
+                "cannot_work_disability": "Cannot work due to disability",
                 "description": "You can get this preference if one adult in the household: <ul><li>Works at a public higher education institution located in San Francisco</li><li>Works at a public healthcare institution located in San Francisco</li><li>Is unable to work because of a disability</li></ul>",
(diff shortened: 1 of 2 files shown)


james-wills-sf:
- 2021-04-15 at 12 34 15 PM

@akegan I noticed that the spacing was a little weird on the price ranges, it looks like google translate is trying to re-translate the currency strings? I thought google translate only tries to translate strings without translation values

james-wills-sf (approved):
If the double-translation thing isn't something we have control over, this lgtm

akegan:
we probably have to add a no-translate tag. I'll see if that helps

akegan:
Google translate is honestly a menace. 

I ended up adding skiptranslate to the whole table for rental and sales directory and listing. I'm thinking that we should see if Sharon would be up for taking a look to make sure that it all looks sane with these changes since they're really the most important part of the listing

james-wills-sf:
@akegan I don't think skip translate is the answer, because now tagalog looks like this:

- 2021-04-16 at 8 53 03 AM

Probably best to just leave it as it was before my comment, sorry!

akegan:
No I think the skip translate is still better for now. Looks like we have most of the things translated for filipino and we'll have the rest soon. Google translate on the dollar amounts has also caused weird currency issues in the past too.

james-wills-sf (approved):
@akegan not sure if you were waiting on my review or content qa but this lgtm

PR #1453, opened 2021-04-20
Remove translation notice from header [DAH-762]

software-project:
Revolves [DAH-762](https://sfgovdt.jira.com/browse/DAH-762)


james-wills-sf (approved):
@software-project lgtm, I'm going to update the translations to remove those keys before merging (not sure if you have access to phrase so I'll just do it)

PR #1454, opened 2021-04-20
Fix gruntfile to check application-* files

james-wills-sf:
We split `application.html.slim` into `application-angular.html.slim` and `application-react.html.slim` but forgot to update the gruntfile, so the t.skip_to_main_content key was missed.

I did a search for `application.html.slim` to see if I needed to update it anywhere else, it wasn't used anywhere but one comment that I went ahead and replaced.


james-wills-sf:
I tested this by running grunt phrasePull after adding strings to es/zh/tl on phrase.com and verified that there were no changes to our translation files

akegan (approved):
LGTM thanks for the quick fix. I just kicked off e2e-legacy tests again. hopefully they pass this time 🤞

PR #1455, opened 2021-04-23
Redirect 404 to react homepage [DAH-756]

software-project:
Resolves react redirect issue: [DAH-756](https://sfgovdt.jira.com/browse/DAH-756)


james-wills-sf (changes requested):
@software-project This solution doesn't work as-is, it redirects all valid (non react) urls to `/`.

For example, try going to https://dahlia-web-full-pr-1455.herokuapp.com/listings/for-rent, it redirects to the homepage.

I think the solution might be to modify the angular to redirect (via rails) to `/` if an unrecognized route is found

james-wills-sf:
@akegan can you help QA this? the original solution didn't work so I replaced it with something I _think_ works but you're better than be at catching edge cases

james-wills-sf (approved):

akegan:
not sure if this is a blocker but I just found this weirdness where if you give an invalid listing id it seems to flicker to the listing details page with no data before redirecting to the homepage?

https://user-images.githubusercontent.com/11825994/116633491-7905ed80-a90e-11eb-9b83-33833fcdeb67.mov

james-wills-sf:
Merging this as it doesn't cause the bug above and we created a separate ticket to fix it: DAH-786

PR #1456, opened 2021-04-26
Update chinese translations sharon DAH-758

akegan:
DAH-758.

[Doc with recommendations from Sharon](https://docs.google.com/document/d/1uEfM03HQEzJD_6JSSOVANR-Ud80zCeSrK78rA99B-RQ/edit)

## TODO
- [x] Make update for "people in household" once I get which version to use from Sharon
- [x] Exclude 1BR, etc from google translate for sale listings
- [ ] (possibly) make an update to AMI translation if we get an answer in a reasonable time frame


app/assets/javascripts/short-form/components/preferenceComponent.js.coffee
```diff
@@ -52,10 +54,13 @@ angular.module('dahlia.components')
 
       # For 588 Mission preference, we need to override some defaults.
       @isEmploymentDisability = =>
-        @title == 'Employment/Disability Preference'
+        @title == 'Employment or Disability Preference'
```
akegan, line 57: Added flagforI18n below so that `grunt translations` wouldn't delete the description.

I had to make this update here too to check that change

akegan:
This is ready for review now. I ended up implementing 4 of the recommendations from sharon (although one of them was just to remove the google translate for 1BR from Sales instead of fixing it for all of them)

I added a comment in the doc to ask her to take a look at some screenshots from this review app too in case I made any copy/paste errors

app/assets/javascripts/short-form/components/preferenceComponent.js.coffee
james-wills-sf: Should this be changed back before merging?

james-wills-sf (approved):
@akegan changes LGTM. Looks like 1 BR isn't translated as expected, but I'm not sure where in the code that happens, can you point me to it?

akegan:
@james-wills-sf turns out I hadn't pushed that commit so good catch! It's in 3e0892e

james-wills-sf:
@akegan got it. All good to merge then

PR #1457, opened 2021-04-28
Reduce size of tailwind DAH-726

aekong:
attempt to reduce the size of resources by:
- overriding bloom's tailwind purge config option to tree-shake unused css
- minifying css

the screen cap here is a comparison of the html dom saved as text files:

Screen Shot 2021-04-29 at 11 11 10 PM

the comparison here is from running the command
```sh
NODE_ENV=production ./bin/webpack --colors --progress
```

| filename | size before | size after |
| -------- | ------------| ----------|
| react_application-some-hash.js | 4.8M | 2.3M
| react_application-some-hash.js.gz | 717K | 545K


tailwind.config.js
```diff
@@ -5,4 +5,11 @@ const bloomTheme = cloneDeep(require("@sf-digital-services/ui-components/tailwin
 // For example:
 // bloomTheme.theme.colors.white = "#f0f0e9"
 
+// tailwind will automatically purge unused styles when `NODE_ENV` is set to `production`
+bloomTheme.purge = [
+  './src/**/*.tsx',
+  __dirname + '/app/assets/javascripts/**/*.html.slim',
```
james-wills-sf, line 11: I don't think we need this line, since anything under /app/assets/javascripts is angular.

tailwind.config.js
```diff
@@ -5,4 +5,11 @@ const bloomTheme = cloneDeep(require("@sf-digital-services/ui-components/tailwin
 // For example:
 // bloomTheme.theme.colors.white = "#f0f0e9"
 
+// tailwind will automatically purge unused styles when `NODE_ENV` is set to `production`
+bloomTheme.purge = [
+  './src/**/*.tsx',
```
james-wills-sf, line 10: I don't think this matches anything, looks like a default path that we can delete

postcss.config.js
```diff
@@ -1,13 +1,24 @@
...
+  require("postcss-preset-env")({
+    autoprefixer: {
+      flexbox: "no-2009"
+    },
+    stage: 3
+  })
+]
+
+if (process.env.NODE_ENV === 'production') {
+  plugins.push(cssnano)
```
james-wills-sf, line 17: Just a question: What does cssnano do and how is it different from the minimize option?

postcss.config.js
```diff
@@ -1,13 +1,24 @@
+const cssnano = require("cssnano")({preset: "default"})
```
james-wills-sf, line 1: tiny nit: I don't think these files are linted but can you add spaces around the `{}`?
```suggestion
const cssnano = require("cssnano")({ preset: "default" })
```

james-wills-sf (approved):
Just a few small comments and questions, but in general this LGTM

postcss.config.js
aekong: It's used in conjunction with the minimize option.  but i have to walk this back because of these potential issues: https://alexpeattie.com/blog/psa-dont-use-cssnano-with-postcss/ which links to this comment on the gh issue: https://github.com/cssnano/cssnano/issues/659#issuecomment-444025691.

also, doing it this way automatically considers the production environment, so we won't need to do an explicit check

postcss.config.js
aekong: i opted to use `cssnano-webpack-plugin` instead

james-wills-sf:
@aekong just checking to make sure I'm not blocking you on this, you were going to look more into the minification before this is ready to merge, right?

aekong:
> @aekong just checking to make sure I'm not blocking you on this, you were going to look more into the minification before this is ready to merge, right?

yup, the minifying isn't actually happening in production, trying to figure that out

aekong:
so i've been mostly unsuccessful trying to get webpacker to pick up on the minification plugins.  the primary task of this pr was to tree shake tailwind, which is confirmed working so far (see screenshots in pr description above).  @akegan @james-wills-sf thoughts on merging this in and saving minification for a separate pr? thanks

akegan:
that sounds totally fine to me!  Maybe you can just write up some of what you tried in the ticket for DAH-788?

akegan (approved):
Good to go for now! Lots of great improvements already

aekong:
yeah, i'll remove them 👍 thanks!

PR #1459, opened 2021-04-29
Fix translation error DAH-758

akegan:
DAH-758

I caught an issue where we sent a4_overview.subtitle over to lionbridge and they changed the translation from "Please provide your answers in english" to a literal translation of what we have in there of "Do not translate - Chinese only". This is a weird key where we are only showing the value in Chinese. I think we wanted to save the time of our translators with this message, but going forward we should just treat it like any other key to avoid this sort of issue again in the future. 

I'll follow-up with another PR that does that

## Review instructions
Look at [the diff](https://github.com/SFDigitalServices/sf-dahlia-web/blame/main/app/assets/json/translations/locale-zh.json) and verify that we're reverting to what we had before


james-wills-sf (approved):
@akegan looks good except there's now a newline in the content, is that expected? See the old string without a newline: https://github.com/SFDigitalServices/sf-dahlia-web/blame/e51e49b7265a5599bd7caf7feb16ac602001e348/app/assets/json/translations/locale-zh.json#L32

akegan:
Fixed, thanks for catching!

james-wills-sf (approved):
@akegan lgtm

PR #1461, opened 2021-05-03
Fix 588 mission missing summary DAH-744

akegan:
DAH-744

This fixes the issue where custom preferences were not showing in the application summary section because of the way we get listing data from an application object. 

Review instructions in the ticket


james-wills-sf (approved):
@akegan Just QA'd, LGTM

PR #1462, opened 2021-05-03
Add comments to point to RTR updates

james-wills-sf:
This is just a diff to help the person who implements the right to return preference know where to make edits. Helps with implementing DAH-790


akegan:
Closed in favor of #1475

PR #1465, opened 2021-05-06
Fix flashing on listing redirect [DAH-786]

james-wills-sf:
DAH-786

## Why was it flashing before?
When redirecting from the listing details page to the welcome page, it was showing an empty listing details page before the redirect finished.

A few reasons why this happened:
1) Redirects to rails take longer than angular state changes, so it used to be basically instant, now it's not.
1) we were calling e.preventDefault on the state change to stop the new view from loading (because we don't want to show the angular homepage necessarily either)
2) without the new view loading, the empty listing template starts showing because even though the listing load failed, "loading" is still complete according to angular

## what was the fix
Instead of telling angular router to not render any template (which inevitably falls back to render the old template), we tell angular to render a pure white, blank template on redirects.

## Why not just make the template a blank div instead of the hacky all white box
A blank div will still show the header and footer. This looked pretty bad to me but I can remove the hacky solution if we don't like it.

### DIFF
--- a/app/assets/javascripts/pages/templates/blank.html.slim
+++ b/app/assets/javascripts/pages/templates/blank.html.slim
@@ -1,0 +1,3 @@
+// Just a big white block that takes up the whole screen.
+// Should only be used for redirects.
+div.topfocus style="position:fixed;padding:0;margin:0;top:0;left:0;z-index:104;width: 100%;height: 100%;background:rgba(255,255,255);"
--- a/app/assets/javascripts/config/angularInitialize.js.coffee
+++ b/app/assets/javascripts/config/angularInitialize.js.coffee
@@ -47,20 +47,21 @@
         $state.go('dahlia.listing', {timeout: true, id: ShortFormApplicationService.listing.Id})
 
     $rootScope.$on '$stateChangeStart', (e, toState, toParams, fromState, fromParams) ->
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/pages/templates/blank.html.slim
```diff
@@ -0,0 +1,3 @@
+// Just a big white block that takes up the whole screen.
+// Should only be used for redirects.
+div.topfocus style="position:fixed;padding:0;margin:0;top:0;left:0;z-index:104;width: 100%;height: 100%;background:rgba(255,255,255);"
```
akegan, line 3: We have the same blank loading screen on page transitions on prod too. Is there a way to reuse the same template we use from that? I have no idea how that's actually setup.

akegan (approved):
LGTM. You can see the url update to "redirect-home" if you look closely but I think that's fine.

app/assets/javascripts/pages/templates/blank.html.slim
james-wills-sf: I don't think that's actually the same thing. In prod the loader between pages is actually translucent I think, only the first load is blank, and that's because the template hasn't loaded behind it yet.

If there is an actual blank template I've been unable to find it

PR #1474, opened 2021-05-19
Add react alert banner [DAH-763]

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAH-763

Notes:
I believe we need two follow up tickets:
- First for making smaller alert
- And a second for making notice optional for header


akegan:
@software-project for "And a second for making notice optional for header" do you mean making the blue banner optional? And for smaller alert, do you mean making the height match the height of the black and blue headers and adding the margins?

software-project:
@akegan yes, `notice` is blue banner. If we don't pass a value (string or html element) it will still show up, but as a thinner bar. 
For alert I will add margins now, I forgot about them. But right now I can't alter height as it's a part of ui-components.

akegan:
OK I think that we actually want to wait til the UI components alert banner gets margins and then just use those?

akegan:
Added a ticket DAH-814 for the 2 things you mentioned (linking to the bloom issue for the margins). I also added this to the agenda for the catch-up next Tuesday

app/assets/images/icons.svg
```diff
@@ -0,0 +1,242 @@
+<svg style="position: absolute; width: 0; height: 0; overflow: hidden;" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
akegan, line 1: are the icons something that we could import from the UI-components lib instead of copying the file over? Might be easier to keep it up-to-date that way? and visualize what icons we have access to?

app/javascript/authentication/token.ts
```diff
@@ -1,6 +1,7 @@
 const ACCESS_TOKEN_LOCAL_STORAGE_KEY = "auth_headers"
 
 const getStorage = () => {
+  console.log(process.env)
```
akegan, line 4: probably want to remove this
```suggestion
```

app/javascript/layouts/Layout.tsx
```diff
@@ -31,12 +33,31 @@ const Layout = (props: LayoutProps) => {
 
   const currentPath = window.location.pathname
 
+  let notice = null
+  console.log("banner", process.env.SHOW_RESEARCH_BANNER)
```
akegan, line 37:
```suggestion
```

config/webpack/environment.js
```diff
@@ -7,10 +7,18 @@ dotenvFiles.forEach((dotenvFile) => {
   dotenv.config({ path: dotenvFile, silent: true })
 })
 
+// Strings needs to be interpolated and single quoted
```
akegan, line 10: Interesting! does this cause issues if we have an apostrophe or anything in our env vars?

akegan (changes requested):
Couple of questions, also I'm not seeing the alert banner show up
image
even with the TOP_MESSAGE env var set

akegan:
@software-project is going to change the environment.js parsing to use JSON.stringify and link to documentation around what the options are for TOP_MESSAGE_TYPE

akegan:
@software-project I confirmed that you do have to do a redeploy and that the dyno restart isn't sufficient to see the changes reflected.

I also noticed that it looks like we're defaulting to the inverted version of notices? 
image
Should we expose a way to set inverted or not too?

app/javascript/layouts/Layout.tsx
```diff
@@ -31,12 +34,33 @@ const Layout = (props: LayoutProps) => {
...
+      <Markdown>
+        {t("nav.researchFeedback", { researchUrl: process.env.RESEARCH_FORM_URL })}
+      </Markdown>
+    )
+  }
+  let topAlert = null
+  if (process.env.TOP_MESSAGE) {
+    topAlert = (
+      <AlertBox
+        type={(process.env.TOP_MESSAGE_TYPE as AlertTypes) || "alert"}
```
james-wills-sf, line 49: I'm a little worried about using `as` here. I'm reading a book on typescript right now and they say to avoid it where possible. Like I'm not sure the implementation of AlertBox or what would happen if a non-AlertTypes string was passed in.

You could add a small top-level function to coerce it for you:

```
const asAlertType = (alertType: string): AlertTypes => {
  switch (alertType) {
    case "notice":
      return "notice"
    case "success":
      return "success"
    case "alert":
    default:
      return "alert"
  }
}
```

and then here you'd do:
```
type={asAlertType(process.env.TOP_MESSAGE_TYPE)}
```

app/javascript/layouts/Layout.tsx
```diff
@@ -31,12 +34,33 @@ const Layout = (props: LayoutProps) => {
...
+        inverted={process.env.TOP_MESSAGE_INVERTED === "true" || false}
```
james-wills-sf, line 50:
```suggestion
        inverted={process.env.TOP_MESSAGE_INVERTED === "true"}
```

nit: I think `|| false` is redundant here

james-wills-sf (commented):
Code and QA look good. Do we want to share env variables with angular, though?

I'm just thinking in the case of an incident we'll have to remember to update:
- ALERT_MESSAGE
- TOP_MESSAGE
- TOP_MESSAGE_TYPE

I could see either angular or react being forgotten

Maybe we should just re-use ALERT_MESSAGE instead of TOP_MESSAGE? I'm okay with calling it "Alert Message" even though it's not always an alert

jest.config.js
```diff
@@ -27,13 +27,12 @@ module.exports = {
   roots: ["<rootDir>/"],
   transform: {
     "^.+\\.(ts|tsx)$": "ts-jest",
+    ".+\\.svg$": "jest-transform-stub",
```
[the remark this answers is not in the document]
software-project: This stubs svg imports for tests. Otherwise tests are failing on that line.

app/javascript/layouts/Layout.tsx
software-project: that works, sure, will amend

software-project:
> Code and QA look good. Do we want to share env variables with angular, though?
> 
> I'm just thinking in the case of an incident we'll have to remember to update:
> 
> * ALERT_MESSAGE
> * TOP_MESSAGE
> * TOP_MESSAGE_TYPE
> 
> I could see either angular or react being forgotten
> 
> Maybe we should just re-use ALERT_MESSAGE instead of TOP_MESSAGE? I'm okay with calling it "Alert Message" even though it's not always an alert

@james-wills-sf regarding above I also had mixed feelings. But I assumed we are moving to React and we should name things as they are, therefor as we are allowing alert, primary and success it's not longer an `ALERT_MESSAGE`. What do you think is the best way to move forward? Should I rename Angular env var?

james-wills-sf:
@software-project sure, let's rename ALERT_MESSAGE in the angular code to TOP_MESSAGE.

akegan:
@software-project when you get a chance can you also update the PR title to be in the format of A good commit message plus the ticket number?

something like `Add react alert banner DAH-763`

james-wills-sf (approved):

PR #1476, opened 2021-05-25
Bump bloom to 1.0.0 and add generic router [DAH-746]

james-wills-sf:
Tangent from [DAH-746]

This PR does the following:
- remove `@sf-digital-services` dependency, we don't need it anymore
- Add back `@bloom-housing` dependency and bump it to the 1.0.0 release
- Add our SF routing solution in the form of `<NavigationProvider />`
  - This implements an interface for routing. Right now it's very simple but can be updated to work with react router or whatever we want.
- Update all existing usages of `<a />` and replace them with `<Link />`. We should use `<Link />` everywhere from now on.

[DAH-746]: https://sfgovdt.jira.com/browse/DAH-746


app/javascript/components/MetaTags.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import Head from "next/head"
+
+export interface MetaTagsProps {
+  title: string
+  image?: string
+  description: string
+}
+
+const MetaTags = (props: MetaTagsProps) => (
```
james-wills-sf, line 11: After bumping to 1.0.0 it looked like this Metatags component was removed from ui-components, does that sound right, @software-project?

I just copied this code directly from the public reference app.

app/javascript/util/routeUtil.ts
```diff
@@ -29,7 +29,7 @@ const getLocalizedPath = (
 }
 
 const localizedPathGetter = (newPathNonLocalized: string) => (
-  currentPath: string | undefined
+  currentPath: string | undefined = window.location.pathname
```
james-wills-sf, line 32: I just added this default so this code is still testable but we don't need to pass the pathname in explicitly everywhere.

app/javascript/components/MetaTags.tsx
software-project: Thats correct. They were moved to the reference app.

software-project (approved):
@james-wills-sf nice work. Links are working, also with translations. Nice work on simplyfying stuff as well.

PR #1478, opened 2021-05-26
Update readme and add ListingDirectory skeleton [DAH-823]

james-wills-sf:
Ticket: [DAH-823]
^ I created this ticket after doing the work because it seemed like a big enough change that we should remember to QA test it.

One of my [offboarding steps](https://docs.google.com/document/d/1w0PQwxFNLyGqLi0-cp86SysTi4EouYlkQsAMZmI1Uqo) is to update the README if anything is inaccurate. While reading the migration steps section I realized it's not accurate anymore since we simplified the migration process. While updating the README I decided to make the process even simpler, so that's what I did. I added the directory skeleton as an example.

So these are the main changes in this PR:
- Created a docs/ folder and added the React page migration guide as its own file
- Updated the way we configure which pages should be redirected. Now you only need to change one place in the angular code
  - This way also supports more complex URLs and will be able to support query params
- Changed behavior so the angular app checks feature flags to decide to redirect via rails or not.
  - This has the immediate effect of making navigation much faster while these flags are turned off
  - It is also likely to reduce bugs in the future. Ex: When we do the application form release, we won't need to do a rails redirect in the middle of the form any more. As long as the feature flag is turned off, it's behave like a single page Angular app  
- Added the listings directory skeleton
  - This was added as a proof of concept both for the readme and for the new redirect logic.

[DAH-823]: https://sfgovdt.jira.com/browse/DAH-823


akegan (approved):
LGTM. Review instructions were great, thank you!

PR #1481, opened 2021-06-02
Add MetaTags [DAH-798]

software-project:
Resolves [DAH-798](https://sfgovdt.jira.com/browse/DAH-798)

As we don't have listings pages that mostly pass the `og:` data I'm not sure how we want to tests this. Right now it's just serving default data. Also I've added comments (as discussed with @akegan) regarding functions serving assets.


app/javascript/layouts/MetaTags.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface MetaTagsProps {
+  title?: string
+  description?: string
+  image?: string
+}
+
+const MetaTags = (props: MetaTagsProps) => {
+  const { getAssetPath } = useContext(ConfigContext)
+  console.log(getAssetPath("apple-icon-57x57.png"))
```
james-wills-sf, line 16: nit: remove console log

james-wills-sf (commented):
@software-project I still need to do more QA but can you add a simple test for the metatags component?

software-project:
> @software-project I still need to do more QA but can you add a simple test for the metatags component?

@james-wills-sf sure will add them today

james-wills-sf (commented):
@software-project While QAing this it doesn't look like the meta information is working when sharing links

- 2021-06-04 at 8 51 03 AM

This is after I updated HOME_PAGE_REACT=true. It looks like it's still using application-angular to render the preview and I'm not sure why.

james-wills-sf (approved):
@akegan I think we can officially say that review apps have messed up meta tags and that it will behave differently once it is merged to full. I'm going to go ahead and merge this and will create a followup PR if it doesn't look right.

PR #1484, opened 2021-06-09
Initial listing directory data mapping [DAH-799]

james-wills-sf:
DAH-799

implement a mapper from our rails json responses to Bloom's Listing object and create a general interface for mappers going forward

Issues:
- Bloom listing object has a ton of extra fields, it would be nice if we could just use the required fields for the directory page
- Bloom listing object doesn't have sale fields
- Bloom doesn't seem to be very strict with their nullability, this makes it hard to know when something is actually nullable


app/javascript/api/adapters/listing/ListingPropertyAdapter.ts
```diff
@@ -0,0 +1,39 @@
...
+  // todo: implement
+  [1, 5]
+
+const ListingPropertyAdapter: Adapter<RailsRentalListing, Property> = (
+  listing: RailsRentalListing
+) => {
+  const [minHousehold, maxHousehold] = getMinAndMaxHouseholdSize(listing)
+  return {
+    unitsSummarized: UnitSummariesAdapter(listing),
+    units: [], // todo: populate this field
```
akegan, line 18: we'll need this more detailed unit info on the listing details page, but we don't need it for the directory so we can skip it for now

app/javascript/api/adapters/listing/RentalListingAdapter.ts
```diff
@@ -0,0 +1,63 @@
...
+import ListingPropertyAdapter from "./ListingPropertyAdapter"
+
+const toDate = (s: string): Date => new Date(s) // todo: handle different formats, actually implement this properly
+
+const ListingAdapter: Adapter<RailsRentalListing, Listing> = (item: RailsRentalListing) => ({
+  status: ListingStatus.active,
+  urlSlug: "", // todo: populate this field
+  displayWaitlistSize: false,
+  CSVFormattingType: CSVFormattingType.basic,
+  countyCode: CountyCode.Alameda, // todo: Add san francisco county to Bloom's enum
```
akegan, line 19: @software-project will add a bloom issue/pr for this

app/javascript/pages/DirectoryPage.tsx
```diff
@@ -11,13 +12,20 @@ interface DirectoryProps {
...
-    <Head>
-      <title>{t("t.dahliaSanFranciscoHousingPortal")}</title>
-    </Head>
-    Is Rental: {String(props.isRental)}
-  </Layout>
-)
+const DirectoryPage = (props: DirectoryProps) => {
+  useEffect(() => {
+    // todo: remove this. it's just for display purposes
+    void getRentalListings().then((listings) => console.log(listings))
```
software-project, line 18: @akegan should we keep it for not to make developing frontend a bit easier?

akegan:
@software-project FYI looks like unit tests are failing in react

app/javascript/pages/DirectoryPage.tsx
akegan: yes let's keep it for now since we don't have any other way of testing this PR and seeing the data while we develop the rest of the page.

app/javascript/api/adapters/listing/RentalListingAdapter.ts
```diff
@@ -0,0 +1,71 @@
...
+  createdAt: null, // todo: update this field
+  updatedAt: null, // todo: update this field
+  applicationMethods: [],
+  assets: [
+    {
+      id: "",
+      createdAt: null,
+      updatedAt: null,
+      label: "building",
+      fileId: item.imageURL,
```
akegan, line 37: something funky is going on here where on full, the imageURL attribute is a cloudfront url, and here it's coming through as a MOHCD image url. I wonder if there's some processing going on in the frontend that we're missing.

akegan: actually after a bit of googling I think it's just because it's a review app and it doesn't have all of the cloudfront urls cached in a local db, so this should be ok! 

Code where that's done here:
 https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/services/force/listing_service.rb#L107-L118

app/javascript/api/adapters/listing/RentalListingAdapter.ts
```diff
@@ -0,0 +1,71 @@
...
+} from "@bloom-housing/backend-core/types"
+
+import RailsRentalListing from "../../types/rails/listings/RailsRentalListing"
+import { Adapter } from "../adapter"
+import ListingPropertyAdapter from "./ListingPropertyAdapter"
+
+const toDate = (s: string): Date => new Date(s) // todo: handle different formats, actually implement this properly
+
+const ListingAdapter: Adapter<RailsRentalListing, Listing> = (item: RailsRentalListing) => ({
+  status: ListingStatus.active,
```
akegan, line 15: hm, is it possible to not default to a status of active here in this PR? or does that need to be in a follow-up?

akegan:
From my first pass, these are the missing pieces that I can see:

1. ListingStatus -> it would be nice if we could wire this up instead of having all of the listings have a status of active, however, it looks like we're not using it because it's not being returned by the browseListings subset. Can we default to null instead?
2. We'll need `Lottery_Status` to be able to determine which listings have had a lottery run, is that something that's missing from bloom or can we wire it up in this PR?
3. `Lottery_Results_Date` we use this for this banner on listing dir page. Not a blocked for this PR, but would be nice to add if it's a field on bloom already
image
4. Priority units: We use this to say if the listings contain units for people with mobility or hearing and vision impairments. In our browseListings it's stored under "prioritiesDescriptior" but maybe it's something that could come from somewhere else. Does it exist at all on bloom?
image
image
5. max income ranges. We need to be able to show the max income ranges on the listing directory page. We use `absoluteMinIncome` and `absoluteMaxIncome` in the unitSummaries. My guess is that that's something we'll have to contribute back to bloom 
 
image
image

software-project:
@akegan I'll address the comments, cheers.

software-project:
@akegan 
1. I've added status for active and closed listings
2/3. We don't have much implemented on Bloom regarding lotteries. We have events that can be a public lottery and then we can add an url with results there. I suppose we need more here, so we should arrange something with Sean and Jessie. I've added a publicLottery event for now.
4. We used to have `prioritiesDescriptior` but it was never used. Now it's replaced with `priorityTypes` in unit summaries. I've mapped them as well.
5. Income range is also mapped on unitSummaries.

app/javascript/api/adapters/listing/RentalListingAdapter.ts
```diff
@@ -37,7 +39,13 @@ const ListingAdapter: Adapter<RailsRentalListing, Listing> = (item: RailsRentalL
       fileId: item.imageURL,
     },
   ],
-  events: [],
+  events: [
```
akegan, line 42: Add a comment in here to make it clear that it's a temporary workaround for the listing directory page. Something like:
```
// TODO: Solve how to clearly communicate about lotteries. This is a temporary solution to be able to display lottery results data on the listing directory page, but the lottery results date is not the same as the lottery start date.
```

app/javascript/api/adapters/listing/RentalListingAdapter.ts
akegan: https://github.com/bloom-housing/bloom/issues/1407

akegan:
Throwing an error now:
```
UnitSummariesAdapter.ts:49 Uncaught (in promise) TypeError: Cannot read property 'map' of null
    at u (UnitSummariesAdapter.ts:49)
    at l (ListingPropertyAdapter.ts:22)
    at s (RentalListingAdapter.ts:24)
    at adapter.ts:5
    at Array.map (<anonymous>)
    at adapter.ts:5
```

app/javascript/api/adapters/listing/UnitSummariesAdapter.ts
```diff
@@ -0,0 +1,60 @@
...
+  max: max != null && asCurrencyString(max),
+})
+
+const getMinMax = (min: number, max: number): MinMax => ({ min, max })
+
+const UnitSummaryAdapter: Adapter<RailsRentalUnitSummary, UnitSummary> = (
+  s: RailsRentalUnitSummary
+) => ({
+  unitType: s.unitType,
+  minIncomeRange: getMinMaxCurrency(s.absoluteMinIncome, s.absoluteMaxIncome),
```
akegan, line 26: So this is called the min income range, but for us it's actually the full income range. This might work for now, but we'll have to come back to it with bloom

akegan (approved):
LGTM as a first pass! thanks for finishing it up. I'll get in touch with Sean about a time to go over some of the missing pieces

akegan:
Follow-up from meeting with @seanmalbert

### Income range
Consider changing `minIncomeRange` to just `incomeRange`, and then Bloom populates the max with the max min income, and we populate it with the max max income, but he's going to talk to Jesse about it

### Lotteries
Trying to figure out if we want to extend events to add more fields, or create new lottery object. Leaning towards keeping it as an event, including a lottery results event

- Lottery Date
- Lottery Results Date (when there are supposed to be results, not that there are results, updated after the fact)
- Are there lottery results?

1. Add an address object to the event which will include venue, street address, city, etc (and address ID for their DB that we would keep blank. 
2. Andrea to figure out what we actually want for lottery stuff on the directory page, do we even need the lottery status? Or can we just use the lottery time? Add lottery status to event? or, add that at the listing level? Maybe lean towards "isLotteryComplete" and we do some logic to look at `listing && listing.Publish_Lottery_Results && (listing.Lottery_Status == 'Lottery Complete')`, and check for lottery results available too. 

### Reserved units
Bloom is moving to have reserved info at the listing instead of the unit too.

PR #1486, opened 2021-06-11
Test out the sfgov eslint plugin

shawnbot:
This installs a pre-release of [eslint-plugin-sfgov](/sfdigitalservices/eslint-plugin-sfgov/pull/1) and replaces the `standard` and `eslint:recommended` entries in `extends` with `plugin:sfgov/recommended`, which [extends](https://github.com/SFDigitalServices/eslint-plugin-sfgov/blob/46b5b428ab2dd512f7beb7ceb71057add15582cf/configs/recommended.js#L3-L4) both of those.

It also includes `plugin:promise/recommended`, which flagged a couple of instances of promises that don't resolve to explicit values, such as:

https://github.com/SFDigitalServices/sf-dahlia-web/blob/758587c6b0a74ae377e5d69287e8da3be590ff06/app/javascript/authentication/context/UserProvider.tsx#L23-L26

This isn't really an error though because it's done in a `useEffect()` to dispatch the user reducer. So it's a promise with side-effects, but this is a totally valid pattern in hooks-heavy React code, so I've disabled the `promise/always-return` and `promise/catch-or-return` rules that it breaks.


akegan:
@james-wills-sf if you have a little extra time can you take a look at this and how it interacts with our custom eslint settings?

james-wills-sf (approved):
@shawnbot this lgtm. I did add one commit to fix the `yarn lint` command. I don't think this PR broke it but it's a small enough change I don't mind throwing it in this PR

@akegan I was surprised our code didn't break a lot of these rules especially since we don't use standard anymore. But I checked out the branch and ran yarn lint (and also checked that linting in vscode worked) and it all seems good!

akegan:
Before we merge, it'd be good to clean out our eslintrc to make sure that we're not duplicating rules, and so it's easier to see where we differ from the sfgov plugin

package.json
```diff
@@ -61,6 +61,7 @@
     "eslint-plugin-promise": "^4.3.1",
     "eslint-plugin-react": "^7.21.4",
     "eslint-plugin-react-hooks": "^4.1.2",
+    "eslint-plugin-sfgov": "0.0.0-03ced82",
```
shawnbot, line 64: I also want to release `1.0.0` and update this before marking this as ready. I'll make this change locally and update the lockfile:

```suggestion
    "eslint-plugin-sfgov": "1.0.0",
```

james-wills-sf (changes requested):
@shawnbot marking this as request changes just so we don't merge just yet but I plan on making the changes myself. I'm just going to remove the redundant plugins (standard and import are the two that come to mind, but I'll check them all)

james-wills-sf (approved):
I just removed the redundant import plugin, this lgtm now

We don't want to remove the prettier/standard plugin, we need that for prettier to work I think

PR #1487, opened 2021-06-15
Polish react homepage DAH-747

akegan:
DAH-747


akegan:
@software-project this is ready for review when you get a chance, review instrux are in the ticket

app/javascript/layouts/Layout.tsx
```diff
@@ -65,10 +65,16 @@ const langItems = getLanguageItems()
...
 
+  const feedbackBanner = (
+    <Markdown>
+      {t("nav.getFeedback", { feedbackUrl: "https://airtable.com/shrw64DubWTQfRkdo" })}
+    </Markdown>
+  )
+
   const topAlert = (
     <>
       {process.env.TOP_MESSAGE && (
```
software-project, line 80: I think we can enable now `boundToLayoutWidth` and `narrow` for `AlertBox`

software-project (commented):
Besides banner there is still a matter with logo. But we will need to either replace it or request a header without text with Bloom. Besides that it looks nice. Also I wonder if we should change link color for AlertBox, that blue one looks a bit off.

akegan:
@software-project I commented out the css purge. Can you re-review and QA. There are review instrux in the ticket

software-project:
@akegan blue banner is not visible on the mobile view. That is a bloom issue I assume. Besides that it looks good.

akegan:
@software-project the blue banner not showing on mobile is intentional, I think. We haven't had it show in the angular version either I think.

For the alert banner link being blue, it seems like the best course of action is to fix it in ui-components css? And if we need an alert message before we incorporate that change we can add some hacky css to the html env var.

Can you approve the PR if it looks good to go?

software-project (approved):

software-project:
@akegan it's approved. But on the angular, the blue banner is showing up for me, thats why I raised the concern. I will add a ticket for bloom to fix link on blue banner.

akegan:
oh I see, thanks. I think we added that to angular fairly recently. Can you link back the issue you create when you create it?

PR #1488, opened 2021-06-18
Fix for idle timeout modal DAH-854

software-project:
[DAH-854](https://sfgovdt.jira.com/browse/DAH-854)


app/views/layouts/application-react.html.slim
```diff
@@ -25,7 +25,7 @@ html lang="en"
         iframe height="0" src="//www.googletagmanager.com/ns.html?id=#{ENV['GOOGLE_TAG_MANAGER_KEY']}" style="display:none;visibility:hidden" width="0"
     #google_translate_element
 
-    #root
```
akegan, line 28: how does this fix it? Can we add a comment here to make it clearer?

akegan:
@software-project I tried to get the modal to show up 3 times and was only successful once. The other two times I waited ~10 mins and didn't get a modal. 

The time that I did get a modal was when I had been logged in for a while, then went to the listing directory page (angular), then went back to the homepage and waited another 5 mins. Not exactly sure what's going on.

software-project:
@akegan I tested it on both localhost and review app and it's working correctly in both places. Actions resetting the timeout are: `"mousemove", "keypress", "scroll"`, so basically any activity on the page cancels the timeout. Tested both case when I was inactive on the page and when the browser was inactive and it worked on both cases.

akegan:
@software-project this time I tried testing not touching my computer at all and still got inconsistent results.

1. When I had it up on a separate screen and didn't put my mouse anywhere on the screen, I didn't get the modal
2. I then navigated to the listing dir page and back, and after 5 mins of not touching my computer I got the modal. I clicked "continue"
3. I didn't touch my computer for another 8-10 minutes, but didn't get the modal again.
4. I then navigated to the listing dir page and back again, then didn't touch my computer for 5 mins again and got the modal. I didn't click on it, and I was logged out after another minute (as expected)

It does really seem tied to navigating to the listing directory (or maybe another angular) page and back somehow, but either way I'm not getting super consistent results, which worries me because I don't want folks to stay logged in forever.

akegan:
@software-project I think I've figured out what's going on! In the cases where I haven't gotten the modal, I'm not getting any timer resets. 

2 cases:
1. I changed the url to ?react=true in the address bar, then didn't actually click into the page or move my mouse around
2. After I then moved my mouse around, then waited and got the modal. I clicked continue but didn't move my mouse after clicking continue, then the modal didn't show up again.

I think this is pretty consistent with what I've done the past few times too. Could it be something where the reset timeout is doing something that's missing from just the start timeout setup?
image

software-project:
@akegan I've added the fix we discussed on the call. Seems to be doing the trick, but please double check. Console logs are also removed now.

akegan (approved):
Works for me now! 🚀

PR #1494, opened 2021-07-13
Add Rental listings directory DAH-834

software-project:
Resolves [DAH-834](https://sfgovdt.jira.com/browse/DAH-834)

Couple notes here:
1. I kind of monkey patched listing events to be able to show upcoming and lottery results. So if lottery status is complete, I'm adding `endDate`. I've added a comment there
2. Our listings paths don't match Blooms's. And as it's hardcoded into the component I've added a temporary route redirect. Also language is not added to the path on the Bloom side, as i18n is handled differently. We need to figure out a way to parametrize this. I think Sean is planning to make some changes to ui-components so they accept singular components rather than full listing. That would help a lot and we would not need the adapter.
3. Listing group headers should be change to `ActionBlock` and allow the use of sub headers.
4. Seems like address formats are a bit different as well.
5. Last two you are already familiar single label and different tables.


app/assets/json/translations/react/en.json
```diff
@@ -32,10 +50,17 @@
   "t.continue": "Continue",
   "t.continueWithYourApplication": "Continue with your application?",
   "t.dahliaSanFranciscoHousingPortal": "DAHLIA San Francisco Housing Portal",
+  "t.income": "income",
+  "t.minimumIncome": "",
```
akegan, line 54: why is this one blank?

app/javascript/api/adapters/listing/RentalListingAdapter.ts
```diff
@@ -43,9 +43,11 @@ const ListingAdapter: Adapter<RailsRentalListing, Listing> = (item: RailsRentalL
     // TODO: Solve how to clearly communicate about lotteries. This is a temporary solution to be able to display lottery results data on
     // the listing directory page, but the lottery results date is not the same as the lottery start date.
     item.Lottery_Results_Date && {
-      type: ListingEventType.publicLottery,
+      type: ListingEventType.lotteryResults,
       startTime: toDate(item.Lottery_Results_Date),
-      endTime: null,
+      // TODO: To temporary differenciate upcoming from complete I'm addind endTime to completed listings
```
akegan, line 48: just a couple minor typos
```suggestion
      // TODO: To temporarily differentiate upcoming from complete I'm adding endTime to completed listings
```

app/javascript/pages/DirectoryPage.tsx
```diff
@@ -12,19 +13,91 @@ interface DirectoryProps {
...
+      hideButtonText={t("listings.lotteryResults.hide")}
+      showButtonText={t("listings.lotteryResults.show")}
+    />
+  )
+
+const DirectoryPage = (_props: DirectoryProps) => {
+  const [listings, setListings] = useState<ListingsGroup>({ open: [], upcoming: [], results: [] })
+  const [loading, setLoading] = useState<boolean>(true)
   useEffect(() => {
     // todo: remove this. it's just for display purposes
```
akegan, line 55: This comment doesn't seem relevant anymore
```suggestion
```

cypress/integration/rentalListings.e2e.ts
```diff
@@ -0,0 +1,17 @@
+describe("Rental listings directory page", () => {
```
akegan, line 1: 🎉

app/javascript/pages/DirectoryPage.tsx
```diff
@@ -12,19 +13,91 @@ interface DirectoryProps {
...
-const DirectoryPage = (props: DirectoryProps) => {
+interface ListingsGroup {
+  open: Listing[]
+  upcoming: Listing[]
+  results: Listing[]
+}
+
+const openListingsView = (listings) =>
+  listings.length > 0 ? (
+    <ListingsList listings={listings} />
```
akegan, line 24: this is really interesting. I didn't realize that this was all tied into a "listings group" component. I sort of assumed that we'd pass in a list of listing directory objects with Image cards and tables, etc. 

Do you have a preference for what a better interface might be? I think we can probably talk to Sean tomorrow when Anthony and I meet with him at 2pm. You are also welcome to join if you'll still be online (I realize it's like 10pm your time).

akegan (commented):
Looks mostly good to me! lots less code than I expected, but I have some questions about the `ListingsList` and `ListingsGroup` components

app/javascript/pages/DirectoryPage.tsx
software-project: Regarding it being just one component I think we need to go the way you assumed it to be. Now we have little control over customization. If we make components less dependent and use flat structure passing singular params rather than whole listing would be a viable way to go. And in that case we are not going to need adapter.

akegan (approved):

PR #1496, opened 2021-07-17
Add TIDA Treasure Island Resident Preference DAH-862

akegan:
DAH-862

TODO:
- [x] Get custom pref record type fixed to allow for certificate #
- [x] Add/update tests
- [x] Maybe add links to certificate # caption
- [x] Push up for translations
- [x] Look into dropping the application member question (going to leave it in)
- [x] Look into overriding description with locale-en so we can translate it.  (I think it's fine to not override)

Review instruction notes:

- Make sure COP + DTHP still show up OK
- Make sure 588 mission pref still works


app/services/force/short_form_service.rb
```diff
@@ -16,6 +16,8 @@ def self.check_household_eligibility(listing_id, params)
 
     def self.create_or_update(params, contact_attrs)
       params[:primaryApplicant].merge!(contact_attrs)
+      puts 'REQUEST TO SALESFORCE'
+      puts JSON.pretty_generate(params)
```
software-project, line 20:
```suggestion
```

software-project (approved):
I think that log can be removed. Besides that all good and working smoothly. Nice one.

PR #1500, opened 2021-08-05
Fix navbar logo styling DAH-882

akegan:
DAH-882

This PR fixes navbar logo styling for both react and angular:
- Replaces angular and react logo with a new svg from Nicole Lee
- In react, this pulls in the bloom update that sets a minimum height of the menu bar at 75px
- In angular, fix the padding issues that come up because of the new react styling

css updates for angular are in this pattern lib pr: https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/174
css updates for react were in this bloom pr: https://github.com/bloom-housing/bloom/pull/1597

Goal is to have a consistent look when switching from angular to react, and to have a legible logo at mobile widths.


akegan:
React unit tests are failing because of a missing .env file. I will investigate early next week!

software-project (approved):
Looks really nice. Ready to go when tests are passing.

app/javascript/pages/base.scss
```diff
@@ -244,4 +244,9 @@ $colors: (
   }
 }
 
-@import "tailwindcss/utilities";
+// Override bloom styles so that navbar doesn't overflow on medium screens.
```
akegan, line 247: @software-project can you re-review this? this feels really hacky but is the only way I could figure out how to avoid this overflow issue. Do you have any other ideas?
image (45)

software-project: Well, I tried to play with shrink and frow, but all in all it got me to the same result. I would assume that those properties were designed to serve exactly that purpose.

PR #1501, opened 2021-08-10
Fix page title translations DAH-880

software-project:
Resolves [DAH-880](https://sfgovdt.jira.com/browse/DAH-880)

This PR fixes issue with tab titles. In some places we were using Next.js head, which is not working and in the other MetaTags were used twice, so they were overriding each other. Some translations were missing as well. All should be fixed now.


app/javascript/layouts/MetaTags.tsx
```diff
@@ -13,6 +13,7 @@ export interface MetaTagsProps {
 
 const MetaTags = (props: MetaTagsProps) => {
   const { getAssetPath } = useContext(ConfigContext)
+  // Description is separeted into two check as Helmet can't handle nested elements
```
akegan, line 16:
```suggestion
  // Description is separated into two check as Helmet can't handle nested elements
```

akegan (approved):
looks like translation for 
```
"t.dahliaSanFranciscoHousingPortal": "DAHLIA San Francisco Housing Portal"
```
is missing for filipino, but it's also missing on the angular page, so maybe it's not something for this PR!

LGTM otherwise

akegan:
oh actually nvm it looks like the translation in Filipino is just the same as english

software-project:
@akegan I was surprised with the Filipino translation as well ;)

PR #1502, opened 2021-08-10
Fix English language prefix DAH-866

software-project:
Resolves [DAH-866](https://sfgovdt.jira.com/browse/DAH-866)

English language prefix was set to `en`, but we are using empty string in that case.


akegan (approved):
LGTM.

PR #1511, opened 2021-09-10
Allow rich text in lottery venue field DAH-932

akegan:
DAH-932


akegan:
Almost ready for review, just need Salesforce code to be deployed to full for testing before we can review.

emilyjablonski:
The link on Civic Center looks great but I'm not seeing a link on the automated listing test, is that expected?

akegan:
I had the same problem because the section is way down the page for the lottery on automated test listing
public lottery section

emilyjablonski (approved):

PR #1512, opened 2021-09-11
Old version proper widths

akegan:
This is just to deploy an older version for comparison. do not merge


akegan:
this was deployed so we don't need this review app anymore

PR #1516, opened 2021-09-16
Require accountholder to be over 18 DAH-922

akegan:
DAH-922

Review instructions in the ticket. 

I included some mild refactoring but the way scope works in AngularJS means that there's still a lot of repeated logic 😕. Let me know if you can think of any better way to organize.

### DIFF
--- a/app/assets/javascripts/account/templates/account-settings.html.slim
+++ b/app/assets/javascripts/account/templates/account-settings.html.slim
@@ -105,20 +105,66 @@
                 .form-group-month.large-2.columns
                   label.sr-only for="date_of_birth_month"
                     | {{'label.dob_month' | translate}}
-                  input(ng-model="userAuth.contact.dob_month" ng-required="true" maxlength="2" name="date_of_birth_month" placeholder="MM" type="text" ng-class="{ error: inputInvalid('date_of_birth_month', '', form.accountNameDOB) }" ng-disabled="isLocked('dob')" pattern="\d*" ui-validate="'DOBValid(\'month\', $value)'" numeric-only="" autocomplete="off" ng-change="recheckDOBDay('accountNameDOB'); displayChangeNotice('nameOrDOBChanged')" aria-describedby="date_of_birth_error")
+                  input(
+                    ng-model="userAuth.contact.dob_month"
+                    ng-required="true"
+                    maxlength="2"
+                    name="date_of_birth_month"
+                    placeholder="MM"
+                    type="text"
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/account/templates/account-settings.html.slim
```diff
@@ -105,20 +105,66 @@
                 .form-group-month.large-2.columns
                   label.sr-only for="date_of_birth_month"
                     | {{'label.dob_month' | translate}}
-                  input(ng-model="userAuth.contact.dob_month" ng-required="true" maxlength="2" name="date_of_birth_month" placeholder="MM" type="text" ng-class="{ error: inputInvalid('date_of_birth_month', '', form.accountNameDOB) }" ng-disabled="isLocked('dob')" pattern="\d*" ui-validate="'DOBValid(\'month\', $value)'" numeric-only="" autocomplete="off" ng-change="recheckDOBDay('accountNameDOB'); displayChangeNotice('nameOrDOBChanged')" aria-describedby="date_of_birth_error")
+                  input(
```
akegan, line 108: reformatted to make this easy to read, but makes the diff hard to read. Keys I touched throughout were `ng-change` and `ui-validate`

spec/javascripts/services/account_service_spec.coffee
```diff
@@ -68,12 +64,11 @@ do ->
       $provide.value '$translate', $translate
       $provide.value 'bsLoadingOverlayService', fakeLoadingOverlayService
       $provide.value 'ShortFormApplicationService', fakeShortFormApplicationService
-      $provide.value 'ShortFormDataService', fakeShortFormDataService
       $provide.value 'ModalService', fakeModalService
       return
     )
 
-    beforeEach inject((_AccountService_, _$state_, _$auth_, _$q_, _$httpBackend_) ->
+    beforeEach inject((_AccountService_, _ShortFormDataService_, _$state_, _$auth_, _$q_, _$httpBackend_) ->
```
akegan, line 71: pulled in real ShortFormDataService because too much mocking makes me worry about the actual functionality

app/assets/javascripts/short-form/ShortFormDataService.js.coffee
```diff
@@ -636,6 +636,7 @@ ShortFormDataService = (ListingDataService, ListingConstantsService, ListingPref
       when 'year'
         year >= 1900 && year <= validYear
 
+  # FIXME: this will fail in the year 2100
```
akegan, line 639: this should be refactored, but I decided against it because in theory we will replace this logic before the year 2100

emilyjablonski (approved):

PR #1517, opened 2021-09-17
Fix missing feedback url on angular blue bar DAH-946

akegan:
DAH-946

Found during our bug bash, the feedback url was missing from the bar at the top of the angular page so it was just linking back to the homepage. 
image


akegan:
No longer a hotfix, we'll deploy in the next release next week.

emilyjablonski (approved):

PR #1519, opened 2021-09-22
feat: uptake new bloom components on listing directory react page

emilyjablonski:
Only includes a subset of AC from the ticket.

In this PR I uptake the new Bloom components which remove dependencies on Bloom's backend. Because of that I was able to remove the data model adapter.

TODOs left:
* Using the ResponsiveTable (will need an update on Bloom's side)
* Subheader under unit summary table header (will need an update on Bloom's side)
* Favorite button (will need an update on Bloom's side)
* Copying over new translation strings
* Directory header
* Date string formatting


.eslintrc.js
```diff
@@ -165,25 +165,7 @@ module.exports = {
       },
     ],
     "unicorn/prevent-abbreviations": "off",
-    "import/order": [
```
emilyjablonski, line 168: This was preventing me from committing since it had an opposite opinion from Prettier

akegan:
@emilyjablonski one minor note is that we include the Jira ticket number in PR titles. I just update the [PR workflow section](https://sfgovdt.jira.com/wiki/spaces/HOUS/pages/2014609427/Eng+Workflow#Pull-Request-Review-Workflow) of our workflow doc to make sure it's up to date.

So it'd look something like `Uptake new bloom components on listing directory react page DAH-959`

It's also handy to then repeat the ticket # in the PR description because we have autolinking setup ➡️   DAH-959, so it'll make a link and then it's really easy to click back into the Jira ticket to look for review instructions and more context

app/javascript/pages/DirectoryPage.tsx
```diff
@@ -23,15 +28,129 @@ interface DirectoryProps {
...
+  Reserved_community_type: string
 }
 
+const getListings = (listings) =>
+  listings.map((listing: Listing, index) => (
+    <ListingCard
+      key={index}
+      imageCardProps={{
+        imageUrl: listing.imageURL,
+        subtitle: `${listing.Building_Street_Address}, ${listing.Building_City} ${listing.Building_State}, ${listing.Building_Zip_Code}`,
```
akegan, line 130: this was an issue before this PR too, but if any of these are undefined they show up as "undefined" would be nice to just e.g. leave out the zip code if not available

app/javascript/pages/DirectoryPage.tsx
```diff
@@ -23,15 +28,129 @@ interface DirectoryProps {
...
+    })
+  }
+
+  return statuses
+}
+
+const getNumberString = (currencyNumber: number) =>
+  currencyNumber ? currencyNumber.toLocaleString() : null
+
+const getRangeString = (min: number, max: number, suffix: string, prefix?: string) => {
```
akegan, line 71: Could you add tests for some of these functions with more complex logic?

akegan (commented):
This already looks great!

A few things that I noticed to fix either in this PR or a follow-up PR
1. Sorting of upcoming lottery vs lottery results is still a bit wonky. Angular has 
22 FRANKLIN UNIT 505 in the upcoming lottery section, for example, because the lottery is not complete. 
2. I'm guessing you're planning to add availability data + update the column titles once the responsive table is exported, but wanted to flag it just in case. 
3. I'm not sure what's going on with this Argenta picture. Is there an extra grey bar added because the image is so short? 
image

akegan:
@emilyjablonski FYI for next time that we squash merge so that we have a clean history on main, which makes it easier for us to get releases ready to go

PR #1521, opened 2021-09-28
Go to sign-in in correct language from react homepage DAH-950

akegan:
DAH-950


emilyjablonski (approved):
LGTM! Only small comment is that the email placeholder in Spanish is still `example.com`

akegan:
We have example.com as the placeholder on prod too, so going to call that ok. Thanks!

PR #1527, opened 2021-10-08
Bring in new SiteHeader DAH-982

emilyjablonski:
#982 
There are some issues visually in the SiteHeader as you are moving to mobile, for a moment the logo is partially off-screen and the menu titles go onto two lines but there is enough space. It is likely a flex issue.

Since this latest flavor of ui-components is on Tailwind 2 and React 17 I had to make those upgrades here too.


emilyjablonski:
@akegan Sounds good that's the same issue I noticed and think the Bloom update will fix it

akegan:
Couple more wonky things after more QA (I had forgotten to test signed in):

1. When you use the url param to add ?react=true, it's no longer persisted when you switch to another language
2. When signed in in english, the "sign out" option styling looks a bit funky like it's short
image
3. I'm not sure why the house icons are in that menu either 
4. The spanish and filipino "My Account" dropdowns are  unusable because of some styling issues. You can't select any of the values
image
image

emilyjablonski:
@akegan 
1) Not 100% sure what you mean here?
2) Fixed in next release
3/4) Testing and fixing now

PR #1530, opened 2021-10-15
Polish Listing Directory DAH-847

akegan:
DAH-847

- [ ] Does not have to match angular exactly, but has to look good on desktop and mobile, and with translations
- [x] “Favorites” button does not need to be there
- [x] All of the date banners and icons look as expected, including that dates from the past calendar year include the year. **It turns out that the DS style guide prefers the bloom format of e.g. "January 1, 2021"**
- [x] Everything is translated using our old translations
- [ ] Anything else lingering from DAH-959
- [x] Refine buckets of when listings show in upcoming lottery vs lottery complete.
- [x] Pull in new StackedTable component to get the tables looking beautiful

Notes on styling differences for the table so far:
- Missing line between table header and each row. 
- Title and subtitle aren't lined up with the content in the table on the left
- Fonts are a bit smaller on react
   - Angular headers are at 14px, "Studio", incomes etc are at 16px
   - React version everything is 13px
- "See Details" is left-aligned on react, right-aligned on angular. We could do this intentionally because the favorite button is gone, but should get an opinion from a designer.   

Angular:
image

React:
image

### DIFF
--- a/app/javascript/__tests__/pages/DirectoryPage.test.tsx
+++ b/app/javascript/__tests__/pages/DirectoryPage.test.tsx
@@ -48,6 +48,7 @@ describe("<DirectoryPage />", () => {
       Application_Due_Date: "2000-10-30T00:00:00.000+0000",
       Lottery_Results_Date: "2000-10-31T00:00:00.000+0000",
       Publish_Lottery_Results: true,
+      Lottery_Status: "Lottery Complete",
     }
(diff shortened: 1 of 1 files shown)


akegan:
@emilyjablonski will make those design tweaks

akegan:
@emilyjablonski this is ready for review!

The only outstanding things I see are:
- Pulling in latest ui-components to get larger black image overlay
- Minor style tweaks to stacked table.

PR #1533, opened 2021-10-22
Fix grouping of Mira units DAH-1003

akegan:
DAH-1003


akegan:
@emilyjablonski reminder to please review this PR

.circleci/config.yml
```diff
@@ -195,6 +195,7 @@ jobs:
       - setup
       - run: RAILS_ENV=test bundle exec rails db:create
       - run: RAILS_ENV=test bundle exec rails db:test:prepare
+      - run: yarn install
```
akegan, line 198: I'm not really sure why this was necessary, because it should have been run in the build step, but it did fix the issue when I ssh'd in to the circleCI job, and just now when the full circleCI suite ran

akegan:
It looks like some of the prices changed since I put up the PR so that they're slightly different but rounding to the same number. I'll ask around and see why that might have happened and fix

akegan:
@emilyjablonski the data migration team fixed the data so this is ready for qa again

emilyjablonski (approved):

PR #1541, opened 2021-12-01
Add babel polyfill to fix IE11 error DAH-1029

akegan:
DAH-1029

This PR:
1. Adds [the new equivalent of babel polyfill](https://babeljs.io/docs/en/babel-polyfill)
2. Adds a min height for site-content for IE11 only [source](https://stackoverflow.com/questions/20541306/how-to-write-a-css-hack-for-ie-11)

There's also an issue with nanoid if you try to use ie11 with a dev build, but our deployed apps don't have this problem.

I'm also going to go back to the Digital Services team to see if we should move off IE 11 support completely and just show a banner instead, since this hack will not work for the listing directory or details pages.


emilyjablonski:
As discussed there are still a lot of overlap/visual issues but the page is functional

emilyjablonski (approved):

PR #1542, opened 2021-12-03
Implement Mira Accessible units only features DAH-1023

akegan:
DAH-1023


akegan:
@emilyjablonski this PR is ready for review.

spec/javascripts/services/short_form_application_service_spec.coffee
```diff
@@ -37,6 +37,19 @@ do ->
       listing:
         Id: ''
       loadListing: ->
+      preferenceMap = {
```
akegan, line 40: This is to fix the flaky unit-angular tests

emilyjablonski:
I think I may not be reviewing this correctly. On the deploy preview I go to sales and open the Mira listing and I'm not seeing the reserved copy or the question.

akegan:
@emilyjablonski  I think it must be because of some testing on salesforce. one sec

emilyjablonski:
Two notes that might not be issues:

* When you click apply, the image card on that screen doesn’t have the orange reserved tag
* The new qualify question is not showing on the application summary page that's right before submit

akegan:
@emilyjablonski thanks for flagging those things. I just confirmed that we don't show the answer to the reserved question on the summary page for other reserved listings, and that we don't show the reserved tag on that first page either, so I think we're good!

emilyjablonski (approved):

PR #1549, opened 2021-12-14
Link sale directory action block to correct links DAH-839

akegan:
Fixes bug in DAH-839 where sale directory action block wasn't linking to the right pages
image

To test, compare the links for react sale directory page to the links on the angular sale directory


emilyjablonski (approved):
LGTM but this is not being released right now, yeah?

akegan:
yep, we're not turning it on for users yet, but wanted to fix the issue since Yindi found it during release testing.

akegan:
merging even with failing e2e tests because those are due to a Salesforce issue

PR #1550, opened 2021-12-14
Remove "suggested address" from address verification page DAH-1025

akegan:
DAH-1025

Removes "suggested address:" from the verified address page per Yindi's request.

Doesn't remove it based on whether the address was found or not because we don't have access to that info in the scope of the page. 
image


akegan:
force merging because e2e test failures are due to a salesforce change.

PR #1560, opened 2022-01-19
Add back listing-results anchor tag DAH-840

akegan:
Fixes a bug caused by merge issues with DAH-840

To QA:
This "See the listings" button should scroll down to the start of the listings, both on desktop and mobile
image


jacksongonzales (approved):
looks good!

PR #1570, opened 2022-02-01
feat: show full page alert if using ie

emilyjablonski:


akegan (approved):
Looks a bit wonky on IE10 and IE9 but I am OK with that! Looks good on IE11
image

I also confirmed that it looks OK on Firefox, Chrome, and Edge on a windows 7 computer

@emilyjablonski this is good to go with 2 small items:
- can you add review instructions before merging? 
- Can you update the PR title to be in the pattern of `Good commit message DAH-[ticket #]` (for example, this could be "Show full page alert for IE users DAH-1051") 
- And a reminder to squash merge when you do merge!

PR #1573, opened 2022-02-02
Add translations for unit types on listing directory page DAH-1053

jacksongonzales:
DAH-1053

Add translations for: 

* Unit types (1BR, Studio)

* 1 BR should be updated to be “1 bedroom”

* Studio can be translated

* Update SRO to "SRO (single-room occupancy)"


app/assets/json/translations/react/en.json
```diff
@@ -64,10 +64,12 @@
   "listings.reservedCommunityType.habitat": "Habitat Greater San Francisco",
   "listings.reservedCommunityType.senior": "Senior Building",
   "listings.reservedCommunityType.veteran": "Veterans Building",
-  "listings.unitTypes.1 BR": "1 BR",
-  "listings.unitTypes.2 BR": "2 BR",
-  "listings.unitTypes.3 BR": "3 BR",
-  "listings.unitTypes.SRO": "SRO",
+  "listings.unitTypes.1 BR": "1 Bedroom",
```
akegan, line 67: I think we don't want these spaces in the keys, I'm not sure that it'll work with Phrase with the type of files we have set up.

It is a nice solution, so we could try it to see if it'll work OK.

app/assets/json/translations/react/en.json
```diff
@@ -64,10 +64,12 @@
...
+  "listings.unitTypes.2 BR": "2 Bedroom",
+  "listings.unitTypes.3 BR": "3 Bedroom",
+  "listings.unitTypes.4 BR": "4 Bedroom",
+  "listings.unitTypes.5 BR": "5 Bedroom",
+  "listings.unitTypes.SRO": "SRO (single-room occupancy)",
```
akegan, line 72: SRO should just be SRO for now. We don't have room for the explanation in the parentheses in the tables we have

app/javascript/__tests__/util/languageUtil.test.ts
```diff
@@ -160,4 +161,13 @@ describe("languageUtil", () => {
       expect(getReservedCommunityType("New Type")).toBe("New Type")
     })
   })
+  describe("emptyIfNotTranslated", () => {
+    it("returns a translated string", () => {
+      expect(emptyIfNotTranslated("listings.unitTypes.2 BR")).toBe("2 Bedroom")
+    })
+
+    it("returns an empty string when no translation", () => {
+      expect(emptyIfNotTranslated("listings.unitTypes.500 BR")).toBe("")
```
akegan, line 170: We want the behavior to be to show whatever salesforce is sending if we don't have a translation for the value. So we don't require a code change if a new type of unit is added

akegan (changes requested):
Couple of minor changes.

app/assets/json/translations/react/en.json
jacksongonzales: right, @christine-sfg and I agreed that it's funky. if this works as a first pass we could dogear it as something to come back to

app/javascript/pages/RentDirectory.tsx
```diff
@@ -38,8 +38,7 @@ const getForRentSummaryTable = (listing: RailsRentalListing) =>
     .filter((summary) => !!summary.unitType)
     .map((summary) => ({
       unitType: {
-        cellText:
-          emptyIfNotTranslated(`listings.unitTypes.${summary.unitType}`) || summary.unitType,
+        cellText: emptyIfNotTranslated(`listings.unitTypes.${summary.unitType}`, summary.unitType),
```
akegan, line 41: this looks like a typo maybe, needs to be renamed to defaultIfNotTranslated?

jacksongonzales: oop, thanks for catching. still getting the hang of vscode and things like...saving files 😬

jacksongonzales:
@akegan would you mind sending me an invite to CircleCI? I'm guessing I can't rerun the failing tests with no account

akegan (approved):
this looks good to me. Feel free to squash merge once tests pass (I just kicked them off again)

PR #1579, opened 2022-02-09
Listing Details Skeleton DAH-990

emilyjablonski:
[DAH-990](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-990)

Creates an initial skeleton of the Listing Details page in React, and has it toggleable on the `react=true` URL param.

This was trickier than anticipated due to remaining work in Bloom components to pull out dependencies on our data model. The content areas should be aligned as expected with 75% of the content rendering as expected but without the appropriate data. The rest will render appropriately after the Bloom work, which will allow us to render the correct data instead of mock data.


app/javascript/api/listingsApiService.ts
```diff
@@ -56,6 +56,9 @@ export const getListings = async (
         ({ data }) => data.listings
       )
 
+export const getListing = async (listingId?: string): Promise<ListingsResponse> =>
```
akegan, line 59: my guess is that we'll want a separate listingApiService file because we'll be making a few other listing-related requests

config/routes.rb
```diff
@@ -83,9 +83,7 @@ def matches?(_)
   get '(:lang)/listings/for-rent' => 'directory#rent', lang: /(en|es|zh|tl)/
   get '(:lang)/listings/for-sale' => 'directory#sale', lang: /(en|es|zh|tl)/
   # TODO: Paths on Bloom needs to be configurable
-  get '(:lang)/listing/:id', lang: /(en|es|zh|tl)/, to: redirect { |params, _|
-    "#{params[:lang] ? params[:lang] + '/' : ''}/listings/#{params[:id]}"
-  }
+  get '(:lang)/listings/:id' => 'listing#index', lang: /(en|es|zh|tl)/
```
akegan, line 86: curious about what this change is.

app/javascript/pages/ListingDetail.tsx
```diff
@@ -0,0 +1,222 @@
+import React, { useContext, useEffect } from "react"
```
akegan, line 1: I'm wondering about what makes the best model for organizing files under the pages directory.

Right now we have:
- ListingDirectory
  - BuyHeader
  - DirectoryHelpers
  - GenericDirectory
  - etc
- BuyDirectory
- RentDirectory
- HomePage
- etc

Is that how you have it setup in bloom? with all top-level pages in pages/TopLevelPage.tsx, then directories with the same name that have related components and utility files?

Or would it make more sense to have:
- ListingDirectory
  - BuyDirectory
  - RentDirectory
  - BuyHeader
  - DirectoryHelpers
  - GenericDirectory
  - etc
- HomePage
- etc

cc @christine-sfg

akegan (commented):
Looks like a good start! just one suggestion and a couple of questions

README.md
```diff
@@ -124,6 +123,7 @@ We have flags for each chunk of the rewrite we release. These will set those pag
 
 - HOME_PAGE_REACT='true'
 - DIRECTORY_PAGE_REACT='true'
+- LISTING_DETAIL_PAGE_REACT='true'
```
christine-sfg, line 126: This doesn't match AC in DAH-990 which notes `REACT_RENTAL_LISTING_DETAILS`, but this one seems to line up with the other existing vars. Should we update the AC?

config/routes.rb
emilyjablonski: The page doesn't load / work with the previous redirect, but I'm honestly not sure what the exact issue was even fixing the `listings` --> `listing` typo

akegan: ok interesting

app/javascript/pages/ListingDetail.tsx
emilyjablonski: Updated the file structure based more closely on best practice, here is what it looks like now:

- components/ (reusable across pages)

- components/ (reusable across pages)
- modules/ (site sections like listings, applications, user management, all helpers and UI pieces live here)
- pages/ (only directly navigable pages, under directories that match their URL, lower case file names to differentiate from components)

akegan:
@emilyjablonski that file structure looks better to me. What do you think @christine-sfg? 

I do also think we should table this pr for a couple of days to focus on getting the directory page on

app/javascript/pages/ListingDetail.tsx
christine-sfg: Yes! This is much better!

akegan:
@emilyjablonski I think this is pretty much ready to go now that we've pulled out our release branch. Only bummer is that there are now merge conflicts. Can you resolve them when you get a chance and let us know when they're resolved?

emilyjablonski:
@akegan Sure thing!

christine-sfg (approved):

PR #1588, opened 2022-02-15
Translate units in habitat listing dir, fix income range styling

akegan:
DAH-1064

This fixes issues Yindi found while testing:
- Unit header not translated for habtiat
- Income ranges going onto two columns in habitat listing in chinese
- Reduced the gap on the ie11 page

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -33,3 +33,7 @@ body .goog-te-banner-frame {
   color: inherit;
   font-weight: 700;
 }
+
+.habitat-income-list li {
+  float: none;
+}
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.css
```diff
@@ -32,4 +32,8 @@ body .goog-te-banner-frame {
   text-decoration: underline;
   color: inherit;
   font-weight: 700;
+}
+
+.habitat-income-list li {
+  float: none;
```
akegan, line 38: float:left seemed to be the culprit, but I didn't want to change the inline-list style without doing more research into what else might be affected, so I just added a class instead.

christine-sfg (approved):

PR #1595, opened 2022-02-23
fix: add optional on general summary

emilyjablonski:
Allows for a listing with a null general unit summary to still render. It will just not render a unit table at all under the ListingCard.


jacksongonzales (approved):

christine-sfg:
We can close this one. We ended up fixing via https://github.com/SFDigitalServices/sf-dahlia-web/pull/1599

PR #1597, opened 2022-02-25
fix: check if general summary exists, otherwise use reserved

emilyjablonski:
DAH-1102

For the unit summary table, first use the general unit summary, then the reserved unit summary, then show no table.


christine-sfg:
fyi, i renamed the branch as per hotfix process and it auto closed this pr. created https://github.com/SFDigitalServices/sf-dahlia-web/pull/1598

PR #1600, opened 2022-02-25
DAH-1101 Prevent users with South SF addresses from getting NRHP preference

christine-sfg:
DAH-1101

-Explicitly check for household members to have city 'San Francisco' to avoid false positives for NRHP preference
-Misc housekeeping/docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,10 +38,11 @@ public/assets/
 public/packs
 public/packs-test
 
-# VS Code
+# IDE
 TAGS
 .vscode/*
 !.vscode-default/*
+.idea/*
 
 # Byebug command history
 .byebug_history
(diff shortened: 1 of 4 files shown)


emilyjablonski:
@christine-sfg I'm a little confused by the bug, the PR description says it needs to check household members and the ticket says it needs to check the applicant?

christine-sfg:
@emilyjablonski good question. From what i saw, the [household members list](https://github.com/SFDigitalServices/sf-dahlia-web/blob/55e359aeb3a3c9e3700a735d13df4cdf082d132c/app/assets/javascripts/short-form/ShortFormApplicationService.js.coffee#L522) also includes the applicant. I think we want to check all addresses included in the application. I will update the jira to make this clearer.

emilyjablonski:
@christine-sfg Is it true that an applicant who has an SF address and also has household members without them can get the preference?

christine-sfg:
@emilyjablonski Yes. From testing on full, the opposite is true as well. As long as at least one member of the household has a qualifying address, then the application should be prompted for NRHP preference.

jacksongonzales (approved):
Looks good after testing scenarios you and emily described above + the expected preference behavior 🎉

PR #1601, opened 2022-03-02
DAH-1077 give time for invalid listing id lookup

christine-sfg:
- The "Attempting to go to a listing page using an invalid ID" test has been flaky. Looks like it can sometimes take ~1 sec for the salesforce lookup and miss, along with the redirects so adding some wait time before checking for the welcome component
- Setup test stats in circleci for new e2e tests

### DIFF
--- a/spec/e2e/features/listings/listing.feature
+++ b/spec/e2e/features/listings/listing.feature
@@ -1,24 +1,26 @@
 Feature: Listings
-    As a web user
-    I should be able to view a listing
+  As a web user
+  I should be able to view a listing
 
-    Scenario: Attempting to go to a listing page using an invalid ID
-      Given I try to go to a listing page with an invalid ID
-      Then I should be redirected to the welcome page
+  Scenario: Attempting to go to a listing page using an invalid ID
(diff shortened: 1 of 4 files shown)


spec/e2e/features/listings/listing.feature
```diff
@@ -1,24 +1,26 @@
...
+  As a web user
+  I should be able to view a listing
 
-    Scenario: Attempting to go to a listing page using an invalid ID
-      Given I try to go to a listing page with an invalid ID
-      Then I should be redirected to the welcome page
+  Scenario: Attempting to go to a listing page using an invalid ID
+    Given I try to go to a listing page with an invalid ID
+    #give time for the lookup failure and redirect
+    And I wait "5" seconds
```
jacksongonzales, line 8: do we have any control over the lookup request, like somewhere we could do some kind of `await` on a response?

christine-sfg: great question. we can't wait for the ajax request, but we can add a wait for the success condition. will try this another way.

jacksongonzales (approved):

PR #1603, opened 2022-03-12
DAH-1103 Add content and page level error boundaries to prevent blank pages

christine-sfg:
DAH-1103

As of React 16, errors that were not caught by any error boundary will result in unmounting of the whole React component tree. See [Error boundaries doc](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html)

Add content and page level error boundaries to display some friendly error messaging instead.

1) Main content level failures -  Visit an affected page and you should see the header and footer with a fallback ui as the main content
image

2) Page level failures - user is redirected to the static error 500 page
image


app/javascript/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,67 @@
...
+interface Props {
+  boundaryScope: BoundaryScope
+  children: ReactNode
+}
+
+interface State {
+  hasError: boolean
+}
+
+class ErrorBoundary extends React.Component<Props, State> {
```
emilyjablonski, line 27: If possible we should use exclusively functional components

christine-sfg: I agree. Unfortunately, react currently only supports error boundaries using class components.

emilyjablonski (approved):
LGTM!

jacksongonzales (approved):

PR #1611, opened 2022-04-01
DAH-1121 Add images to section headers

jacksongonzales:
DAH-1121

Adds images!

fyi, I am leaving the `imageAlt` prop an empty string since I think these register as [W3C defined "decorative images"](https://www.w3.org/WAI/tutorials/images/decorative/): 
> In these cases, a null (empty) alt text should be provided (alt="") so that they can be ignored by assistive technologies, such as screen readers.


app/javascript/pages/listings/listing-detail.tsx
```diff
@@ -44,10 +44,19 @@ const ListingDetail = () => {
...
+              />
+              <ListingDetailsFeatures
+                listing={listing}
+                imageSrc={getAssetPath("listing-features.svg")}
+              />
+              <ListingDetailsNeighborhood imageSrc={getAssetPath("listing-neighborhood.svg")} />
+              <ListingDetailsAdditionalInformation
+                listing={listing}
+                imageSrc={getAssetPath("listing-legal.svg")}
+              />
```
christine-sfg, line 59: Do we think the images may change with the usage of the components? If not, we may want to just put the image info in the components themselves.

jacksongonzales: I wasn't sure since the module components seem pretty specific to a set of content atm, and adding the context for `getAssetPath` on each component feels a bit heavy

christine-sfg:
@jacksongonzales thanks for linking the w3c doc. good to know. i had thought adding the alt attribute was ideal in all cases - if the image wasn't providing benefit or context to the user experience then why not leave it out altogether?

jacksongonzales:
> @jacksongonzales thanks for linking the w3c doc. good to know. i had thought adding the alt attribute was ideal in all cases - if the image wasn't providing benefit or context to the user experience then why not leave it out altogether?

yes! this is def a conversation we had over in the design system. I'm not sure we have a firm answer, since, along with all the concerns with decorative images, we're still compelled to have a visual brand identity or 'look and feel.' So it's probably a case-by-case thing, which is the worst answer 😅

app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
```diff
@@ -1,18 +1,23 @@
...
   return (
     <ListingDetailItem
       imageAlt={"Image Alt"}
-      imageSrc={""}
-      title={"Features"}
-      subtitle={"Amenities, unit details and additional fees"}
+      imageSrc={imageSrc}
+      title={t("listings.features.header")}
+      subtitle={
+        isSale(listing)
```
emilyjablonski, line 17: 🥳

emilyjablonski:
Curious why this PR didn't deploy a preview, @christine-sfg do you have an idea as to why?

app/javascript/modules/listingDetails/ListingDetailsAdditionalInformation.tsx
```diff
@@ -6,17 +6,19 @@ import { stripMostTags } from "../../util/filterUtil"
...
 }: ListingDetailsAdditionalInformationProps) => {
   return (
     <ListingDetailItem
       imageAlt={""}
-      imageSrc={""}
-      title={"Additional information"}
-      subtitle={"Required documents and selection criteria"}
+      imageSrc={imageSrc}
+      title={t("listings.additionalInformation.header")}
+      subtitle={t("listings.additionalInformation.header")}
```
emilyjablonski, line 21: We will need to use the subHeader string here

emilyjablonski:
Looking good! I have one small code change and one potential code change, and then my only other note is that the images are ever so slightly smaller on this version than on prod. If that is an important update to design, we can make an easy Bloom change 👍

christine-sfg:
> > @jacksongonzales thanks for linking the w3c doc. good to know. i had thought adding the alt attribute was ideal in all cases - if the image wasn't providing benefit or context to the user experience then why not leave it out altogether?
> 
> yes! this is def a conversation we had over in the design system. I'm not sure we have a firm answer, since, along with all the concerns with decorative images, we're still compelled to have a visual brand identity or 'look and feel.' So it's probably a case-by-case thing, which is the worst answer 😅

I wonder if screen readers give users an option to disable decorative image info. I can understand it may get noisy, especially if a site is decorative image heavy, but it would be cool if the user could make that decision for themself whether they wanted that info...on a case by case basis :) Like if it's a new site they are visiting, they may want that info. If it's somewhere they frequent or the alt text hasn't changed for a long time, then no.

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
```diff
@@ -5,20 +5,25 @@ import {
...
+}: ListingDetailsEligibilityProps) => {
   return (
     <ListingDetailItem
       imageAlt={""}
-      imageSrc={""}
-      title={"Eligibility"}
-      subtitle={"Income, occupancy, preferences, subsides"}
+      imageSrc={imageSrc}
+      title={t("listings.eligibility.header")}
+      subtitle={t("listings.eligibility.subheader")}
```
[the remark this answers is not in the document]
christine-sfg: [Looks intentional](https://github.com/SFDigitalServices/sf-dahlia-web/blob/5accc6220f1b28eea3f6545547b4111abdedd6b4/app/assets/javascripts/listings/components/listing/eligibility-section.html.slim#L4-L4), but good point. Seems like it should be included for sales pages too.

christine-sfg:
Looking better on mobile. Can we add the Process image and subheader too? The sections are in a different order for mobile, but that may be out of scope for this story
image

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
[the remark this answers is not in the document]
jacksongonzales: went with what emily s and yindi recommended [here](https://sfdigitalservices.slack.com/archives/CCV0019TN/p1649108494038529)

jacksongonzales:
> Looking good! I have one small code change and one potential code change, and then my only other note is that the images are ever so slightly smaller on this version than on prod. If that is an important update to design, we can make an easy Bloom change 👍

@emilyjablonski I'm fine with trucking on and noting this detail; I suspect we'll have some kind of final design review before the page goes live that could put together any little design tweaks we need

> I wonder if screen readers give users an option to disable decorative image info. I can understand it may get noisy, especially if a site is decorative image heavy, but it would be cool if the user could make that decision for themself whether they wanted that info...on a case by case basis :) Like if it's a new site they are visiting, they may want that info. If it's somewhere they frequent or the alt text hasn't changed for a long time, then no.

@christine-sfg [aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) is usually the way to go on the dev side, but this raises good questions!

christine-sfg:
> Curious why this PR didn't deploy a preview, @christine-sfg do you have an idea as to why?

We kept hitting the max # of apps in heroku, but that was fixed around noon today and i manually created this app

christine-sfg (approved):

PR #1612, opened 2022-04-01
DAH-1123 Lottery Preference ordinal info cards

christine-sfg:
DAH-1123

1) Adding the ordinal preference cards for both rental and sales listings
image

2) Threw in some extra work for the Before Applying list (displays on the sale details and directory pages) because it also has circled numbers 😂  and i did too much work before realizing i was working on the wrong thing
image

3) Passing in entire listing objects for the `ListingDetailsApplicationDate` tests and removed the eslint-disable's

4) Learned that the listing preferences component needs more info so added in a call to the `/api/v1/listings/${listingId}/preferences` api


app/javascript/__tests__/data/listing-rental-closed.ts
```diff
@@ -0,0 +1,301 @@
...
+    reserved: null,
+    general: [
+      {
+        unitType: "1 BR",
+        totalUnits: 1,
+        minSquareFt: null,
+        minRentalMinIncome: 5255,
+        // "minPriceWithParking": null,
+        // "minPriceWithoutParking": null,
+        minPercentIncome: null,
```
christine-sfg, line 14: I copied these test data files from api responses on full. Ts is mad because not all of these fields are defined in the type. Any thoughts on adding these all to the appropriate types or leaving these commented out? Some of the offenders are odd e.g. the `listing-sale-open.ts` response has `minRentalMinIncome`. The other option is asking whether we can suppress inappropriate fields on the Salesforce side

app/javascript/__tests__/modules/listingDetails/ListingDetailsPreferences.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const getPreferencesMock = getPreferences as jest.MockedFunction<typeof getPreferences>
+    getPreferencesMock.mockReturnValue(Promise.resolve(defaultPreferences))
+
+    const tree = renderer.create(<ListingDetailsPreferences listingID={"test"} />)
+
+    // wait for state changes
+    await act(() => new Promise((resolve) => setTimeout(resolve)))
+
+    expect(tree.toJSON()).toMatchSnapshot()
+  })
```
christine-sfg, line 23: was debating whether to add in [msw](https://mswjs.io/) but opted for this shorter path for now

app/assets/json/translations/react/en.json
```diff
@@ -1,174 +1,217 @@
 {
```
emilyjablonski, line 1: I don't think we want the wider space here so that the formatting matches all of our other files

app/assets/json/translations/react/en.json
```diff
@@ -1,174 +1,217 @@
...
+    "footer.giveFeedback": "Give Feedback",
+    "footer.inPartnershipWith": "in partnership with<br /><a className='text-white' href='%{sfdsUrl}' target='_blank'>San Francisco Digital Services</a><br /><a className='text-white' href='%{mayorUrl}' target='_blank'>Mayor's Office of Civic Innovation</a>",
+    "footer.privacyPolicy": "Privacy Policy",
+    "idleTimeout.sessionExpired": "Your session has expired.",
+    "idleTimeout.sessionInactivity": "To protect your identity, your session will expire in one minute due to inactivity. You will lose any unsaved information you've entered into the application for this listing if you choose not to respond.",
+    "idleTimeout.sessionInactivityLoggedIn": "To protect your identity, your session will expire in one minute due to inactivity. You will lose any unsaved information and be logged out if you choose not to respond.",
+    "idleTimeout.stayLoggedIn": "Stay logged in?",
+    "label.clear": "Clear",
+    "label.editEligbility": "Edit Eligibility",
+    "label.read_more": "Read More",
```
emilyjablonski, line 20: This string on prod is all lowercase `read more` - I personally think it looks better uppercase like this, just wanted to make sure this was intentional

app/javascript/__tests__/data/listing-rental-closed.ts
emilyjablonski: If I'm understanding correctly I like the idea of adding these to the type since I think we will need them on the type eventually so that we can be confident the API response is what we expect it to be

app/assets/json/translations/react/en.json
christine-sfg: This is driven by [.editorconfig](https://github.com/SFDigitalServices/sf-dahlia-web/blob/5accc6220f1b28eea3f6545547b4111abdedd6b4/.editorconfig#L15-L15) Do we want to change this configuration?

app/assets/json/translations/react/en.json
christine-sfg: It was intentional, but you're right we should limit changes. I will revert to `read more`

In terms of camelCase vs snake_case...I wasn't sure if there was some magic happening to map angular keys to react so I was keeping the same keys, but I think you're also right here and we should remain consistent. Let's go with camel case fyi @jacksongonzales

christine-sfg:
> I might be missing some sort of setup, but at the moment for me locally on the react listing detail page, the page doesn't load / spins indefinitely on a different language.

I actually had a question about that. Perhaps related to google translate? That issue is not unique to this pr (exists on full now) and I'm not sure that we've tested that since the listing details skeleton was created.

app/assets/json/translations/react/en.json
emilyjablonski: There's no angular-react mapping so we can name them whatever we want :)

app/javascript/modules/listingDetails/ListingDetailsPreferences.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <ListSection title={t("listings.lottery.title")} subtitle={t("listings.lottery.preferences")}>
+      <LoadingOverlay isLoading={preferences.length === 0}>
+        <>
+          <PreferencesList
+            listingPreferences={preferences.map((preference, index) => {
+              const links = []
+              if (preference.readMoreUrl) {
+                links.push({ title: t("label.read_more"), url: preference.readMoreUrl })
```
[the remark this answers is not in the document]
christine-sfg: Bloom change done. Ignore what i said about expandable text ☺️

jacksongonzales (approved):
no more feedback from me! looks great, lgtm

PR #1616, opened 2022-04-15
DAH-1146 Update text for Alice Griffith Preference to support rule changes

christine-sfg:
DAH-1146

-This preference will no longer be date bound. Update text accordingly. 
-Description text will also be updated in Salesforce for new AG Listing Preferences


christine-sfg:
had to rename branch for hotfix. will open another pr

PR #1617, opened 2022-04-16
DAH-1126 Add spacing between error and checkbox

christine-sfg:
DAH-1126

Design feedback is to add some spacing between the error message and checkbox below it


emilyjablonski (approved):
LGTM

PR #1618, opened 2022-04-18
DAH-976 uptake a11y fixes for homepage

christine-sfg:
DAH-976

Uptake fixes for the following 3 a11y issues on the homepage
image


emilyjablonski:
I'm not seeing the newly added nav aria label come through when inspecting the page

christine-sfg:
@emilyjablonski The `aria-label` attribute is on the `<nav>` element 
image

jacksongonzales (approved):

PR #1619, opened 2022-04-18
DAH-1146 update alice griffith text

christine-sfg:
[DAH-1146](https://sfgovdt.jira.com/browse/DAH-1146)

-This preference will no longer be date bound. Update text accordingly.
-Description text will also be updated in Salesforce for new AG Listing Preferences


christine-sfg:
We only want to deploy the text changes. Closing this and will try again.

PR #1621, opened 2022-04-18
DAH-1134 finish additional info section

jacksongonzales:
This PR:
- Adds text truncation
  - After some exploration, I found that the approach necessary for our weird use-case (of truncating a string containing HTML tags before injecting the string into raw HTML) was going to be overcomplicated to reconcile with the Bloom component. Basically, the Bloom component works well for a string that is in its final state, i.e. not going to be parsed as a separate type of entity, which is what we need to do with the Salesforce data we're pulling.

### DIFF
--- a/app/javascript/__tests__/components/TextTruncate.test.tsx
+++ b/app/javascript/__tests__/components/TextTruncate.test.tsx
@@ -1,0 +1,14 @@
+import React from "react"
+import { TextTruncate } from "../../components/TextTruncate"
+import { renderAndLoadAsync } from "../__util__/renderUtils"
+
+describe("TextTruncate", () => {
+  it("truncates a string and adds an ellipsis when longer than 400 characters", async () => {
+    const content =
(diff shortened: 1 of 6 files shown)


app/javascript/components/TextTruncate.tsx
```diff
@@ -0,0 +1,39 @@
...
+          onClick={() => setIsExpanded(!isExpanded)}
+          onKeyPress={() => setIsExpanded(!isExpanded)}
+          role="button"
+          tabIndex={0}
+        >
+          {isExpanded ? t("label.less") : t("label.more")}
+        </span>
+      )}
+    </div>
+  )
```
christine-sfg, line 38: Is it possible to add a unit test for this?

christine-sfg:
@jacksongonzales It's working for 'Special Notes' which has alot of text, but for 'Important Program Rules' there is a 'More' link, but no additional text displays after clicking.

image

christine-sfg:
'Important Program Rules' is fixed now. No longer seeing a 'More' link on that section

app/javascript/__tests__/components/TextTruncate.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+    const { getByText } = await renderAndLoadAsync(<TextTruncate text={content} />)
+
+    // check for presence of an ellipsis
+    getByText(/\.\.\./)
+  })
```
christine-sfg, line 13: Thanks for adding the test! Another scenario to test would be to make sure it doesn't unnecessarily truncate.

christine-sfg (approved):

app/javascript/__tests__/components/TextTruncate.test.tsx
jacksongonzales: weirdly this is the scenario that was hard to configure 😭 maybe we can work on getting that working another time

PR #1622, opened 2022-04-19
DAH-1126 fix link text

christine-sfg:
DAH-1126

Quick fix for the link text. It had 'email us at' twice. Design would like it to be part of the link.

image


jacksongonzales (approved):
🎉

PR #1623, opened 2022-04-21
DAH-991 uptake new category table component for pricing tables

emilyjablonski:
[DAH-991](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-991)

Uptakes the new pricing table for the details page. We still have work to do on the pricing table API, so I just added some mock data so we can see it rendered.


christine-sfg:
Looks great! Noticed 2 things when testing
1) On mobile the accordion close icon is a hamburger instead of an x
image

2) There is some excessive margin spacing on the columns that causes unnecessary wrapping. This could be more of an issue with longer languages such as Filipino.
image

emilyjablonski:
@christine-sfg I think these would be Bloom changes (a pretty small PR, but still Bloom), would you prefer to merge this and iterate or wait for another Bloom PR?

christine-sfg:
@emilyjablonski If it's a trivial bloom change, let's go ahead and wait for it, assuming it would be coming soon.

emilyjablonski:
@christine-sfg The table fix around the icon is turning out to be weirdly challenging, I might recommend merging this and working iteratively.

PR #1627, opened 2022-04-27
add github action for creating/removing review apps

jacksongonzales:
following the solution Shawn found [here](https://github.com/fastruby/manage-heroku-review-app), an attempt at a workaround to bring back review apps


app/javascript/pages/index.tsx
```diff
@@ -34,6 +34,7 @@ const HomePage = (_props: HomePageProps) => {
         secondaryButtonLink={getSaleDirectoryPath()}
         secondaryButtonTitle={t("welcome.seeSaleListings")}
       />
+      <div>hi!</div>
```
jacksongonzales, line 37: note to self to remove this before merging 😅

christine-sfg:
lgtm! thank you @jacksongonzales and @shawnbot! 🎉  

will approve after removing the test markup.

christine-sfg (approved):

jacksongonzales:
> Yay, I'm so glad this works! Just one quick note because I made this mistake:
> 
> ⚠️ **Removing the Heroku GitHub integration will disable review apps!** I was getting annoyed by Heroku's banner about the integration and removed it, only to discover that I couldn't then re-enable review apps, and that the ones that we had were shut down. So, no matter how annoying you may find that banner, don't do it until Heroku gets it working again:
> 
> image

*sigh* so misleading, it makes it sound like it'll fix the problem we're after! glad I didn't follow it, I was certainly tempted

PR #1628, opened 2022-04-28
DAH-1147 Listing Details Page in All Languages

christine-sfg:
DAH-1147 

-Fixes issue where listing details page would not load for non-english languages
-Manually add existing translations...fun!
-Cleanup to convert any snake case keys to use camel case


app/javascript/pages/listings/listing-detail.tsx
```diff
@@ -27,7 +28,8 @@ const ListingDetail = () => {
   const [listing, setListing] = useState<RailsListing>(null)
 
   useEffect(() => {
-    void getListing(router.pathname.split("/")[2]).then((listing: RailsListing) => {
+    const path = getPathWithoutLanguagePrefix(router.pathname)
+    void getListing(path.split("/")[2]).then((listing: RailsListing) => {
```
christine-sfg, line 32: This was the root cause of the issue. Not google translate after all.

emilyjablonski (approved):

PR #1632, opened 2022-05-04
DAH-1156 Prevent overflow of footer links

emilyjablonski:
[DAH-1156](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1156)

Uptakes latest Bloom which includes a fix to allow for overflowing links in the `SiteFooter` component. Due to the breaking changes in the AdditionalFees component (which has a PR out waiting on a string question), I have just commented that out for now. Also adds the new stylesheets coming in from the Bloom theming work. To test, ensure the footer links do not overflow inappropriately on desktop and mobile screens.

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsApplicationDate.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsApplicationDate.test.tsx.snap
@@ -35,7 +35,7 @@ exports[`ListingDetailsApplicationDate displays Applications Closed when listing
     data-test-id={null}
   >
     <svg
-      fill="#E41D3D"
+      fill="var(--bloom-color-red-700)"
       viewBox="0 0 32 32"
     >
       <path
--- a/app/javascript/components/base.scss
+++ b/app/javascript/components/base.scss
@@ -57,7 +57,6 @@ $text: $black;
 $border: 2px solid #ccc; // not sure how this gets used -JW
 $radius-large: 8px;
 $radius: 4px;
-$desktop: $tailwind-screens-md;
 $speed: 86ms;
 $easing: ease-out;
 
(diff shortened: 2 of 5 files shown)


app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsApplicationDate.test.tsx.snap
```diff
@@ -35,7 +35,7 @@ exports[`ListingDetailsApplicationDate displays Applications Closed when listing
     data-test-id={null}
   >
     <svg
-      fill="#E41D3D"
+      fill="var(--bloom-color-red-700)"
```
emilyjablonski, line 38: New theming starting to come through 😁

christine-sfg:
I noticed the links are hard to read in Filipino, but actually are an improvement over what's currently in prod imo since at least each group of text stays together. I think we may still want to check with @yindipei though if they have any suggestions on wrapping or placement.
image

yindipei:
I agree with Christine. Ideally, words that make up one link shouldn't wrap. In the case of Filipino, `Pagwawaksi ng Pananagutan (Disclaimer)` should all be on the same line. 

Additionally, for `© Lungsod at County ng San Francisco`, it looks like there's enough room to put `Francisco` up with `San`. I would prefer if that whole string on the left side was on one line.

christine-sfg:
Do you mean the wrapping of the left column (copyright)? Here's a review app https://dahlia-web-full-pr-1632.herokuapp.com/tl/listings/a0W4U00000JqTLIUA3?react=true

emilyjablonski:
@christine-sfg This is ready for your eyes again! For some reason it looks like the deploy preview did not update after the last change

christine-sfg:
Bummer the workaround isn't as smart as the normal review apps deployments. I actually had to add the `heroku-review-app` label earlier to get it created.

Saved time this time around and just tested on my local with the latest. Looks great!

christine-sfg (approved):

PR #1635, opened 2022-05-10
DAH-1136 Uptake new waitlist component

emilyjablonski:
[DAH-1136](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1136)

Uptakes the waitlist component, which has been renamed in Bloom to `QuantityRowSection` so that it can be abstracted to render any number of rows, in any context.

Note: The review app does not reflect the latest state.

The four states:
a0W4U00000JqTLIUA3?react=true (Open waitlist and available units, so has 2 rows)
a0W4U00000KnAMSUA3?react=true (Open waitlist and no available units, so has 3 rows)
a0W4U00000KnLRRUA3?react=true (Has a waitlist, but it is in pending lottery state, so it should not show)
a0W4U00000KRHeEUAX?react=true (Has no waitlist, so it should not show)


app/assets/json/translations/react/en.json
```diff
@@ -47,9 +47,14 @@
   "listings.atTotalIncome": "at <span class='eligibility-subtext'>$%{income}</span> %{per}",
   "listings.availableUnits": "Available Units",
   "listings.availableUnitsAndOpenWaitlist": "Available Units & Open Waitlist",
+  "listings.availableUnitsAndWaitlist": "Available units and waitlist",
+  "listings.availableUnitsAndWaitlistDescription": "Once lottery winners fill all available units, additional lottery winners will be placed on the waitlist for %{number} %{unit}",
+  "listings.availableUnitsAndWaitlistDescription.pluralUnits": "units",
+  "listings.availableUnitsAndWaitlistDescription.singularUnit": "unit",
```
christine-sfg, line 53: Concatenating text props is usually not a good idea. This could be simplified to 2 text props instead of 3, but the 0 units scenario is also awkward because with the current logic it looks like the text would be `...additional lottery winners will be placed on the waitlist for 0 unit`. Is this a possible scenario?

I am inclined to just remove the number of units from this text
image

emilyjablonski: This change was to solve an existing `1 units` bug - open to removing the units text if Yindi/Emily are!

christine-sfg: @yindipei @emilyshore  is it ok to remove the number of units within the paragraph? see above screenshot

jacksongonzales:
ugh, getting the white-screen of death when trying to review locally 😞 

otherwise, this looks good. I'll keep troubleshooting this issue/see if i can get the review app to update

Screen Shot 2022-05-11 at 2 05 51 PM

app/assets/json/translations/react/en.json
yindipei: Assuming this is the only instance of this situation, we will not be compromising the UX if we remove the italicized part. My answer is Yes! It is okay to remove the number of units within the paragraph.

app/javascript/modules/listingDetailsAside/ListingDetailsApply.tsx
```diff
@@ -1,22 +1,102 @@
...
+  How to Apply
+  Need Help
+*/
 export const ListingDetailsApply = ({ listing }: ListingDetailsApplyProps) => {
+  const descriptionValues = {
+    number: listing.Total_number_of_building_units,
+    unit:
+      listing.Total_number_of_building_units > 1
+        ? t("listings.availableUnitsAndWaitlistDescription.pluralUnits")
+        : t("listings.availableUnitsAndWaitlistDescription.singularUnit"),
```
christine-sfg, line 22: Is this still being used?

emilyjablonski: Good catch thank you! Updated

jacksongonzales (approved):
not sure if this got overwritten in the merge, but pretty sure we can uncomment line 32 of `ListingDetailsAside` and import the component 🙂  after that LGTM!

app/javascript/modules/listingDetailsAside/ListingDetailsApply.tsx
```diff
@@ -1,22 +1,85 @@
...
 export interface ListingDetailsApplyProps {
   listing: RailsListing
 }
 
+/*
+  TODO: this component will also include
+  How to Apply
+  Need Help
+*/
 export const ListingDetailsApply = ({ listing }: ListingDetailsApplyProps) => {
```
christine-sfg, line 16: can we add some unit tests for this component?

christine-sfg (approved):

PR #1645, opened 2022-05-31
DAH-1157 add styles to match angular modal and make scrollable

jacksongonzales:
[DAH-1157](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1157)

1/2 of the changes necessary to make the Lottery Results modal scrollable and approximate angular styles. This also makes the header and listing name stick as you scroll.

| Before | After! |
| --- | --- |
Screen Shot 2022-05-31 at 9 21 10 AM | Screen Shot 2022-05-31 at 9 21 42 AM

We need Bloom's [#2769](https://github.com/bloom-housing/bloom/pull/2769) to be merged and uptaken on this side before merging this, though if you really wanna get hacky you can just copy the changes from the Bloom PR into a local instance of this PR 😬


kramduckner:
It's looking so good @jacksongonzales! What do you think about removing that bottom margin from the 'Read Housing Preferences' button? No sweat if it's a hassle.
image

PR #1647, opened 2022-06-01
DAH-1167 Prevent Apt or Unit # Duplication

christine-sfg:
DAH-1167

Add validation to prevent submission of duplicate apartment or unit #. This is similar to the existing validation which prevents users from using a PO Box.
image


app/services/address_validation_service.rb
```diff
@@ -30,25 +30,38 @@ def po_box?(validation)
...
+      validation.street1.end_with?(validation.street2)
+  end
+
   def invalid?
     # we don't treat timeouts as "invalid" since we want to allow them to proceed
     Rails.logger.warn('Address validation: Easypost request timed out') if timeout?
 
     # we do not accept PO Boxes
-    if @validation.present? && po_box?(@validation)
+    if po_box?(@validation)
```
christine-sfg, line 52: moved this into the `po_box` function because rubocop said this method was getting to complex

emilyjablonski (approved):

app/services/address_validation_service.rb
```diff
@@ -30,25 +30,38 @@ def po_box?(validation)
...
     validation.street1.match(/P\.?\s*O\.?\s*BOX/i) || (
       !validation.verifications.delivery.success &&
       validation.verifications.delivery.errors[0].code == 'E.BOX_NUMBER.INVALID'
     )
   end
 
+  def duplicate_unit?(validation)
+    # Check for duplicate apt/unit no on street lines 1 and 2
+    validation.present? && validation.street2.present? &&
+      validation.street1.end_with?(validation.street2)
```
kramduckner, line 44: Is it guaranteed that validation.street1 will exist? What do you think about adding an optional chain operator after validation.street1 (e.g. validation.street1?.end_with...)? I'm also no ruby maestro so maybe that's not even possible.

christine-sfg: Good question. There is front end validation for street1 since it is a required field so unlikely that it would be submitted empty, but I can add just in case.

PR #1648, opened 2022-06-02
DAH-1118 translations for june 22 release

christine-sfg:
DAH-1118

Prepping language files for translations which included
-sorting language keys
-whitespace cleanup
-removing keys no longer used
-copying translations missed from angular to react


emilyjablonski:
@christine-sfg Is there any manual testing you think we should do beyond looking at the diff?

kramduckner (approved):

jacksongonzales (approved):
looks good! +1 emily's question above though

christine-sfg:
@emilyjablonski @jacksongonzales good question. it's quite tedious to find each component with a text change and view in each language and tbh, the text props should just work. If you do want to do some testing, I'd say focus on text props that have parameters. 

That would be this component that displays on the details page for habitat listings
image

PR #1650, opened 2022-06-09
DAH-1152 Add reserved banner

christine-sfg:
DAH-1152
-Banner displays if the Reserved Community Type is one of the following 4 types
  - Accessible Units Only
  - Artist Live/Work
  - Senior
  - Veteran

-Banner does not display if the Reserved Community Type is blank or is 'Habitat for Humanity'
-For desktop, the banner should be the same width as the listing image. For mobile, the banner should be full width.

image

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/ListingDetailsReservedBanner.test.tsx
+++ b/app/javascript/__tests__/modules/listingDetails/ListingDetailsReservedBanner.test.tsx
@@ -9,6 +9,15 @@ describe("ListingDetailsReservedBanner", () => {
 
     expect(tree).toMatchSnapshot()
   })
+
+  it("does not display banner when unknown reserved community type", () => {
+    const tree = renderer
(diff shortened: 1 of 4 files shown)


app/assets/json/translations/react/en.json
```diff
@@ -49,6 +49,10 @@
   "listings.additionalEligibilityRules.title": "Additional Eligibility Rules",
   "listings.additionalInformation.header": "Additional Information",
   "listings.additionalInformation.subheader": "Required documents and selection criteria",
+  "listings.allUnitsReservedFor.Accessible Units Only": "All units reserved for people who need accessibility features",
```
jacksongonzales, line 52: not sure if it's overboard, but it might be a good idea to reference these with the RESERVED_COMMUNITY_TYPES `consts` since we're accessing these dynamically

jacksongonzales (approved):
one comment above, otherwise looks great!

PR #1652, opened 2022-06-10
DAH-50 translate studio + unabbreviate CA and BR

jacksongonzales:
We actually had most of the cases where we were using abbreviations resolved in the React pages, but plz comment if there are spots I missed. 

One we'll have to look out for, when we get to implementing this section in the listing details aside:
Screen Shot 2022-06-10 at 12 42 24 PM

### DIFF
--- a/app/javascript/modules/listings/SharedHelpers.tsx
+++ b/app/javascript/modules/listings/SharedHelpers.tsx
@@ -65,8 +65,12 @@ export const getListingAddressString = (listing: RailsListing) => {
     (listing.Building_Street_Address &&
       listing.Building_City &&
       listing.Building_State &&
-      listing.Building_Zip_Code &&
-      `${listing.Building_Street_Address}, ${listing.Building_City} ${listing.Building_State}, ${listing.Building_Zip_Code}`) ??
+      listing.Building_Zip_Code && (
+        <span>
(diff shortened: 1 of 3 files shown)


app/javascript/modules/listings/SharedHelpers.tsx
```diff
@@ -65,8 +65,12 @@ export const getListingAddressString = (listing: RailsListing) => {
     (listing.Building_Street_Address &&
       listing.Building_City &&
       listing.Building_State &&
-      listing.Building_Zip_Code &&
-      `${listing.Building_Street_Address}, ${listing.Building_City} ${listing.Building_State}, ${listing.Building_Zip_Code}`) ??
+      listing.Building_Zip_Code && (
+        <span>
+          {listing.Building_Street_Address}, {listing.Building_City}{" "}
+          <abbr title="California">{listing.Building_State}</abbr>, {listing.Building_Zip_Code}
```
christine-sfg, line 71: Feels odd to hard code the abbreviation `title` and not the state code. Sounds like we won't ever get a non Cali listing, but i would recommend we either hard code both or neither. So, one option could be to add constants for all the state codes and look up the CA value. I'm not sure if we currently have a scenario where we'd use this elsewhere. Perhaps someone in an applicant's household could have an out of state address.

christine-sfg (approved):

PR #1657, opened 2022-06-14
DAH-1178 ignore Priority Type of Adaptable

christine-sfg:
DAH-1178

Ignore 'Adaptable' priority type as it is not something we should display to applicants at this time


kramduckner:
Just curious - should there be any kind of equivalent change needed in the react version of the code?

emilyjablonski:
@christine-sfg Heads up I am seeing the adaptable piece on the deploy preview (https://dahlia-full.herokuapp.com/listings/a0W4U00000KnHO6UAN?react=false)

And then @kramduckner since the listing details page in React doesn't have the priority units piece added yet, we'll include that change whenever the component is built in React :)

emilyjablonski (approved):

kramduckner (approved):

PR #1665, opened 2022-06-28
DAH-1180 add conditional to pdf lottery results

kramduckner:
[DAH-1180](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1180)

- Add conditional to Lottery Results Footer to account for where the lottery has been run, but results are not yet posted
Screen Shot 2022-06-28 at 9 56 52 PM

### **Review Instructions**

In the React version:
- Find a listing in the state described above and verify that the 'Download Lottery Results PDF' link isn't displaying and instead the 'Complete lottery results will be posted in pdf format on {date}' text is displaying. 
- Find a listing where the lottery has been run and the results have been posted and verify that the 'Download Lottery Results PDF' is displaying and functioning properly.
- Within the state described above, verify that the bottom text  "The PDF contains.." doesn't display Screen Shot 2022-06-28 at 9 55 37 PM

### DIFF
--- a/app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFooter.tsx
+++ b/app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFooter.tsx
@@ -27,11 +27,19 @@ export const ListingDetailsLotterySearchFooter = ({
               date: localizedFormat(listing.Lottery_Results_Date, "LL"),
             })}
           </p>
-          <p className="py-2">
-            <a href={listing.LotteryResultsURL} target="_blank">
-              {t("lottery.lotteryResultsPdfLink")}
-            </a>
-          </p>
+          {listing.lotteryResultsPdfLink ? (
+            <p className="py-2">
+              <a href={listing.LotteryResultsURL} target="_blank">
+                {t("lottery.lotteryResultsPdfLink")}
+              </a>
+            </p>
+          ) : (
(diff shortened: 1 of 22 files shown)


app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFooter.tsx
```diff
@@ -22,21 +22,31 @@ export const ListingDetailsLotterySearchFooter = ({
...
+                  date: localizedFormat(listing.Lottery_Date, "LL"),
+                })}
+              </p>
+            </>
+          ) : (
+            <p className="py-2">
+              {t("lottery.resultsWillBePostedOn", {
+                date: localizedFormat(listing.Lottery_Results_Date, "LL"),
+              })}
+            </p>
```
christine-sfg, line 48: Can we rely on the Lottery_Results_Date being populated? Should we add a check just in case?

kramduckner: let's do it!

kramduckner: @christine-sfg pushed the update

christine-sfg (approved):

PR #1666, opened 2022-06-28
DAH-1155 add support for fallback image on listings

jacksongonzales:
When a listing doesn't provide an `imageURL`, use the fallback image used on angular in similar cases.

I had to add some support for importing static assets into react/typescript modules as well as for Jest support of the same 🥴


app/javascript/modules/listings/SharedHelpers.tsx
```diff
@@ -61,14 +62,16 @@ export const getListingImageCardStatuses = (
...
-})
+export const getImageCardProps = (listing, hasFiltersSet?: boolean) => {
+  return {
+    imageUrl: listing?.imageURL ?? fallbackImg,
+    href: `/listings/${listing.listingID}`,
+    tags: listing.Reserved_community_type
+      ? [{ text: getReservedCommunityType(listing.Reserved_community_type) }]
+      : undefined,
+    statuses: getListingImageCardStatuses(listing, hasFiltersSet),
+  }
```
christine-sfg, line 73: would be good to have unit test for this function

christine-sfg:
Testing on the review app passed. Looks good in mobile and desktop. Just one small ask.

kramduckner (approved):
lgtm!

PR #1671, opened 2022-07-06
DAH-1118 fix zh period

christine-sfg:
DAH-1118

Should be an open circle, not a dot


jacksongonzales (approved):
such amazing, thorough work 😛

PR #1675, opened 2022-07-08
DAH-1190 Cleanup react unit test errors

christine-sfg:
DAH-1190

Cleanup unit test errors 🎉

There's also one fix on the bloom side in https://github.com/bloom-housing/bloom/pull/2874

Logging/output will be so tidy now
image

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -28,7 +28,6 @@ module.exports = {
   transform: {
     "^.+\\.(ts|tsx)$": "ts-jest",
   },
-  setupFiles: ["dotenv/config"],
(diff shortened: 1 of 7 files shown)


jest.config.js
```diff
@@ -28,7 +28,6 @@ module.exports = {
   transform: {
     "^.+\\.(ts|tsx)$": "ts-jest",
   },
-  setupFiles: ["dotenv/config"],
```
christine-sfg, line 31: Doesn't seem like the unit tests need `dotenv` so removing which will get rid of these lovely errors we were seeing for every test 
>    console.error
      Error: ENOENT: no such file or directory, open '.env'

PR #1681, opened 2022-07-16
DAH-1135 handle no preference results

christine-sfg:
DAH-1135

Handle the scenario for a lottery bucket with no preference results (the preference should not display in the preferences list.

Added a 4th scenario to the review instructions


christine-sfg:
recreating off release branch

PR #1682, opened 2022-07-16
DAH-1135 handle the case for no preference results

christine-sfg:
[DAH-1135](https://sfgovdt.jira.com/browse/DAH-1135)

Handle the scenario for a lottery bucket with no preference results (the preference should not display in the preferences list.

Added a 4th scenario to the review instructions


kramduckner (approved):
lgtm

PR #1685, opened 2022-07-20
DAH-1192 Add lottery info component

christine-sfg:
DAH-1192

### DIFF
--- a/app/javascript/modules/listingDetailsLottery/LotteryDetailsLotteryInfo.tsx
+++ b/app/javascript/modules/listingDetailsLottery/LotteryDetailsLotteryInfo.tsx
@@ -1,0 +1,33 @@
+import React from "react"
+import { SidebarBlock, t } from "@bloom-housing/ui-components"
+import { RailsListing } from "../listings/SharedHelpers"
+import { localizedFormat, renderWithInnerHTML } from "../../util/languageUtil"
+import { isLotteryComplete, isOpen } from "../../util/listingUtil"
+
+export interface ListingDetailsLotteryInfoProps {
+  listing: RailsListing
+}
+
+export const ListingDetailsLotteryInfo = ({ listing }: ListingDetailsLotteryInfoProps) => {
+  if (isOpen(listing) || !listing.Lottery_Date || isLotteryComplete(listing)) {
+    return null
+  }
+
+  return (
+    <>
+      <SidebarBlock title={t("label.lottery")}>
+        <p className="flex justify-between mb-4">
+          <span>{localizedFormat(listing.Lottery_Date, "LL")}</span>
+          <span className="font-bold">{localizedFormat(listing.Lottery_Date, "LT")}</span>
+        </p>
+        {listing.Lottery_Venue && renderWithInnerHTML(listing.Lottery_Venue)}
+        {process.env.COVID_UPDATE !== "true" && <p className="mt-4">{t("label.preLotteryInfo")}</p>}
+      </SidebarBlock>
+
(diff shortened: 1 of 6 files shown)


app/javascript/modules/listingDetailsLottery/LotteryDetailsLotteryInfo.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <>
+      <div className="border-b border-gray-400 md:border-b-0">
+        <SidebarBlock title={t("label.lottery")}>
+          <p className="flex justify-between mb-4">
+            <span>{localizedFormat(listing.Lottery_Date, "LL")}</span>
+            <span className="font-bold">{localizedFormat(listing.Lottery_Date, "LT")}</span>
+          </p>
+          {listing.Lottery_Venue && renderWithInnerHTML(listing.Lottery_Venue)}
```
kramduckner, line 24: re: renderWIthInnerHTML - one thing I discovered working with the html coming from salesforce in another ticket - we already have a markdown-to-jsx package that has a Markdown component which can be used instead of dangerouslySetInnerHTML (which renderWithInnerHTML uses). I ended up spending too much time trying to figure out exactly how that Markdown component sanitizes the html, but supposedly that component does some basic sanitization to prevent malicious markup

app/javascript/modules/listingDetailsLottery/LotteryDetailsLotteryInfo.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const ListingDetailsLotteryInfo = ({ listing }: ListingDetailsLotteryInfoProps) => {
+  if (isOpen(listing) || !listing.Lottery_Date || isLotteryComplete(listing)) {
+    return null
+  }
+
+  return (
+    <>
+      <div className="border-b border-gray-400 md:border-b-0">
```
kramduckner, line 18: love the solution to the borders not being the full width!

kramduckner (commented):
Looks great! Curious about your thoughts re: that Markdown component vs renderInnerHTML before approval.

app/javascript/modules/listingDetailsLottery/LotteryDetailsLotteryInfo.tsx
christine-sfg: @kramduckner good question. Throughout the repo both are being used. Reading thru the docs for both options it seems Markdown may be the better approach, which is odd because at first glance one would assume a markdown lib would be for... markdown. I don't know if Salesforce users are aware we interpret this type of content as markdown.  Although it's highly unlikely any of our Salesforce users would introduce malicious content intentionally it's good to avoid the `dangerouslySetInnerHTML` if and when we can.

@jacksongonzales / @emilyjablonski  do you recall a situation where we needed to use `renderWithInnerHTML()` instead of `Markdown`? Perhaps we should add more comments to the function on when to use.

christine-sfg: Discussed during storytime mtg. We can can leave this `renderWithInnerHTML()` for now and will work on standardization in DAH-1229

jacksongonzales, line 24: noticed the text color here differs slightly from the block below/angular — I think we want to wrap this with an element with `class="text-gray-750"` to match

jacksongonzales (approved):
one small fix, otherwise looks great!

PR #1692, opened 2022-08-03
DAH-1090 remove Get Directions link

christine-sfg:
DAH-1090

Remove the 'Get Directions' link at the bottom since google maps already includes a directions link


kramduckner (commented):
I don't see the native google maps 'get directions' functionality on mobile - screen shot is https://dahlia-full.herokuapp.com/listings/a0W4U00000KRHeEUAX?react=true
Screen Shot 2022-08-04 at 5 22 30 PM

If that's reproducible, should we keep the get directions button on mobile widths?

christine-sfg:
> I don't see the native google maps 'get directions' functionality on mobile 
...
> If that's reproducible, should we keep the get directions button on mobile widths?

@kramduckner good question. I'm going to trust that google thinks there's not enough space to show multiple links. User can get to directions by clicking on the available 'View Larger Map' link and then the directions button so I think we're ok.

kramduckner (approved):

kramduckner:
Sounds good! Approved

PR #1693, opened 2022-08-04
DAH-1192 format time according to sf.gov guidelines

christine-sfg:
DAH-1192

PA testing feedback (2 small changes)
-use the same time format as sf.gov. display times in 12 hour clock format (1:00 pm) rather than 24 hour clock (13:00). Use proper spacing and case.
-as per previous convo's, removing uppercase styling from lottery results date


kramduckner (approved):
lgtm!

PR #1694, opened 2022-08-04
HMI occupancy table fixes

jacksongonzales:
per Yindi's comments on [DAH-1173](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1173)


app/assets/json/translations/react/en.json
```diff
@@ -821,9 +821,8 @@
   "listings.includesPriorityUnits": "Includes Priority Units for %{priorities}",
   "listings.includingChildren": "(including <span class='eligibility-subtext'>%{number} %{children}</span> under 6)",
   "listings.incomeExceptions.intro": "People in your household may need <a href='%{url}' target='_blank'>special income calculations</a> if they:",
-  "listings.incomeExceptions.nontaxable": "Receive <span data-tooltip='%{tooltip}' title='%{tooltip}' aria-haspopup='true' class='has-tip'>non-taxable income</span>",
-  "listings.incomeExceptions.nontaxableTooltip": "Non-taxable income might include SSI, SSDI, child support payments, and worker’s compensation benefits.",
-  "listings.incomeExceptions.students": "Are full-time students (but not the primary applicant)",
+  "listings.incomeExceptions.nontaxable": "Receive non-taxable income (non-taxable income might include SSI, SSDI, child support payments, and worker’s compensation benefits).",
```
kramduckner, line 824: should these changes be applied to other languages?

kramduckner (commented):
looking so good! Just that translation comment and this one question - is it ok is the special income calculations link doesn't work in the deployed version?
Screen Shot 2022-08-04 at 7 44 33 PM

app/assets/json/translations/react/en.json
jacksongonzales: oop, that link above should prob not be broken 😅 good catch. for those string changes, they are dropped into the document we have for keeping track of new/updated translations we're gonna order. I assume we'll do a lump PR for filling in those translations once we get them back, so I think better to wait for those than try my bad guess at how they should translate 😄

kramduckner: re: strings Makes sense! Thanks for fixing that link. I'll approve.

kramduckner (approved):

PR #1696, opened 2022-08-12
DAH-1198 lottery results styling

kramduckner:
[DAH-1198](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1198&assignee=628e5a18f2261e00682a98db)

Review Instructions in Ticket


app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFooter.tsx
```diff
@@ -16,7 +17,11 @@ export const ListingDetailsLotterySearchFooter = ({
   return (
     <>
       <div className="bg-gray-100 py-4 text-center">
-        <a href="http://sfmohcd.org/housing-preference-programs" target="_blank">
+        <a
+          className="lottery-results-modal-anchor"
```
jacksongonzales, line 21: do we not have a tailwind utility that approximates this color? not sure we've adopted a team-wide preference, but my own would be for trying tailwind before stylesheets. or maybe this case is more complicated, as this modal proves to be over and over again 🤪

jacksongonzales (approved):
just one comment, otherwise, looks great!!

app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFooter.tsx
kramduckner: Good idea! I found a close color and used a tailwind class in the jsx instead of the sass file. Also replaced another hard coded hex value with a sass var. Thanks @jacksongonzales!

christine-sfg: 👏

PR #1698, opened 2022-08-16
DAH-1163 sale apply e2e

christine-sfg:
DAH-1163

-upgrade cypress from v8 -> v10
-rewrite this test https://github.com/SFDigitalServices/sf-dahlia-web/blob/production/spec/e2e/features/short_form/navigating_sale_application.feature#L1 
-rather than stopping on the income form, takes it all the way to sumit.


christine-sfg:
Will pick this up again when we rewrite the application forms

PR #1699, opened 2022-08-17
DAH-1188 add priority list to listing eligibility

kramduckner:


app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
```diff
@@ -111,6 +164,45 @@ export const ListingDetailsEligibility = ({
...
+                          "unit"
+                        )}`
+                      : `${unit.numberOfUnits} ${defaultIfNotTranslated("t.units", "units")}`
+                  }
+                >
+                  {defaultIfNotTranslated(
+                    "listings.unitsHaveAccessibilityFeaturesFor",
+                    `These units have accessibility features for people with ${unit.name}.`,
+                    {
+                      type: priorityLabelMap[unit?.name]?.description || unit.name,
```
christine-sfg, line 196: Looks like we're carrying over translation concatenation logic from angular which is not good practice. Can we flatten out the text props which will allow for proper translations? e.g.

```
"listings.unitsHaveAccessibilityFeaturesFor.impaired mobility":  "These units have accessibility features for people with impaired mobility.",
"listings.unitsHaveAccessibilityFeaturesFor.impaired vision and/or hearing": "These units have accessibility features for people with impaired vision and/or hearing."
```

I'm also inclined to leave the default description blank.

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
```diff
@@ -111,6 +164,45 @@ export const ListingDetailsEligibility = ({
...
+      {priorityUnits?.length > 0 ? (
+        <ListSection
+          title={t("listings.priorityUnits")}
+          subtitle={t("listings.priorityUnitsDescription")}
+        >
+          {priorityUnits
+            .filter((unit: ReducedUnit) => {
+              return unit?.name !== "Adaptable"
+            })
+            .map((unit: ReducedUnit) => {
```
christine-sfg, line 176: I noticed a slightly different ordering (the last two are swapped). Any idea what's going on here?
image

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
kramduckner: I was thinking we'd only leave it in until we get the new translations. I'm happy to add those the flattened versions in! I considered it, but was worried about the non english translations - is it ok if those are fully english until we get the proper translations? If we add those fields to the en.json file, then the non-english translations will default to those values. By leave the default description blank - meaning it'll just render an empty string if it can't find a translation?

christine-sfg: Yes, that is our current process. For any new text properties, they will display only in english until we process our next batch of translations. We should note it in the review instructions to avoid confusion (which you already did, thanks!) and then we should also make sure to add it to our 'next batch of translations' doc.

I hesitate to add any programmatic string concatenation, but i guess if it's just for the default/missing scenario and english only it can be ok in this case.

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
kramduckner: just an fyi - I updated the ticket's Review Instructions re: this. We looked into why this is the case and nothing was obvious and didn't want to spend too much time on something that might not be be desired. The ordering was built in a way that mirrored Salesforce.

jacksongonzales (approved):
🎉

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
```diff
@@ -9,19 +9,74 @@ import {
...
+    "Vision impairments": {
+      titleTranslation: "listings.prioritiesDescriptor.vision",
+      descriptionTranslation: "listings.unitsHaveAccessibilityFeaturesFor.impaired vision",
+    },
+    "Hearing/Vision (Communication)": {
+      titleTranslation: "listings.prioritiesDescriptor.hearingVisionCommunication",
+      descriptionTranslation:
+        "listings.unitsHaveAccessibilityFeaturesFor.impaired hearing and/or vision (communication)",
+    },
+  }
```
christine-sfg, line 78: is this data structure needed? could we change the text prop names to include the listing value, similar to what was done [here](https://github.com/SFDigitalServices/sf-dahlia-web/blob/69c740b8e1fd180e9fecd1dae6391940e085db6b/app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryPreferences.tsx#L35-L35)?

then you could access the values with `<prefix>.${priorityUnit.name}.title` and `<prefix>.${priorityUnit.name}.description`?

app/javascript/util/languageUtil.tsx
```diff
@@ -170,8 +174,20 @@ export function getReservedCommunityType(type: string | undefined): string {
 /**
  * If no translation exists for current key, return default salesforce value
  */
-export function defaultIfNotTranslated(key: string, value: string): string {
-  const translatedKey = t(key)
+export function defaultIfNotTranslated(
```
christine-sfg, line 177: would be cool to have unit test coverage for these changes

kramduckner:
@christine-sfg Added the unit test and replaced the label map!

christine-sfg (approved):

PR #1702, opened 2022-08-18
DAH-1209 Full image width on single listing view at tablet size

emilyjablonski:
Uptakes latest ui-components which includes a change that will push the image card on a single listing view at a tablet screen size to take up the full width of the screen.

Look at a single listing view and shorten the screen width, and when the right and left columns collapse into one, the image should take up the full width of the screen.


emilyjablonski:
Will connect with @kramduckner on the `read more` changes :)

kramduckner:
lgtm! One thing - I was thinking we'd have less margin in dahlia to match the angular. thoughts?
Screen Shot 2022-08-18 at 3 51 43 PM

emilyjablonski:
@kramduckner Hey Mark! I'm not sure, would poke product, and I think that might need to be a separate effort if yes :)

kramduckner (approved):

PR #1709, opened 2022-08-30
DAH-1229 bring back html filter to fix buggy parsing behavior

jacksongonzales:
This addresses a fix from Emily S's comments in [DAH-1229](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1229)

There appear to be a few listed issues with our parsing library and nested `<span>`s, particularly [this one](https://github.com/probablyup/markdown-to-jsx/issues/374) 😬 

Reinstantiating the filterUtil should keep us safe from this happening

### DIFF
--- a/app/javascript/util/filterUtil.tsx
+++ b/app/javascript/util/filterUtil.tsx
@@ -1,0 +1,17 @@
+// This is adapted from the filter implemented in customFilters.js.coffee
+// for doing a barebones sanitization of incoming raw html. Our html/markdown
+// parsing library (markdown-to-jsx) currently has reported issues
+// with nested <span> tags so it's necessary to strip them to
+// avoid buggy parsing behavior.
+
+// https://github.com/probablyup/markdown-to-jsx/issues?q=is%3Aissue+is%3Aopen+span
+
+export const stripMostTags = (input, allowed?: string) => {
+  if (!input) return ""
+  allowed = (((allowed || "<br><a>") + "").toLowerCase().match(/<[a-z][a-z0-9]*>/g) || []).join("")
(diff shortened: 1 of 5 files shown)


app/javascript/util/filterUtil.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const stripMostTags = (input, allowed?: string) => {
+  if (!input) return ""
+  allowed = (((allowed || "<br><a>") + "").toLowerCase().match(/<[a-z][a-z0-9]*>/g) || []).join("")
+  const tags = /<\/?([a-z][a-z0-9]*)\b[^>]*>/gi
+  const res = input.replace(tags, ($0, $1) => (allowed.includes(`<${$1.toLowerCase()}>`) ? $0 : ""))
+
+  // replace any newlines with break tags
+  return res.trim().replace(/\n/g, "<br>")
+}
```
christine-sfg, line 17: it was nice seeing all the formatting (markup) coming through. bummer that markdown-to-jsx doesn't handle nested elements well. seems like this is for the best tho. there may be other unexpected surprises allowing more than line breaks and links to come thru.

christine-sfg (approved):

app/javascript/util/filterUtil.tsx
jacksongonzales: ikr 😭 the example that emily found was messy enough to make me nervous though — in the future it might be worth looking into whether this is something we could make the salesforce WYSIWYG more opinionated about so we can accept more markup

PR #1723, opened 2022-09-29
DAH-1229 add option to allow tags in specified cases

jacksongonzales:
Yindi found [another issue](https://sfgovdt.jira.com/browse/DAH-1229?focusedCommentId=72314) that trickled into our new markdown approach at some point 😓 hopefully this addition adds some helpful control over how we sanitize.

In this case, we are being explicit about the content coming in, and want to display <span>s, so this shouldn't open any vulnerabilities.

### DIFF
--- a/app/javascript/modules/listings/DirectoryHelpers.tsx
+++ b/app/javascript/modules/listings/DirectoryHelpers.tsx
@@ -379,19 +379,19 @@ export const eligibilityHeader = (
       size: filters.household_size,
       people: filters.household_size === "1" ? t("listings.person") : t("listings.people"),
     })
-    const childrenContent = ` ${t("listings.includingChildren", {
+    const childrenContent = t("listings.includingChildren", {
       number: filters.children_under_6,
       children: filters.children_under_6 === "1" ? t("t.child") : t("t.children"),
-    })}`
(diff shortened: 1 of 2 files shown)


app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -379,19 +379,19 @@ export const eligibilityHeader = (
...
-    })}`
+    })
     return (
       <>
-        {renderInlineMarkup(householdSizeContent)}
-        {filters.include_children_under_6 && renderInlineMarkup(childrenContent)}
-        {renderInlineMarkup(incomeContent)}
+        {renderInlineMarkup(householdSizeContent, "<span>")}{" "}
+        {filters.include_children_under_6 && renderInlineMarkup(childrenContent, "<span>")}{" "}
+        {renderInlineMarkup(incomeContent, "<span>")}
```
christine-sfg, line 394: eek, you know how i feel about concatenating strings like this. these phrases may not be in the same order for all languages. ideally we would flatten this out.

that being said, i think you fixed the issue yindi found and afaik this is working for the 4 languages we currently support. will approve. i know this logic was carried over from the angular components

christine-sfg (approved):

PR #1724, opened 2022-09-29
Features/DAH-1244 machine translate

kramduckner:
[DAH-1244](https://sfgovdt.jira.com/browse/DAH-1244)


app/javascript/hooks/useLanguageStringInUrl.tsx
```diff
@@ -0,0 +1,7 @@
+import { getRoutePrefix } from "../util/languageUtil"
+
+const useLanguageStringInUrl = () => {
+  return window.location?.pathname && getRoutePrefix(window.location.pathname)
```
christine-sfg, line 4: can we use `getCurrentLanguage` in `languageUtil`?

app/javascript/pages/listings/listing-detail.tsx
```diff
@@ -45,89 +47,91 @@ const ListingDetail = () => {
...
-          <SiteAlert type="success" className={alertClasses} timeout={30_000} />
-        </div>
-        {listing && (
-          <article className="flex flex-wrap relative max-w-5xl m-auto w-full">
-            <ListingDetailsImageCard listing={listing} />
-            {!isApplicationOpen && (
-              <Mobile>
-                <ListingDetailsApplicationDate
-                  isApplicationOpen={isApplicationOpen}
+    <>
```
christine-sfg, line 50: is this change needed?

app/javascript/modules/listingDetails/ListingDetailsAdditionalInformation.tsx
```diff
@@ -38,7 +38,7 @@ export const ListingDetailsAdditionalInformation = ({
         {(!!listing.Required_Documents || isSale(listing)) && (
           <div className="info-card bg-gray-100 border-0">
             <h3 className="text-serif-lg">{t("listings.requiredDocuments")}</h3>
-            <div className="text-sm">
+            <div className="text-sm translate">
```
christine-sfg, line 41: was expecting more of these based on the compiled list (as noted in the AC). is that coming in a separate pr or separate ticket?

app/javascript/hooks/useTranslate.tsx
```diff
@@ -0,0 +1,60 @@
...
+    if (languageInRoute && googleTranslateDropdownElHasBeenRendered) {
+      const selectInDom = document.querySelector("select.goog-te-combo")
+      selectInDom.value = languageMap[languageInRoute] || languageInRoute
+      const ev = new Event("change", { bubbles: true })
+      selectInDom.dispatchEvent(ev)
+    }
+  }, [languageInRoute, googleTranslateDropdownElHasBeenRendered])
+}
+
+export default useTranslate
```
christine-sfg, line 60: [an article](https://kentcdodds.com/blog/how-to-test-custom-react-hooks) on testing custom hooks.  not sure if it's relevant here. if we're unable to add any test coverage in this pr, then let's make sure to file an e2e ticket

app/javascript/hooks/useTranslate.tsx
```diff
@@ -0,0 +1,60 @@
...
+      multilanguagePage: true,
+    },
+    "google_translate_element"
+  )
+}
+
+tWindow.initGoogleTranslate = initGoogleTranslate
+
+const useTranslate = () => {
+  const languageInRoute = useLanguageStringInUrl()
```
christine-sfg, line 39: wondering if this hook is necessary. what if we just call `getCurrentLanguage()` here. afaik changing languages currently results in a page load.

christine-sfg:
manual testing was great! awesome job!

app/javascript/modules/listingDetails/ListingDetailsAdditionalInformation.tsx
kramduckner: I should've made a comment in the ticket! I felt like this was likely to spill over into the next sprint and we're at a good place to make that a new ticket.

app/javascript/pages/listings/listing-detail.tsx
kramduckner: nope! good call. updated!

app/javascript/hooks/useLanguageStringInUrl.tsx
kramduckner: Didn't see that function. GetCurrentLanguage feels better suited. updated!

app/javascript/hooks/useTranslate.tsx
kramduckner: sounds good! I like that approach

app/javascript/modules/listingDetails/ListingDetailsAdditionalInformation.tsx
christine-sfg: yes, lets update the AC in the jira ticket so whoever works on testing is up to date. perhaps just cross out the original AC and link to the new ticket

app/javascript/hooks/useTranslate.tsx
christine-sfg: I guess i don't understand why checking for the language code needs to be a separate hook. How will that functionality be used again? What happens if we move that state here? Also wondering if we need to run any translation code if the page is english.

If a user clicks on a language link, the browser will navigate to a new page so seems that would reload everything (housing portal is not a SPA). For other non-page loading scenarios, I'm wondering if we may need additional functionality. For example, at some point we will support overriding of preference descriptions which are loaded from a separate request from the rest of the listing data. When those get loaded will the machine translations magically work or we'll somehow have to call google translate again? Sounds like a fun run race condition.

christine-sfg (approved):

PR #1727, opened 2022-10-03
DAH-1243 add units into waitlist desc

kramduckner:
[DAH-1243](https://sfgovdt.jira.com/browse/DAH-1243)

### DIFF
--- a/app/javascript/__tests__/modules/listingDetailsAside/__snapshots__/ListingDetailsWaitlist.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetailsAside/__snapshots__/ListingDetailsWaitlist.test.tsx.snap
@@ -19,7 +19,7 @@ exports[`ListingDetailsWaitlist renders waitlist section if listing has waitlist
         className="text-tiny text-gray-800 pb-3"
       >
         <p>
-          Once lottery winners fill all available units, additional lottery winners will be placed on the waitlist.
+          Once lottery winners fill all available units, additional lottery winners will be placed on the waitlist for 1 units
         </p>
       </p>
       <ul>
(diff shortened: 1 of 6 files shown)


app/javascript/__tests__/modules/listingDetailsAside/__snapshots__/ListingDetailsWaitlist.test.tsx.snap
```diff
@@ -19,7 +19,7 @@ exports[`ListingDetailsWaitlist renders waitlist section if listing has waitlist
         className="text-tiny text-gray-800 pb-3"
       >
         <p>
-          Once lottery winners fill all available units, additional lottery winners will be placed on the waitlist.
+          Once lottery winners fill all available units, additional lottery winners will be placed on the waitlist for 1 units
```
emilyjablonski, line 22: Is there a way we can prevent `1 units` vs `1 unit`?

app/javascript/modules/listingDetailsAside/ListingDetailsWaitlist.tsx
```diff
@@ -17,7 +17,13 @@ export const ListingDetailsWaitlist = ({ listing }: ListingDetailsWaitlistProps)
     </>
   )
 
-  const waitlistAvailableDescription = <p>{t("listings.availableUnitsAndWaitlistDescription")}</p>
+  const waitlistAvailableDescription = (
+    <p>
+      {t("listings.availableUnitsAndWaitlistDescription", {
+        number: listing.Total_number_of_building_units,
```
christine-sfg, line 23: Would be a #small-win if we wanted to support single vs plural - just check if the `listing.Total_number_of_building_units` is 1 or greater and use the appropriate text prop. That being said, we've spent plenty of time talking about the waitlist component. I'm ok with carrying forward the existing functionality.

christine-sfg (approved):

app/javascript/modules/listingDetailsAside/ListingDetailsWaitlist.tsx
kramduckner: sounds good! I didn't do it because all the different translations needed for that felt like it went beyond the scope

PR #1728, opened 2022-10-03
DAH-1223 Leasing Agent padding fixes

emilyjablonski:
This PR uptakes latest ui-components which includes the fixes in [DAH-1223](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1223).

For testing DAH-1223:
To test the padding issue where there is no email, see this listing `listings/a0W4U00000KnazlUAB` ([link](https://dahlia-webapp-pr-1728.herokuapp.com/listings/a0W4U00000KnazlUAB?react=true))
To test the empty phone issue, see this listing `listings/a0W4U00000KnScyUAF` ([link](https://dahlia-webapp-pr-1728.herokuapp.com/listings/a0W4U00000KnScyUAF?react=true))

In terms of other changes that have gone into Bloom that we are picking up:
A number of accessibility improvements which include updating spans with onClicks to buttons, updating the heading priority level of the list sections on the detail page so that they don't increase by 2 levels, and adding a tab index to some focusable elements. We also removed some class names from being sent as `undefined` if its prop was empty, and renamed our text styles which was a breaking change for the `Heading` component ([context](https://exygy.slack.com/archives/C040EKUAKTK/p1664839936759559)).

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsImageCard.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsImageCard.test.tsx.snap
@@ -18,12 +18,12 @@ exports[`ListingDetailsImageCard displays image card when no tag 1`] = `
         className="image-card__inner"
       >
         <img
-          alt="TEST Sale Listing (do not modify) - Homeownership Acres"
+          alt="A picture of the building"
           src="https://d31h23p2gsu1kr.cloudfront.net/images/listings/a0W0P00000GlKfBUAV-4d20089ed488fce1a695b73a8a4f090d.jpg"
         />
       </figure>
       <aside
-        aria-label="TEST Sale Listing (do not modify) - Homeownership Acres Statuses"
+        aria-label="A picture of the building Statuses"
         className="image-card__status"
       />
     </div>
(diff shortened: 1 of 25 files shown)


christine-sfg:
padding issue fix looks great
image


for empty phone issue, it looks like this listing listings/a0W4U00000KnScyUAF is supposed to have a leasing agent phone number but was removed

app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsImageCard.test.tsx.snap
```diff
@@ -18,10 +18,14 @@ exports[`ListingDetailsImageCard displays image card when no tag 1`] = `
         className="image-card__inner"
       >
         <img
-          alt="A picture of the building"
```
christine-sfg, line 21: would rather keep the old `alt` values. a user with a screen reader will already know the title of the listing. ideally we would have real descriptions for the listing images e.g. 'a three story building with mid century modern architecture and a curved walkway surrounded with yucca plants and shrubs'. however, we're not going to get that any time soon. another option is to leave off the alt rather than include redundant info which will be noisy for a11y users

package.json
```diff
@@ -8,7 +8,7 @@
     "@babel/plugin-transform-react-jsx": "^7.12.16",
     "@babel/preset-react": "^7.12.13",
     "@babel/preset-typescript": "^7.12.16",
-    "@bloom-housing/ui-components": "^5.1.1-alpha.8",
+    "@bloom-housing/ui-components": "^6.0.1-alpha.3",
```
christine-sfg, line 11: appreciate all the bloom fixes!!

emilyjablonski:
@christine-sfg For `for empty phone issue, it looks like this listing listings/a0W4U00000KnScyUAF is supposed to have a leasing agent phone number but was removed` - I removed the phone in Salesforce in order to be able to test the bug, so I think that is expected behavior unless I'm misunderstanding :)

emilyjablonski:
@christine-sfg Updated the alt text!

christine-sfg (approved):
Approved! Looks like we just need to fix a conflict.

PR #1734, opened 2022-10-15
DAH-1247 listing details e2e tests for open listings

christine-sfg:
DAH-1247

-upgrade cypress v8 -> v10
-found missing apply button for mobile
-first pass at e2e tests for the listing details page with smoke tests


kramduckner (approved):
lgtm!

PR #1735, opened 2022-10-25
DAH-1288 uptake button, apply style props

kramduckner:
[DAH-1288](https://sfgovdt.jira.com/browse/DAH-1288)

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsEligibility.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsEligibility.test.tsx.snap
@@ -270,19 +270,15 @@ exports[`ListingDetailsEligibility displays listing details eligibility section 
           className="expandable-text text-sm text-gray-700"
         >
            
-          <div>
-            <p>
-              Provide a credit report with score from Equifax, Experian, or TransUnion dated within thirty (30) days of the application. 
-            </p>
-            <p>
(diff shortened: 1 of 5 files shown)


app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsEligibility.test.tsx.snap
```diff
@@ -1334,19 +1314,15 @@ exports[`ListingDetailsEligibility displays listing details eligibility section
...
-              Landlord references will only check for...
-            </p>
-          </div>
+          <span>
+            Provide minimum of 4 years rental history with at least two prior rentals in which you were responsible for paying the rent.  Applicants without rental history will still be considered. No Guarantors permitted. 
+
+Previous rental history will be reviewed and must exhibit no derogatory references. 
+
+Landlord references will only check for...
+          </span>
```
christine-sfg, line 1323: previously we were supporting markup here, e.g. see Additional Eligibility Rules on this listing https://dahlia-full.herokuapp.com/listings/a0W0P00000GwGl3UAF?react=true 

we are now stripping out the markup and putting the content in a `span`. is that expected? can we still preserve the markup/formatting?

image

christine-sfg: I took a look at the updated review app. It looks like we're visually retaining the whitespace now. Thanks! 

However I did notice that we're using lots of `span` tags now rather than `p`. Do we have any concerns semantically?

image

emilyjablonski (approved):

christine-sfg (approved):
lgtm!

PR #1743, opened 2022-11-28
DAH-1303 add class for machine translated fields

jacksongonzales:
[DAH-1303](https://sfgovdt.jira.com/browse/DAH-1303)

A few notes:

- I noticed that our `TextTruncate` friend doesn't play nice with google translate. This seems to be related to issues with google translate discussed [here](https://github.com/facebook/react/issues/11538), which the workaround involves......putting `<span>`s around the translated content, which is exactly what we removed a while back after messy markup was making its way into the page.
    - For now, I removed the translate tag, since it totally breaks the page when you toggle the 'more/less' button. We should discuss a workaround, but, blegh
- For translating the office hours field in the lottery info aside, we'll need to make changes to the Bloom `Contact` component.
- For translating the additional fees field (under Features), we'll need to make changes to the Bloom `AdditionalFees` component.
- For addressing the issues Christine points out below (point 2), we'll need to make changes to the Bloom `Description` component.
---
If you think both/either of these require another ticket, I'm happy to write one up.


app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
```diff
@@ -23,7 +23,14 @@ interface FeatureItemProps {
...
 
-  return <Description term={title} description={stripMostTags(content)} markdown={true} />
+  return (
+    <Description
+      term={title}
+      description={stripMostTags(content)}
+      markdown={true}
+      dtClassName="translate"
+    />
+  )
```
christine-sfg, line 33: I think we may want to add a boolean to this function to flag whether a feature item should be machine translated. Some of fields may. We can probably leave out `Neighborhood` and `Year Built`.

app/javascript/modules/listingDetails/ListingDetailsAdditionalInformation.tsx
```diff
@@ -32,13 +32,15 @@ export const ListingDetailsAdditionalInformation = ({
         {listing.Listing_Other_Notes && (
           <div className="info-card bg-gray-100 border-0">
             <h3 className="text-serif-lg">{t("listings.specialNotes")}</h3>
-            <TextTruncate text={listing.Listing_Other_Notes} />
+            <div className="text-sm">
```
kramduckner, line 35: I think TextTruncate already wraps the jsx in a div with text-sm. The [snapshot](app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsAdditionalInformation.test.tsx.snap) shows a div with a text-sm class wrapping a div with a text-sm class. Not sure if that's intended, just something I noticed!

christine-sfg:
Eng review testing notes. I checked all 3 non english languages and compared to angular. they are translating as expected! see notes for each.

1. Eligibility > Credit History & Eviction History
the numbering is missing, but looks like an existing bug not part of this story. or perhaps we decided not to support this markup? 
image


2. Features: machine translate all field values except for Neighborhood, Year Built, Unit Features, and Additional Fees
couple questions (in the screenshot)
image

4. Lottery Results > Lottery Summary
tested with this listing a0W4U00000Ign3hUAB. lgtm
image

5. Lottery Info > Lottery Venue
I just saw in angular this one actually has a skip translate. sorry for the confusion here. that would explain why it wasn't translating before. let's remove machine translations here

jacksongonzales:
@christine-sfg weirdly I don't see the same numbered list you see on the same listing in Angular 🤔 

Screen Shot 2022-12-06 at 2 48 36 PM

jacksongonzales:
and for point #2, that involves changes to the bloom `Description` component. I can add that to the additional ticket for Bloom work

christine-sfg (approved):

PR #1759, opened 2023-01-04
DAH-1089 add unit accordion to features

jacksongonzales:
[DAH-1089](https://sfgovdt.jira.com/jira/software/c/projects/DAH/boards/122?modal=detail&selectedIssue=DAH-1089)

[DAH-1089]: https://sfgovdt.jira.com/browse/DAH-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DAH-1089]: https://sfgovdt.jira.com/browse/DAH-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/javascript/modules/listingDetails/ListingDetailsUnitAccordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+      content: (
+        <>
+          <span className="font-semibold">{unit.Unit_Square_Footage}</span>
+          {t("listings.features.sqft")}
+        </>
+      ),
+    },
+    baths: { content: <span className="font-semibold">{unit.Number_of_Bathrooms}</span> },
+    floor: { content: <span className="font-semibold">{unit.Unit_Floor}</span> },
+    accessibility: { content: <span className="font-semibold">TODO</span> },
```
jacksongonzales, line 38: I wasn't sure the set of possible values here; pinning this as a reminder to ask the team

app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
```diff
@@ -38,6 +46,42 @@ const FeatureItem = ({ content, title, toTranslate }: FeatureItemProps) => {
 }
 
 export const ListingDetailsFeatures = ({ listing, imageSrc }: ListingDetailsFeaturesProps) => {
+  const [units, setUnits] = useState({})
+
+  useEffect(() => {
+    void getUnits(listing.Id).then((units) => {
+      // eslint-disable-next-line unicorn/no-array-reduce
```
jacksongonzales, line 53: this no reduce rule seemed to come with an external linting ruleset. happy to try a different approach if folks find it difficult to read*

*imo I hate `.reduce`, but find that it's a symptom of data that hasn't been massaged properly for the front end

app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
```diff
@@ -109,6 +153,19 @@ export const ListingDetailsFeatures = ({ listing, imageSrc }: ListingDetailsFeat
 
           <Description term={t("listings.features.unitFeatures")} description={""} />
         </dl>
+        {Object.keys(units).length > 0 ? (
```
kramduckner, line 156: Sorry if this is easy to figure out and I just couldn't find an example of it while testing - is it possible that there'd be a valid API response with a length of 0 and the user would see an infinite spinner?

app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
kramduckner: ha I kind of like reduce. Even though I don't mind it, I feel similarly, if @christine-sfg isn't a fan I'd say find a different way!

app/javascript/modules/listingDetails/ListingDetailsUnitAccordion.tsx
kramduckner: I'd +1 that you covered them all! Looking at https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/assets/javascripts/listings/components/listing/features-section.html.slim#L54

kramduckner:
Looking great @jacksongonzales ! One question while testing - on listing a0W4U00000KnMyXUAV, is it ok that angular doesn't have TODO as the field for accessibility?

jacksongonzales:
> Looking great @jacksongonzales ! One question while testing - on listing a0W4U00000KnMyXUAV, is it ok that angular doesn't have TODO as the field for accessibility?

that was an earmark for myself, I'll have a fix for that soon!

app/javascript/modules/listingDetails/ListingDetailsFeatures.tsx
jacksongonzales: oops, I had the same question. I'd think no, but I'll ask the team to be sure on Monday

jacksongonzales: @kramduckner christine and I had a quick convo on [slack](https://sfdigitalservices.slack.com/archives/CCV0019TN/p1673290249038089) about this and I went with a quick solution that uses state to set a loading status. There's probably a more robust way to handle this, like lifting a loading state/spinner up into a higher-order component but that feels like scope creep. I'll add it as a nice-to-have ticket though!

kramduckner:
I see that the accessibility TODO string is fixed! Looks good. While verifying that I did notice that some of the translations aren't getting through (/listings/a0W4U00000KnMyXUAV?react=true)
Screenshot 2023-01-10 at 4 07 34 PM

jacksongonzales:
> I see that the accessibility TODO string is fixed! Looks good. While verifying that I did notice that some of the translations aren't getting through (/listings/a0W4U00000KnMyXUAV?react=true) Screenshot 2023-01-10 at 4 07 34 PM

@kramduckner that's my b! should be fixed now 💅

kramduckner (approved):
lgtm! with a fresh merge or rebase, the e2e tests should be working!

PR #1761, opened 2023-01-06
upgrade ffi for mac m1

kramduckner:
A recent ruby upgrade caused an issue with the ffi dependency on mac's with the m1 chip. This fixes it!


christine-sfg:
seems ok to me. @chadbrokaw-exygy could you please take a look?

christine-sfg (approved):

PR #1764, opened 2023-01-09
update id and text for machine translations

kramduckner:
DAH-1301

Since the Sale listing that was tested against isn't a protected test listing, we needed to update the sale listing the machine translations were using! This required updating the text that was being tested.


cypress/e2e/listingDetails/machineTranslations.e2e.ts
```diff
@@ -11,7 +11,7 @@ const listings = {
     title: "TEST Automated Listing (do not modify)",
   },
   HABITAT_SALE: {
-    id: "a0W4U00000KnMyXUAV",
+    id: "a0W0P00000GlKfB",
     address: "36 Amber Drive, San Francisco, CA 94131",
     title: "Habitat Amber Drive",
   },
```
christine-sfg, line 17: since this is no longer habitat listing, should we rename the var, address, title?

christine-sfg (approved):

PR #1771, opened 2023-01-24
DAH-1089 don't use abbr for square feet

jacksongonzales:
re: comments on [DAH-1089](https://sfgovdt.jira.com/browse/DAH-1089), opting to use the non-abbreviated "square feet" over "sqft." The latter will be read by screen readers as "ess cue eff tee" which isn't that helpful, and unfortunately there appears to be [no universally supported way to both abbreviate for sighted users and make sure a non-abbreviated format is read out for screen reader users](https://openinclusion.com/blog/presenting-abbreviations-acronyms-for-screen-reader-users/)*

*from a content perspective, avoiding abbreviations seems like good guidance for readability anyway, IMO :) 

[DAH-1089]: https://sfgovdt.jira.com/browse/DAH-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emilyjablonski:
Seems like we could alternatively use a solution like the below, if it's preferable to keep the abbrev for sighted users?
`<span aria-hidden="true">M</span>
<span class="sr_only">Monday</span>`

jacksongonzales:
> Seems like we could alternatively use a solution like the below, if it's preferable to keep the abbreviation for sighted users? Totally get it if we generally want to remove them from a UX perspective tho! `<span aria-hidden="true">M</span> <span class="sr-only">Monday</span>`
> 
> Tailwind has a helpful "screen reader only" utility: https://tailwindcss.com/docs/screen-readers

good idea! I'll defer to yindi on the UX question and we can go based on what they suggest

app/javascript/modules/listingDetails/ListingDetailsUnitAccordions.tsx
```diff
@@ -47,8 +47,8 @@ const getTableData = (units: RailsListingUnits[]) =>
     area: {
       content: (
         <>
-          <span className="font-semibold">{unit.Unit_Square_Footage}</span>
-          {t("listings.features.sqft")}
+          <span className="font-semibold">{unit.Unit_Square_Footage}</span>{" "}
+          {t("listings.features.squareFeet")}
```
christine-sfg, line 51: this reads out better, but i agree can be long. i'm not sure how the abbreviations would work in the other languages, but according to chrome 'sq ft' reads out as expected -> "two hundred fifty five square feet". i can show you how i'm testing.

app/javascript/modules/listingDetails/ListingDetailsUnitAccordions.tsx
```diff
@@ -47,8 +47,9 @@ const getTableData = (units: RailsListingUnits[]) =>
     area: {
       content: (
         <>
-          <span className="font-semibold">{unit.Unit_Square_Footage}</span>
-          {t("listings.features.sqft")}
+          <span className="font-semibold">{unit.Unit_Square_Footage}</span>{" "}
+          <span aria-hidden="true">{t("listings.features.sqft")}</span>
+          <span className="sr-only">{t("listings.features.squareFeet")}</span>
```
christine-sfg, line 52: reads out as expected on my mac w/ browser tools in chrome and safari. i also tested with [browser stack](https://www.browserstack.com/guide/test-websites-with-screen-readers) and this did not read out 'square feet', but it's possible i'm doing it wrong. i'm also trying to get access to the assistiv labs tooling. will let you know how that goes

christine-sfg: Looks like NVDA is ok, but Jaws is skipping over the square feet

https://user-images.githubusercontent.com/98352600/215902490-01898f15-ca09-4daa-9e10-c90f0a49ad84.mov

https://user-images.githubusercontent.com/98352600/215902509-95a8b2c8-dbed-4f4c-8150-7c7a520e0dd9.mov

PR #1792, opened 2023-03-15
Uptake Modal, ActionBlock breaking change in Bloom UIC

emilyjablonski:
The scrollable prop was renamed in the Modal component in Bloom, so this PR uptakes that breaking change ([UIC PR](https://github.com/bloom-housing/ui-components/pull/26)). The prop type for the heading in an ActionBlock was also changed, and that breaking change is updated here ([UIC PR](https://github.com/bloom-housing/ui-components/pull/44)).

To test, ensure the lottery results modal scrolls as expected / matches the behavior on full, and that the action block (as an example one is used on the home page) has no regression.


kramduckner (commented):
@emilyjablonski It's looking good! Is that build error expected?

emilyjablonski:
@kramduckner My b I didn't see that! I don't see any way to see more details, do you have access to the logs?

PR #1794, opened 2023-03-17
DAH-1436 Pricing table UI updates + React 18

emilyjablonski:
* Upgrades to React 18
* Removes enzyme since it is no longer compatible
* The RTL version of snapshot tests wraps them in a `<DocumentFragment>`, and renders them closer to actual DOM output so all snapshot tests will have at least the fragment wrapping change + most have lines that changed from `className` --> `class`
* Some snapshot tests have a much bigger diff bc they were never actually testing what they were supposed to be testing (some were capturing an empty or loading state because we weren't properly mocking the API)
* All async tests are now using done() which usually isn't necessary, but without them the tests bleed into each other when run as a full suite
* Updating to data-testid
* Adding a few new test ids
* Adding `type` for type imports - eslint was otherwise v upset


.eslintrc.js
```diff
@@ -45,14 +45,9 @@ module.exports = {
     "plugin:import/typescript",
     // Make sure we follow https://reactjs.org/docs/hooks-rules.html
     "plugin:react-hooks/recommended",
-    // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin
```
emilyjablonski, line 48: These were all merged into `prettier`

app/javascript/__tests__/pages/listings/__snapshots__/ForSale.test.tsx.snap
```diff
@@ -1,214 +1,461 @@
...
-        role="alert"
+        </nav>
+      </header>
+      <main
+        class="md:overflow-x-hidden"
+        data-testid="main-content-test-id"
+        id="main-content"
       >
         <div
-          className="loading-overlay"
```
emilyjablonski, line 177: Example issue in an old snapshot - we were just looking at the loading state

app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsUnitAccordions.test.tsx.snap
```diff
@@ -1,3 +1,111 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ListingDetailsUnitAccordion displays the unit accordions for a given listing 1`] = `null`;
```
emilyjablonski, line 3: Another buggy old snapshot

PR #1802, opened 2023-04-17
Non-Preference Lottery Modal

chadbrokaw-exygy:
[DAH-1451](https://sfgovdt.jira.com/browse/DAH-1451)

[DAH-1451]: https://sfgovdt.jira.com/browse/DAH-1451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kramduckner (approved):
This looks great. Do we need to add that translation to the google doc? 
For some reason, the View Lottery Results isn't opening the modal : (. I don't think it's a code change from here though. Mind double checking it's reproducible and not from this PR before merge?

cade-exygy (approved):
Testing Criteria passes. Everything looks good. I am curious why the one e2e test is failing? I tried to look a little into it, and the only difference in the language tests I could see is passing an empty string vs null for the address fields.

PR #1821, opened 2023-05-18
return list instead of string for tableSubHeader

cade-exygy:
JIRA Ticket: https://sfgovdt.jira.com/browse/DAH-1070

## Changes:

- Instead of generating a string with commas separating priorityTypes, we are now generating a list with each priorityType as a list item.
- **Changed the translation to remove the %priorities% field** - I think I edited all the language json files correctly, but I will put the updated text in the translation doc to ensure.
- Refactored the getTableSubheader function to be getPriorityTypes. I just returns a list of priority types now
- Created TableSubHeader component to render a list based on passed priority types. **Curious for feedback here**
- In progress: Tests are fixed, looking at some potential refactors still

## Review Instructions

- Go to /listings/for-rent?react=true
- Look for a listing with Priority Types
- It should be displayed as a bulleted list
- Check other languages and ensure everything looks consistent
- You may notice some priority types are not translated correctly. I am looking into a fix. That bug exists on full as well. https://dahlia-full.herokuapp.com/zh/listings/for-rent?react=true

### DIFF
--- a/app/assets/json/translations/react/en.json
+++ b/app/assets/json/translations/react/en.json
@@ -844,7 +844,7 @@
   "listings.householdMaximumIncome": "Household Maximum Income",
   "listings.housingProgram": "Housing Program",
   "listings.importantProgramRules": "Important Program Rules",
-  "listings.includesPriorityUnits": "Includes Priority Units for %{priorities}",
+  "listings.includesPriorityUnits": "Includes Priority Units for:",
   "listings.includingChildren": "(including <span class='eligibility-subtext'>%{number} %{children}</span> under 6)",
   "listings.incomeExceptions.intro": "People in your household may need <a href='%{url}' target='_blank'>special income calculations</a> if they:",
   "listings.incomeExceptions.nontaxable": "Receive non-taxable income (non-taxable income might include SSI, SSDI, child support payments, and worker’s compensation benefits).",
--- a/app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
+++ b/app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
@@ -1,0 +1,22 @@
+import React from "react"
+import renderer from "react-test-renderer"
+import TableSubHeader from "../../../modules/listings/TableSubHeader"
+
+describe("TableSubHeader", () => {
+  it("renders TableSubHeader component", () => {
+    window.matchMedia = jest.fn().mockImplementation((query) => {
+      return {
+        matches: true,
+        media: query,
+        onchange: null,
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+        addEventListener: jest.fn(),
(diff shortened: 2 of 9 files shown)


app/assets/json/translations/react/en.json
```diff
@@ -844,7 +844,7 @@
   "listings.householdMaximumIncome": "Household Maximum Income",
   "listings.housingProgram": "Housing Program",
   "listings.importantProgramRules": "Important Program Rules",
-  "listings.includesPriorityUnits": "Includes Priority Units for %{priorities}",
+  "listings.includesPriorityUnits": "Includes Priority Units for:",
```
chadbrokaw-exygy, line 847: Thank you for digging into these strings! I would honestly recommend deleting them from the es, tl, and zh files, just to make sure that we get the translation correct.

app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+        matches: true,
+        media: query,
+        onchange: null,
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+        dispatchEvent: jest.fn(),
+      }
+    })
```
chadbrokaw-exygy, line 18: I'm curious, why did you add this?

app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
chadbrokaw-exygy, line 1: Thank you for adding tests!! Would you also mind adding a test to check that the `getPriorityTypes` switch statement works? For example, "Mobility/hearing/vision impairments".

app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -147,7 +148,7 @@ export const getTableHeader = (listing: RailsRentalListing) => {
 }
 
 // Get the summary table subheader
-export const getTableSubHeader = (listing: RailsRentalListing) => {
+export const getPriorityTypes = (listing: RailsRentalListing): string[] | null => {
```
chadbrokaw-exygy, line 151: Nice!

chadbrokaw-exygy (approved):
Great job!

app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
cade-exygy: Good catch, it was a holdover from a different test file. We should be able to delete it!

app/javascript/__tests__/modules/listings/TableSubHeader.test.tsx
cade-exygy: Those tests should be in DirectoryHelpers.test.tsx

PR #1835, opened 2023-06-09
chore(deps): add nokogiri x86-darwin bits

shawnbot:


chadbrokaw-exygy:
@shawnbot Is this PR still active?

chadbrokaw-exygy (approved):
It looks like the builds are working... this change looks good to me

PR #1842, opened 2023-06-13
DAH-1536: Disable Flaky Test

alulabeshue-sfgov:
A couple of flaky tests asserting the translations of google translate are causing instability in our CircleCI pipeline. Because the same test is asserting multiple translations on different parts of the page, for now we will disable the assertion that is causing the flakiness. [DAH-1539](https://sfgovdt.jira.com/browse/DAH-1539) will investigate further the root cause of the flaky test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,12 +82,19 @@ To run Angular unit tests:
 To run React unit tests:
 - `yarn test`
 
-To run E2E tests:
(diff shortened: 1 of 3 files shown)


kramduckner (approved):
Looks great!

PR #1844, opened 2023-06-20
DAH-1507 Sales pricing range

cliu02:
[DAH-1507](https://sfgovdt.jira.com/browse/DAH-1507)

[DAH-1507]: https://sfgovdt.jira.com/browse/DAH-1507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsPricingTable.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsPricingTable.test.tsx.snap
@@ -56,7 +56,7 @@ exports[`ListingDetailsPricingTable does not render ListingDetailsPricingTable w
 
 exports[`ListingDetailsPricingTable renders ListingDetailsPricingTable component when rental listing 1`] = `
 <div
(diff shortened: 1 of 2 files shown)


app/javascript/api/types/rails/listings/RailsUnit.d.ts
```diff
@@ -21,10 +21,10 @@ type RailsUnit = {
   Unit_Square_Footage?: number
   attributes: ListingAttributes
   isReservedCommunity: boolean
-  Price_Without_Parking?: number
-  Price_With_Parking?: number
-  HOA_Dues_Without_Parking?: number
-  HOA_Dues_With_Parking?: number
+  Price_Without_Parking?: any
```
cade-exygy, line 24: What are the possible types for these fields? I am curious if we can be more specific than the any type. Could it be number | string | undefined | null ?

cade-exygy (commented):
Testing criteria passes, looks good!  I did notice that the 'with' and 'without parking' text on the react pricing table is not being translated from english.

cade-exygy (approved):

PR #1845, opened 2023-06-20
fix: add redis ssl configs

jimlin-sfgov:
Attempt to fix an SSL issue related to how Heroku utilizes Redis.

References:
- https://ogirginc.github.io/en/heroku-redis-ssl-error
- https://www.sandipmane.dev/openssl-connection-error-with-redis6-on-heroku

This may not be the correct way to configure things with our particular app set up, but due to the urgency of this issue, we should attempt to deploy this change on staging, and then production today.


cade-exygy (approved):

jimlin-sfgov:
Closing without merge, this branch was used as a smoke test in the staging environment.

PR #1848, opened 2023-06-20
Potential fix for banner translation

cade-exygy:
Pushing up this change for future reference.

Looks like the Listing Detail page is still being translated correctly, but worth double checking to ensure we are making redundant calls to the Google Translate service. @kramduckner


chadbrokaw-exygy:
@cade-exygy Can this PR be closed?

PR #1851, opened 2023-06-23
Merge production 06 23 23

kramduckner:


jimlin-sfgov:
The file I created for the Redis config hotfix isn't here: https://github.com/SFDigitalServices/sf-dahlia-web/blob/production/config/initializers/sidekiq.rb

Is that expected?

PR #1853, opened 2023-06-23
DAH-1484 Privacy Policy

cliu02:
[DAH-1484](https://sfgovdt.jira.com/browse/DAH-1484)

[DAH-1484]: https://sfgovdt.jira.com/browse/DAH-1484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/javascript/pages/getAssistance/privacy.tsx
```diff
@@ -1,12 +1,133 @@
...
+            </div>
+            <div className="md:pr-11 md:pl-0">
+              <hr />
+            </div>
+            <div className="space-y-4 p-6 md:py-11 md:pr-11 lg:pl-0">
+              <h2>{t("privacyPolicy.questionsTitle")}</h2>
+              <p>
+                {renderMarkup(
+                  `${t("privacyPolicy.questionsP1", {
+                    telLink: '<a href="tel:4157015500">415-701-5500</a>',
```
jimlin-sfgov, line 118: Just wondering if there's a reason we don't use translations for this phone number. Is it because nested `t()` calls are not supported?

app/javascript/pages/getAssistance/privacy.tsx
```diff
@@ -1,12 +1,133 @@
...
+                    emailLink:
+                      '<a href="mailto:sfhousinginfo@sfgov.org">sfhousinginfo@sfgov.org</a>',
```
jimlin-sfgov, line 120: Same question here about nested `t()` calls.

jimlin-sfgov (commented):
Posted some non-blocking questions

app/javascript/pages/getAssistance/privacy.tsx
jimlin-sfgov: Sorry, I meant that we have the phone number stored in this json:
https://github.com/SFDigitalServices/sf-dahlia-web/blob/be9da3a21a01b5a5ad3896d1a38a174ad904f0dc/app/assets/json/translations/react/en.json#L2

I think it would be better to use that, rather than hard-code the number here. But maybe it is not possible to do that?

jimlin-sfgov: Cool, looks like it is possible to nest `t()` calls. I do think that the name of the json key for this phone number could be better, but that is out of the scope of this ticket.

app/javascript/pages/getAssistance/privacy.tsx
jimlin-sfgov: Just to clarify, I do think it good practice to _not_ hardcode strings like email addresses. However, you could have to create a new entry into the translation file, something like this:
https://github.com/SFDigitalServices/sf-dahlia-web/blob/4dd1d9bc8f42c2b57c2f099afee0e213dd003112/app/assets/json/translations/react/en.json#L46

jimlin-sfgov (approved):
Looks good! I just had one non-blocking suggestion about avoiding hard-coded email addresses.

PR #1859, opened 2023-06-27
DAH-1485: Disclaimer page

cliu02:
[DAH-1485](https://sfgovdt.jira.com/browse/DAH-1485)

[DAH-1485]: https://sfgovdt.jira.com/browse/DAH-1485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/pages/getAssistance/disclaimer.tsx
+++ b/app/javascript/pages/getAssistance/disclaimer.tsx
@@ -2,8 +2,8 @@ import React from "react"
 import Layout from "../../layouts/Layout"
 import withAppSetup from "../../layouts/withAppSetup"
 import { ConfigContext } from "../../lib/ConfigContext"
-import { t, PageHeader } from "@bloom-housing/ui-components"
 import { MailingListSignup } from "../../components/MailingListSignup"
+import { t, PageHeader } from "@bloom-housing/ui-components"
 
 const Disclaimer = () => {
   const { getAssetPath } = React.useContext(ConfigContext)


jimlin-sfgov (changes requested):
I noticed that on the angular version of this page, the disclaimer intro text is hidden for screens smaller than the 768px breakpoint.

This is caused by the `show-for-medium-up` class here:https://github.com/SFDigitalServices/sf-dahlia-web/blob/2d41e1f0214f43ac8117263c50a1fe5d25ee6bcd/app/assets/javascripts/pages/templates/disclaimer.html.slim#L6

app/javascript/pages/getAssistance/disclaimer.tsx
```diff
@@ -1,12 +1,47 @@
...
+    <Layout title={t("pageTitle.disclaimer")}>
+      <PageHeader
+        title={t("pageTitle.disclaimer")}
+        subtitle={t("disclaimer.intro")}
+        inverse
+        backgroundImage={getAssetPath("bg@1200.jpg")}
+      />
+      {
+        <article className="flex flex-wrap relative max-w-5xl m-auto w-full">
+          <div className="w-full md:w-2/3" data-test-id="assistance-main-content">
```
jimlin-sfgov, line 20: Just wondering what `data-test-id` does? We're using the value `"assistance-main-content"` both here and in `AssistanceLayout.tsx`

jimlin-sfgov:
@cliu02 Sorry, I should have made my comment more clear. I meant that the React disclaimer page is not following same style behavior as the Angular disclaimer page, the behavior being that some text getting hidden when the screen gets small enough.

The Jira ticket says "Everything stays the same as Angular", so I assumed we want the hiding-text behavior from the Angular page copied over to the React page.

jimlin-sfgov (approved):
looks good!

chadbrokaw-exygy (approved):

PR #1860, opened 2023-06-28
DAH-1361: Upgrade to React 18

chadbrokaw-exygy:
- Upgrades to React 18
- Removes Webpacker and Webpacker-React in favor of Shakapacker and React-On-Rails, respectively
- Rebuilds build process
- Removes enzyme since it is no longer compatible
- The RTL version of snapshot tests wraps them in a <DocumentFragment>, and renders them closer to actual DOM output so all snapshot tests will have at least the fragment wrapping change + most have lines that changed from className --> class
- Some snapshot tests have a much bigger diff bc they were never actually testing what they were supposed to be testing (some were capturing an empty or loading state because we weren't properly mocking the API)
- All async tests are now using done() which usually isn't necessary, but without them the tests bleed into each other when run as a full suite
- Updating to data-testid
- Adding a few new test ids
- Adding type for type imports - eslint was otherwise v upset
- Upgrades dependencies


jimlin-sfgov:
The README instructions for spinning up the app locally seem to be outdated now, can you update that so we can manually test this PR locally?

There might be other things that need to be updated in the README, but I haven't gotten far enough in my review.

jimlin-sfgov:
Just making a note here that the branch name does not reflect the work done, and we need to consider that before merging to the main branch.

app/controllers/assistance_controller.rb
```diff
@@ -1,18 +1,22 @@
 # Controller for the get assistance pages
 class AssistanceController < ApplicationController
   def housing_counselors
+    @AssistanceProps = {assetPaths: static_asset_paths}
```
jimlin-sfgov, line 4: Just wondering why we're creating all of these instance variables. We get equivalent functionality by calling the `static_asset_paths` helper method inside the view:

```slim
// app/views/assistance/get_assistance.html.slim
== react_component 'GetAssistance', props: { assetPaths: static_asset_paths }
```

chadbrokaw-exygy: For sure, we could totally do it that way too! The reason I changed it was because the implementation of just writing `assetPaths: static_asset_paths` wasn't working. My logic for creating the instance variables was that it felt cleaner and easier to read for folks who are new to ruby/react-on-rails. 

Totally happy to change it to your way if you think that is better 😄

jimlin-sfgov: I usually prefer to have less code, but if you think it will help others then it may be worth keeping (there's always a balance between terseness and readability). One nit I do have is to use snake_case for variables coming from ruby-land, e.g. `@assistance_props`.

app/javascript/__tests__/data/RailsListingPricingTableUnits/listing-pricing-table-units-default.ts
```diff
@@ -0,0 +1,200 @@
+import type { RailsListingPricingTableUnit } from "../../../api/types/rails/listings/RailsListingPricingTableUnit"
```
jimlin-sfgov, line 1: The file that this import is using doesn't exist. It's strange that our linter is not catching this. I'm kind of worried now that there may be other things our linter isn't catching.

app/javascript/__tests__/data/RailsListingPricingTableUnits/listing-pricing-table-units-default.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export const pricingTableUnits: RailsListingPricingTableUnit[] = [
```
jimlin-sfgov, line 3: Is this data used anywhere? I feel like this file is a work-in-progress, and was forgotten about during the handoff.

app/javascript/authentication/SignInForm.tsx
```diff
@@ -20,6 +20,7 @@ import UserContext from "./context/UserContext"
 const SignInForm = () => {
   const { signIn } = useContext(UserContext)
   /* Form Handler */
+  // When you update to Version 7 of react-hook-form, "errors" becomes: "formState: { errors }""
```
jimlin-sfgov, line 23: Is this a note for update work that we have planned?

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
```diff
@@ -160,14 +159,14 @@ const buildAccordions = (
 
         const responsiveTableHeaders = listingIsSale
           ? {
-              units: { name: "t.unitType" },
```
jimlin-sfgov, line 163: This change doesn't seem related to the React 18 upgrade. I'm just now realizing that this PR combines two unrelated(?) Jira tickets, as indicated by the PR title.

jimlin-sfgov (commented):
I wasn't able to review all the changes in one go, here are the questions related to what I've reviewed so far.

app/javascript/components/base.scss
```diff
@@ -99,17 +98,17 @@ $colors: (
...
-      header-height: 83px,
-      row-height: 58px,
-    )
-  );
+  // @include ag-theme-alpine(
+  //   (
+  //     header-height: 83px,
+  //     row-height: 58px,
+  //   )
+  // );
```
jimlin-sfgov, line 111: Looks like we can remove this class entirely now?

app/javascript/hooks/useScript.tsx
```diff
@@ -13,7 +13,7 @@ const useScript = (src) => {
       }
       // Fetch existing script element by src
       // It may have been added by another intance of this hook
-      let script = document.querySelector(`script[src="${src}"]`)
+      let script: any = document.querySelector(`script[src="${src}"]`)
```
jimlin-sfgov, line 16: This is probably a quick fix to appease the linter, but we should try to avoid `any` types in general.

I realize that this file isn't great about typing things, and working to overhaul it is outside the scope of this ticket.

app/javascript/hooks/useTranslate.tsx
```diff
@@ -49,7 +50,8 @@ const useTranslate = () => {
 
   useEffect(() => {
     if (languageInRoute && googleTranslateDropdownElHasBeenRendered) {
-      const selectInDom = document.querySelector("select.goog-te-combo")
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const selectInDom: any = document.querySelector("select.goog-te-combo")
```
jimlin-sfgov, line 54: See previous comment about `any` types.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchForm.tsx
```diff
@@ -36,7 +36,7 @@ export const ListingDetailsLotterySearchForm = ({
   )
 
   const onSubmit = (data: { lotterySearchNumber: string }) => {
-    const { lotterySearchNumber } = data
+    const lotterySearchNumber = data?.lotterySearchNumber
```
jimlin-sfgov, line 39: If the `data` param is not guaranteed to have `lotterySearchNumber`, the function should be typed to indicate that.

app/javascript/pages/index.tsx
```diff
@@ -36,7 +36,7 @@ const HomePage = (_props: HomePageProps) => {
       />
       <div className="homepage-extra mt-2">
         <ActionBlock
-          header={t("welcome.newListingEmailAlert")}
+          header={<Heading priority={2}>{t("welcome.newListingEmailAlert")}</Heading>}
```
jimlin-sfgov, line 39: Just curious, is this change due to changes with Bloom UIC?

config/webpack/base.js
```diff
@@ -0,0 +1,58 @@
...
+    "process.env": {
+      COVID_UPDATE: JSON.stringify(process.env.COVID_UPDATE),
+      DIRECTORY_PAGE_REACT: JSON.stringify(process.env.DIRECTORY_PAGE_REACT),
+      GOOGLE_PLACES_KEY: JSON.stringify(process.env.GOOGLE_PLACES_KEY),
+      RESEARCH_FORM_URL: JSON.stringify(process.env.RESEARCH_FORM_URL),
+      SHOW_RESEARCH_BANNER: process.env.SHOW_RESEARCH_BANNER,
+      SRO_PLURAL_LISTINGS: process.env.SRO_PLURAL_LISTINGS,
+      TOP_MESSAGE: JSON.stringify(process.env.TOP_MESSAGE),
+      TOP_MESSAGE_TYPE: JSON.stringify(process.env.TOP_MESSAGE_TYPE),
+      TOP_MESSAGE_INVERTED: JSON.stringify(process.env.TOP_MESSAGE_INVERTED),
```
jimlin-sfgov, line 44: Why do some env vars need `JSON.stringify` while others do not?

config/webpack/base.js
```diff
@@ -0,0 +1,58 @@
...
+    },
+  })
+)
+
+
+const ignoreWarningsConfig = {
+  ignoreWarnings: [/Module not found: Error: Can't resolve 'react-dom\/client'/]
```
jimlin-sfgov, line 51: I'm sure there is a good reason for this, but it would helpful to add the reason in a comment

jimlin-sfgov (commented):
This is part 2 of my review. As mentioned earlier, I had to break it up because I couldn't get through the review in one day.

config/initializers/assets.rb
```diff
@@ -10,4 +10,8 @@
 # application.js, application.css, and all non-JS/CSS in app/assets
 # folder are already added.
 # Rails.application.config.assets.precompile += %w( search.js )
+# We may need to add   jsonfile/index.js
```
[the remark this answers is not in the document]
chadbrokaw-exygy: Will remove!

config/shakapacker.yml
```diff
@@ -1,11 +1,11 @@
-# Note: You must restart bin/webpack-dev-server for changes to take effect
+# Note: You must restart bin/webpacker-dev-server for changes to take effect
```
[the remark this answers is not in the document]
chadbrokaw-exygy: Good Catch!

app/javascript/pages/index.tsx
chadbrokaw-exygy: Yes

app/javascript/__tests__/data/RailsListingPricingTableUnits/listing-pricing-table-units-default.ts
chadbrokaw-exygy: Good catch, and agreed on the linter. I will look into the linter to make sure that it is correctly flagging these types of things.

I wonder if this file was brought in during Mark's work on the pricing table and was forgotten... It doesn't look like it is being used anywhere in the code base (which might be why the linter didn't catch it?). I am going to delete the file unless there is any objection from @emilyjablonski

app/javascript/__tests__/data/RailsListingPricingTableUnits/listing-pricing-table-units-default.ts
chadbrokaw-exygy: It doesn't look like it is being used anywhere, should be deleted IMO

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
chadbrokaw-exygy: @kramduckner Can you comment on this? I think this is a botched merge and should be reverted, but I wanted to confirm

app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchForm.tsx
chadbrokaw-exygy: Agreed!

config/webpack/base.js
chadbrokaw-exygy: I've gone back and forth on this, most [resources](https://github.com/shakacode/react-webpack-rails-tutorial/pull/519) online used it, but the error is not present in our code. 

I have commented out this line and left a comment above it outlining that it could be needed in the future.

app/javascript/hooks/useScript.tsx
chadbrokaw-exygy: Ugh, I completely agree with you. I'm not a fan of any types. This was a quick fix to make the linter happy since we were effectively treating `script` (and below `selectInDom`) as any type anyway. 

I will work on this today to see if I can get the typing right, if I can't then I will create a ticket to tackle it later.

config/webpack/base.js
chadbrokaw-exygy: I'm not completely sure. I couldn't figure out a justification for that, but this was in the previous config so I decided to leave it as-is.

app/javascript/__tests__/data/RailsListingPricingTableUnits/listing-pricing-table-units-default.ts
emilyjablonski: No objection!

jimlin-sfgov (approved):
My feedback has been addressed, looks good!

I'm approving with the expectation that we will get a few more approvals before merging.

app/javascript/hooks/useScript.tsx
chadbrokaw-exygy: Created a ticket to track this effort: https://sfgovdt.jira.com/browse/DAH-1581

app/javascript/hooks/useTranslate.tsx
chadbrokaw-exygy: Created a ticket to track this effort: https://sfgovdt.jira.com/browse/DAH-1581

PR #1871, opened 2023-07-11
DAH-1539 reduce timeout and add retries

cade-exygy:
Pushing up a branch without the Cypress upgrade. 
- Reduces test timeout and add retries.
- Intercept and wait for the google translate request before checking the DOM

run `yarn test:e2e` and ensure all the Cypress tests pass

you can also run `yarn run cypress open` to open the Cypress desktop app and see more details as the tests run


chadbrokaw-exygy (approved):
Nice job!

PR #1878, opened 2023-07-21
DAH-1436.2: Pricing Table UI Fixes

chadbrokaw-exygy:
[DAH-1436](https://sfgovdt.jira.com/browse/DAH-1436)

This is to make some more fixes to the pricing table and is a follow up to the previous PR.

[DAH-1436]: https://sfgovdt.jira.com/browse/DAH-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsApplicationDate.test.tsx.snap
+++ b/app/javascript/__tests__/modules/listingDetails/__snapshots__/ListingDetailsApplicationDate.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`ListingDetailsApplicationDate displays Application Deadline when listing is a sale and due date has not passed 1`] = `
 <DocumentFragment>
   <div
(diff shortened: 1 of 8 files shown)


jimlin-sfgov:
I read this in the Jira ticket:
> This must be done before the listing detail page is live to the public.

Does this mean that the changes in this PR need to be in the July release?

chadbrokaw-exygy:
> I read this in the Jira ticket:
> 
> > This must be done before the listing detail page is live to the public.
> 
> Does this mean that the changes in this PR need to be in the July release? Or is it okay to split this PR across two releases, because the React listing detail page is still behind a feature flag?

I think this will be okay to split between two releases. This just needs to done before we launch the listing detail page to the public

PR #1890, opened 2023-08-10
DAH-1446: Multiple Images on Listing Detail Page

chadbrokaw-exygy:
[DAH-1446](https://sfgovdt.jira.com/browse/DAH-1446)

Deployment to production will be timed with a migration on Salesforce.

[DAH-1446]: https://sfgovdt.jira.com/browse/DAH-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/__tests__/pages/listings/ForSale.test.tsx
+++ b/app/javascript/__tests__/pages/listings/ForSale.test.tsx
@@ -30,9 +30,7 @@ describe("For Sale", () => {
     const { findByAltText } = render(<ForSale assetPaths="/" />)
 
     const image = await findByAltText(`${openSaleListing.Building_Name} Building`)
-    expect(image.getAttribute("src")).toBe(
-      "https://d31h23p2gsu1kr.cloudfront.net/images/listings/a0W0P00000GlKfBUAV-4d20089ed488fce1a695b73a8a4f090d.jpg"
-    )
+    expect(image.getAttribute("src")).toBe(openSaleListing.Listing_Images[0].Image_URL)
     done()
   })
 })
--- a/app/javascript/__tests__/pages/listings/ForSale.test.tsx
+++ b/app/javascript/__tests__/pages/listings/ForSale.test.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { render, cleanup } from "@testing-library/react"
 import ForSale from "../../../pages/listings/for-sale"
+import { openSaleListing } from "../../data/RailsSaleListing/listing-sale-open"
 
 const axios = require("axios")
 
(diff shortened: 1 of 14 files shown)


app/javascript/__tests__/pages/listings/ForSale.test.tsx
```diff
@@ -22,4 +23,16 @@ describe("For Sale", () => {
...
   })
+
+  it("listings with multiple listings render the first image in the array", async (done) => {
+    axios.get.mockResolvedValue({ data: { listings: [openSaleListing] } })
+
+    const { findByAltText } = render(<ForSale assetPaths="/" />)
+
+    const image = await findByAltText(`${openSaleListing.Building_Name} Building`)
+    expect(image.getAttribute("src")).toBe(
+      "https://d31h23p2gsu1kr.cloudfront.net/images/listings/a0W0P00000GlKfBUAV-4d20089ed488fce1a695b73a8a4f090d.jpg"
```
cade-exygy, line 34: I would need to test this further, but on first glance: this test may fail if image caching is toggled off. I think we would use the sfmohcd domain instead of cloudfront

app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx
```diff
@@ -19,18 +20,25 @@ export const ListingDetailsImageCard = ({ listing }: ListingDetailsImageCardProp
...
+    }
+  })
 
   return (
     <header className="image-card--leader">
       <ImageCard
-        imageUrl={listing?.imageURL ?? fallbackImg}
-        href={`/listings/${listing.listingID}`}
+        imageUrl={(listingImages === undefined && listing?.imageURL) ?? fallbackImg} // Fallback if there is no listing images array.
+        images={listingImages?.length === 0 ? [{ url: fallbackImg }] : listingImages}
```
cade-exygy, line 34: Does the `fallbackImg` have a description that we would want to include?

cade-exygy (approved):
Testing criteria passes. Looks amazing!
I will keep testing some edge cases but this is really great.

app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx
chadbrokaw-exygy: Talked with Bridget and T and decided that we should be hiding the fallback Image (and any descendants of the component) from a11y tools since the fallback is decorative and provides no functional value

app/javascript/__tests__/pages/listings/ForSale.test.tsx
chadbrokaw-exygy: Great call out! I think changing out the image urls in stubbed data probably warrants its own ticket. I did, however, change the test so that it just pulls the URL right from the stubbed data, that way we only have to change it in one place.

What do you think?

cade-exygy: I think that is a good solution for now!

app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx
cade-exygy: That makes sense. The only other consideration I can think of is having some helpful alt text describing that the fallback image is just a placeholder. May be helpful for low vision folks to make clear that the fallback image is not a photo representing the listing.

chadbrokaw-exygy: Made a note to ask about this during our A11Y audit: https://sfgovdt.jira.com/browse/DAH-1641

PR #1894, opened 2023-08-15
Dah 1572 listing image cache flag

cade-exygy:
Branch for misc bug fixes


jimlin-sfgov:
@cade-exygy Which ticket does this PR belong to? I see that DAH-1572 is already complete, while DAH-1411 is still in progress.

app/services/force/request.rb
```diff
@@ -20,11 +20,18 @@ def get(endpoint, params = nil)
...
+      variable.to_s.casecmp('true').zero?
+    end
+
     def cached_get(endpoint, params = nil, force = false)
       key = "#{endpoint}#{params ? '?' + params.to_query : ''}"
       force = ActiveModel::Type::Boolean.new.cast(force)
-      force_refresh = force || !ENV['CACHE_SALESFORCE_REQUESTS']
-      if ENV['FREEZE_SALESFORCE_CACHE']
+      if force == true
+        puts 'Force is true'
```
jimlin-sfgov, line 31: Was this line intentionally kept in the PR?

alulabeshue-sfgov (commented):
Setting the `CACHE_SALESFORCE_REQUESTS` in full forces the listing details page to display the new image from salesforce as expected. 

The listing directory image on the other hand was not updating when setting this env var or providing `preview=true`. Looking at the network response for the directory page it looks like rails returns both a `Building_URL`, which has the correct link to the updated image, and `Listing_Images` which has both the old and new link and it seems the old link is being used. Is this something we are aware of and investigating or do we need to look into this more?

Screenshot 2023-08-23 at 1 50 47 PM

app/services/force/request.rb
cade-exygy: Yeah I think more log messages would be helpful. I can add some more robust logging.

jimlin-sfgov:
> I can move these commits to a new branch

I don't think that's necessary. I just want to confirm which ticket this PR belongs to, it makes the release process less confusing.

alulabeshue-sfgov (approved):
After meeting with Cade and testing we found that the reason in my earlier comment that the directory page was not updating was my misunderstanding of multi images vs the old single image. After getting access to the multi image field in salesforce UI things look better. 

The only case I found that is failing is when caching is enabled the directory page for react is not updating even when preview is set to true. I think this is part of what we are investigating for the caching issues and I dont see a need to hold this PR for that.

jimlin-sfgov (approved):

jimlin-sfgov:
@cade-exygy I'm changing this PR to refer to DAH-1411, based on what you have told me

jimlin-sfgov:
Code changes are not related to the failing flaky tests, safe to merge

PR #1918, opened 2023-08-31
DAH-1645 Condtionally Run Angular e2e tests

jimlin-sfgov:


jimlin-sfgov:
New github workflows will not work until their config files are on the `main` branch. I will merge this now, to test; I will revert if needed.

PR #1921, opened 2023-08-31
DAH-1645 Fix circleci filter for e2e-legacy

jimlin-sfgov:


jimlin-sfgov:
Same as other DAH-1645 PRs, need to merge to test.

PR #1923, opened 2023-08-31
DAH-1645 Use CircleCI Branch Filter

jimlin-sfgov:
CircleCI does not have access to Github labels, so the strategy of auto-labeling PRs with Github workflows to manage CircleCI jobs is not feasible.

This PR reverts the auto-labeling work, and also switches the CircleCI filtering to use branch names.


jimlin-sfgov:
Need to merge to test.

PR #1925, opened 2023-08-31
DAH-1634: Fix Minimum Values in Pricing Table

chadbrokaw-exygy:
https://sfgovdt.jira.com/browse/DAH-1634

The document linked in the JIRA review instructions has information about the expected behavior. Note that this does not include pricing as a percentage of income, which will come in the follow up PR: #1927.

This PR also does not completely address testing - though the tests should all pass - that will be covered in this [ticket](https://sfgovdt.jira.com/browse/DAH-1686).

NOTE: This PR also takes in the newest Bloom UIC which addresses this ticket: https://sfgovdt.jira.com/browse/DAH-1612

### DIFF
--- a/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
+++ b/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
@@ -323,7 +323,7 @@ const buildContent = (
   let groupedUnitsByOccupancy: GroupedUnitsByOccupancy[] = []
 
   if (units?.length) {
-    groupedUnitsByOccupancy = groupAndSortUnitsByOccupancy(units, amiCharts)
+    groupedUnitsByOccupancy = groupAndSortUnitsByOccupancy(units, amiCharts, listingIsSale)
   }
 
   if (listingIsHabitat) {
--- a/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
+++ b/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
@@ -2,7 +2,7 @@ import React, { useContext, useEffect } from "react"
 import { CategoryTable, ContentAccordion, Icon, t } from "@bloom-housing/ui-components"
 import { RailsListing } from "../listings/SharedHelpers"
 import { isHabitatListing, isSale, groupAndSortUnitsByOccupancy } from "../../util/listingUtil"
-import type { RailsUnitWithOccupancyAndMaxIncome } from "../../api/types/rails/listings/RailsUnit"
+import type { RailsUnitWithOccupancyAndMinMaxIncome } from "../../api/types/rails/listings/RailsUnit"
 import type RailsUnit from "../../api/types/rails/listings/RailsUnit"
 import type { RailsAmiChart } from "../../api/types/rails/listings/RailsAmiChart"
 import ListingDetailsContext from "../../contexts/listingDetails/listingDetailsContext"
@@ -16,7 +16,7 @@ export interface ListingDetailsPricingTableProps {
 
 export interface AmiRow {
   ami: { min: number | undefined; max: number }
-  units: RailsUnitWithOccupancyAndMaxIncome[]
+  units: RailsUnitWithOccupancyAndMinMaxIncome[]
 }
 
 export interface GroupedUnitsByOccupancy {
@@ -26,7 +26,7 @@ export interface GroupedUnitsByOccupancy {
   amiRows: AmiRow[]
 }
 
-const buildSalePriceCellRow = (unit: RailsUnitWithOccupancyAndMaxIncome) => {
+const buildSalePriceCellRow = (unit: RailsUnitWithOccupancyAndMinMaxIncome) => {
   if (unit.Price_With_Parking && unit.Price_Without_Parking) {
     return [
       {
@@ -59,7 +59,7 @@ const buildSalePriceCellRow = (unit: RailsUnitWithOccupancyAndMaxIncome) => {
   }
 }
 
-const buildSaleHoaDuesCellRow = (unit: RailsUnitWithOccupancyAndMaxIncome) => {
+const buildSaleHoaDuesCellRow = (unit: RailsUnitWithOccupancyAndMinMaxIncome) => {
   if (unit?.HOA_Dues_With_Parking && unit?.HOA_Dues_Without_Parking) {
     return [
       {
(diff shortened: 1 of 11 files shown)


jimlin-sfgov:
This might just be me, but I think I'm spending more time figuring out _how_ to test this PR, than actually testing it. 

I read the "Acceptance Criteria" section of the Jira ticket, but it doesn't go into much detail. E.g., "some wording and pricing changes", but doesn't list out what the changes are; "some small monetary deviations", but doesn't explain what the math behind the deviations.

It then says to check out a table in a separately linked doc. Is it correct to say that I should be reading through each rule in that table, and then reviewing the pricing tables in the PR app for each listing example in the table?

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
```diff
@@ -206,7 +200,7 @@ const buildAccordions = (
               </span>
               <span className={"flex items-center mr-2 text-sm md:text-base"}>
                 {(() => {
-                  return occupancy?.absoluteMinIncome?.valueOf() === 0 ? (
+                  return occupancy?.absoluteMinIncome?.valueOf() <= 0 ? (
```
jimlin-sfgov, line 203: Do we actually need to use `valueOf`? I know that this is existing code, but since this line is getting updated, I'm using this opportunity to discuss removing it.

app/javascript/util/listingUtil.ts
```diff
@@ -192,17 +194,36 @@ export const deriveIncomeFromAmiCharts = (
...
+ * For Sale listings, we try to derive the minimum AMI, if we can't then we return -1.
+ */
+const determineMinIncomeNeeded = (
+  unit: RailsUnitWithOccupancy,
+  amiCharts: RailsAmiChart[],
+  isSale: boolean
+) => {
+  return isSale
+    ? deriveIncomeFromAmiCharts(unit, unit.occupancy, amiCharts, true) || -1
+    : unit?.BMR_Rental_Minimum_Monthly_Income_Needed || -1
```
jimlin-sfgov, line 218: Just wondering if there was a particular reason to choose `-1`, as opposed to `0` or just letting falsy values go through.

app/javascript/util/listingUtil.ts
```diff
@@ -371,16 +393,23 @@ export const groupAndSortUnitsByOccupancy = (
...
   /*
-   * We have to derive the max income using ami charts, so this mapping goes through each unit
-   * and does that and adds that max income to the unit object
+   * We have to derive the max and min income using ami charts, so this mapping goes through each unit
+   * and does that and adds that max and min income to the unit object.
+   * The Min income is derived through two means: by the AMI Chart for that value or via the BMR_Rental_Minimum_Monthly_Income_Needed field.
+   * The former is only applicable when the listing is a Section 415 housing listing (meaning there are discrete AMI ranges), while the latter
+   * is calculated in Salesforce by multiplying the base rent by the rent multiple (usually 2x).
+   * To determine this, we check if there is a BMR value and use that if there is, if not, we try to use the minimum AMI field.
+   * If neither of those are available, the value becomes -1, which forces the table to display "Up To".
```
jimlin-sfgov, line 402: Long multi-line blocks of comments are a bit of a code smell. The comments themselves are not the core issue, but they indicate that the logic for creating the pricing table is overly complex and/or hard to read, and may become difficult to update or debug in the future.

With the current priorities of our engineers, we don't have the luxury of trying to refactor this code, but we should look into this later when we have time.

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
chadbrokaw-exygy: Honestly, I doubt that we need it. If you want to remove it that is good with me

app/javascript/util/listingUtil.ts
chadbrokaw-exygy: For sure! I think there are a lot of correct answers here, but the reason I didn't pick a falsy value was because we use this value to calculate the absolute min/max in the header of the accordion and I didn't want to throw a falsy value into the calculation (though I _think_ it would still work, it just felt more confusing that necessary). 0 could work here, but it just seemed logically confusing since a value _could_ be zero so it would be difficult to tell if the value was zero of we just didn't have a value to start with.

Happy to reconsider this decision if you feel strongly either way

app/javascript/util/listingUtil.ts
chadbrokaw-exygy: For sure

chadbrokaw-exygy:
> This might just be me, but I think I'm spending more time figuring out _how_ to test this PR, than actually testing it.
> 
> I read the "Acceptance Criteria" section of the Jira ticket, but it doesn't go into much detail. E.g., "some wording and pricing changes", but doesn't list out what the changes are; "some small monetary deviations", but doesn't explain what the math behind the deviations.
> 
> It then says to check out a table in a separately linked doc. Is it correct to say that I should be reading through each rule in that table, and then reviewing the pricing tables in the PR app for each listing example in the table?

Definitely not just you! The pricing table and all its deviation can definitely be confusing. I agree that testing this leaves a lot to be desired which is why I created a more robust step-by-step testing document, linked in the JIRA.

The hard part about testing this is that there are a lot of little changes, I tried to identify those as best I could, but it is possible some slipped through the cracks.

app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -73,7 +73,7 @@ const isNumber = (val: number) => val || val === 0
 // 100, 100, false, "%" --> 100%
 export const getRangeString = (min: number, max: number, currency?: boolean, suffix?: string) => {
   // If min is 0, return "up to {max}"
```
jimlin-sfgov, line 75: since the comparison logic has changed, we should update this comment to mention the `-1` scenario

app/javascript/util/listingUtil.ts
jimlin-sfgov: No preference from me, I was just curious

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
jimlin-sfgov: I think removing `valueOf` is a good idea, it's not normally used this way and could be confusing for other developers.

app/javascript/modules/listings/DirectoryHelpers.tsx
chadbrokaw-exygy: Good catch, thank you

jimlin-sfgov (approved):
great job on this big update!

PR #1926, opened 2023-08-31
DAH-1667: Add banner to listing image with custom content

alulabeshue-sfgov:
[DAH-1667](https://sfgovdt.jira.com/browse/DAH-1667)

1. in salesforce select `Educator 1: SFUSD employees only` as the custom listing type for a listing

Screenshot 2023-08-31 at 3 23 52 PM

2. banner with the content matching DAH-1667 should appear on the details page

Screenshot 2023-08-31 at 3 24 10 PM

3. clicking the link on the banner should scroll down to the eligibility section

Screenshot 2023-08-31 at 3 24 26 PM

[DAH-1667]: https://sfgovdt.jira.com/browse/DAH-1667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/javascript/modules/listingDetails/ListingDetailsReservedBanner.tsx
```diff
@@ -1,33 +1,60 @@
...
+    return (
+      <Message warning={true}>
+        {renderInlineMarkup(`${t("listings.customListingType.educator.banner")}`, "<p><b>")}
+        {renderInlineMarkup(
+          t(`listings.customListingType.educator.banner.link.content`, {
+            url: `#listing-details-eligibility`,
+          }),
+          "<p><a>"
+        )}
+      </Message>
```
jimlin-sfgov, line 27: Correct me if I'm wrong, but It looks like a lot of work has been done to work around the limitations of the Bloom `Message` component, which will normally display text in all-caps with no line breaks. E.g. Adding HTML tags into translations and using the override feature of the `markdown-to-jsx` package.

I think the work arounds are fine, in the interest of time, but we should add some comments to indicate that these code changes are workarounds, and also make a feature request to Bloom developers, to make their `Message` component more flexible.

alulabeshue-sfgov: Yes partly. Today the `Message` component does take in React components and respects things like break lines, anchor tags, and bold text. But it does by default make everything capitalized. I reverted updating the `Markdown` component and added the styling to the html itself to force `Message` to respect the non capitalized letters. I think there also may be a way to override some of the classNames of `Message` that force it to be all capital, I'll take a look and if not we can open a request with Bloom.

jimlin-sfgov: Hmm, updating the `Markdown` component and adding inline styles to translation files both seem equally unideal to me, but are fine as short term workarounds. I'll leave it up to you to decide which way is better.

jimlin-sfgov:
Can you check with a designer if they want margins around the banner on mobile? They seem to be picky about that kind of stuff.
Screenshot 2023-08-31 at 6 28 49 PM

jimlin-sfgov (approved):
Looks good, I'm liking the new approach to overriding `Message` styling. The lack of margins on mobile is unchanged, so I'm assuming designers are okay with that.

PR #1927, opened 2023-09-01
DAH-1659: Rent a percent of income

chadbrokaw-exygy:
https://sfgovdt.jira.com/browse/DAH-1659

### DIFF
--- a/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
+++ b/app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
@@ -146,19 +146,16 @@ const buildHeader = (amiRow: AmiRow, showFullText: boolean): string => {
 
 const buildAccordions = (
   groupedUnitsByOccupancy: GroupedUnitsByOccupancy[],
-  listingIsSale: boolean
+  listingIsSale: boolean,
+  forceZeroInRange: boolean
 ) => {
   return groupedUnitsByOccupancy?.map(
     (occupancy: GroupedUnitsByOccupancy, index: number, array) => {
       const accordionLength = array.length
-      let forceZeroInRange = false
 
       const categoryData = occupancy?.amiRows?.map((amiRow: AmiRow, amiRowIndex: number) => {
         const responsiveTableRows = amiRow.units.map(
           (unit: RailsUnitWithOccupancyAndMinMaxIncome) => {
-            if (unit.Rent_percent_of_income) {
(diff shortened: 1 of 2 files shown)


app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
```diff
@@ -213,7 +223,9 @@ const buildAccordions = (
                     <div>
                       {renderInlineMarkup(
                         t("listings.incomeRange.minMaxPerMonth", {
-                          min: Math.round(occupancy?.absoluteMinIncome).toLocaleString(),
+                          min: Math.round(
+                            forceZeroInRange ? 0 : occupancy?.absoluteMinIncome
+                          ).toLocaleString(),
```
jimlin-sfgov, line 228: With the nested conditional, it's getting hard to see what the intended behavior is. Just want to double check that this is the the behavior for each scenario:
1. `occupancy?.absoluteMinIncome <= 0 && forceZeroInRange`
    - "$0 to $X per month"
2. `occupancy?.absoluteMinIncome <= 0 && !forceZeroInRange`
    - "up to $X per month"
3. `occupancy?.absoluteMinIncome > 0 && forceZeroInRange`
    - "$0 to $X per month"
4. `occupancy?.absoluteMinIncome > 0 && !forceZeroInRange`
    - "$Y to $X per month"

There's probably some way to express all the scenarios explicitly, while also keeping the code compact, but I haven't figured one out yet.

app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -71,15 +71,21 @@ const isNumber = (val: number) => val || val === 0
 // Transforms two numbers into a range string with optional currency formatting and optional suffix
 // 100, 200, true, null --> $100 - $200
 // 100, 100, false, "%" --> 100%
```
jimlin-sfgov, line 73: With the new `forceZeroInRange` argument introduced, existing comments should be updated.

jimlin-sfgov (changes requested):

app/javascript/modules/listingDetails/ListingDetailsPricingTable.tsx
chadbrokaw-exygy: Yes, that is correct

jimlin-sfgov (approved):
looks good!

PR #1936, opened 2023-09-14
Dah 1452 cache multiple listing images

cade-exygy:
[1452](https://sfgovdt.jira.com/browse/DAH-1452)

Engineer Review Instructions

- Add a Listing Image to a listing in the full salesforce environment
- Go to PR environment (or localhost)
- Wait 10 minutes for the prefetch job to run. (we are running the prefetch job every ten minutes) Or if on local: Run the prefetch cache job following these instructions https://sfgovdt.jira.com/wiki/spaces/HOUS/pages/3468525649/Ruby+Debugging+Guide#Debugging-Ruby-Rake-jobs
- Once ran, go to the listing
- View the listing images, right click, and ensure the url is from cloudfront
- You can check the Postgres table and the S3 bucket as well if desired
Link to Postgres: https://data.heroku.com/dataclips/tctufwlnkivpmdbueefmsekcvorn
Link to S3 (might not have access): **https://s3.console.aws.amazon.com/s3/buckets/sf-gov-housing-dahlia-full?region=us-west-1&tab=objects**

More to come, ironing out a few things to be release ready
 - Need to add more logging
 - There is a long existing issue with our prefetch service where it does not correctly identify changed listings based on 'LastModifiedDate'. We can either use the prefetch:daily as our main job or change the 10 minute prefetch to always run the cache update. Realistically the jobs are doing the same thing. The main difference is the prefetch:daily runs a force=true to the salesforce cache.
 - Both of the rake jobs are relatively slow (has always been this way). May be opportunities to reduce unneeded calls to speed up execution time
 

Non-localhost testing ideas:
  - Either run the cache job directly from heroku? Not sure how to do this.
  - Or wait for the prefetch job to run - Runs at :01 every 10 minutes
  - Or change the frequency of prefetch daily in heroku for PR environment - our only options are 10 minute, hourly, or daily.
  
  Potential conflict between Photo_URL field caching and multiple listing images
   - investigating now, but long story short, we may need to add a boolean column to the listing_images table to indicate whether a listing_image record belongs to the Photo_URL field versus a listing image array item


db/migrate/20230914205521_add_column_and_index_to_listing_images.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumnAndIndexToListingImages < ActiveRecord::Migration[6.1]
```
cade-exygy, line 1: @jimlin-sfgov Pinging you on this since you mentioned you have some experience with this. Let me know your thoughts!

jimlin-sfgov: My only feedback is to make the migration file name more specific by including the name of the added column.

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,138 @@
...
+      if !resized_image(image_name, @resized_listing_images)
+        p '     resizing and uploading'
+        if resize_and_upload_image(li_raw_image_url, tmp_image_path, remote_image_path, @listing_id)
+          p '     create or update'
+          create_or_update_listing_image(@listing_id, image_url, li_raw_image_url)
+        end
+      else
+        create_or_update_listing_image(@listing_id, image_url, li_raw_image_url)
+      end
+      # elsif !listing_image_current(@listing_id, image_url)
```
cade-exygy, line 45: Need to add a feature toggle here

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,138 @@
...
+      )
+    end
+    File.delete(tmp_image_path)
+    uploaded
+  else
+    false
+  end
+end
+
+def resize_image(image_url, listing_id, tmp_image_path)
```
cade-exygy, line 108: Could pull this out into a shared service

lib/tasks/listing_images.rake
```diff
@@ -4,6 +4,7 @@ namespace :listing_images do
     listings = Force::ListingService.listings
```
cade-exygy, line 4: This job runs after a deploy

app/services/cache_service.rb
```diff
@@ -33,8 +34,12 @@ def cache_only_updated_listings
         l['Id'] == fresh_listing['Id']
       end
 
-      cache_single_listing(fresh_listing) unless
-        listing_unchanged?(prev_cached_listing, fresh_listing)
+      # TODO: find a differnt way to determin if listing is unchanged. This function seems to always return true.
```
jimlin-sfgov, line 37: Did you test this by modifying a field on a `Listing`, like the name?
Under the hood (i.e. in the Salesforce relational database), the `LastModifiedDate` is a column on the `Listings` table. If we update listing images, that will modify a different table, the `Listing_Images` table, which means `LastModifiedDate` would not change in that scenario.

It is possible to get an understanding of what is happening behind the Salesforce API layer by reading the code available to us in the Salesforce web interface: https://sfhousing--full.sandbox.lightning.force.com/lightning/setup/ApexClasses/page?address=%2F01pU0000001hD1h

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+def get_image_url(remote_image_path)
+  base_image_url = ENV['RESOURCE_URL']
+  "#{base_image_url}/#{remote_image_path}"
+end
+
+def get_tmp_image_path(image_name)
+  "tmp/images/#{image_name}"
+end
+
+def resized_image(image_name, resized_listing_images)
```
jimlin-sfgov, line 74: It's good practice to append `?` to the end of names  for methods that only return boolean values, with no side effects.

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+  resized_listing_images.count { |file| file.key.end_with?(image_name) }.positive?
+end
+
+def listing_image_current(listing_id, image_url)
```
jimlin-sfgov, line 78: It's good practice to append `?` to the end of names  for methods that only return boolean values, with no side effects.

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+  listing_images_by_id = ListingImage.where(salesforce_listing_id: listing_id)
+  current_listing_image = listing_images_by_id.select do |li|
+    li && li.image_url == image_url
+  end
+  current_listing_image&.length&.positive?
```
jimlin-sfgov, line 83: This method can be more compact and performant if we just stick to `ActiveRecord` methods:
```ruby
ListingImage.where(salesforce_listing_id: listing_id).where(image_url: image_url).exists?
```

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+end
+
+def resize_image(image_url, listing_id, tmp_image_path)
+  Dir.mkdir('tmp/images') unless Dir.exist?('tmp/images')
+  image = MiniMagick::Image.open(image_url)
+  p image
+  throw MiniMagick::Invalid unless image.valid?
+  # set width only and height is adjusted to maintain aspect ratio
+  image.resize(768.to_s)
```
jimlin-sfgov, line 111: Use constants for hardcoded numbers, like how it's done in ListingImageService.

app/services/cache_service.rb
```diff
@@ -52,14 +57,20 @@ def cache_single_listing(listing)
...
     Raven.capture_exception(e, tags: { 'listing_id' => listing['Id'] })
   end
 
+  def process_listing_images(listing)
+    Rails.logger.info("Calling ListingImageService for #{listing['Id']} ")
+    image_processor = ListingImageService.new(listing).process_image
+    Rails.logger.error image_processor.errors.join(',') if image_processor.errors.present?
+    Rails.logger.info("Calling MultipleListingImageService for #{listing['Id']} ")
+    multiple_listing_image_processor = MultipleListingImageService.new(listing).process_images
+    Rails.logger.error multiple_listing_image_processor&.errors&.join(',') if multiple_listing_image_processor&.errors&.present?
```
jimlin-sfgov, line 71: I don't see a need for the safe navigator operators `&.`, but correct me if I'm wrong.

jimlin-sfgov (changes requested):

app/services/force/listing_service.rb
```diff
@@ -122,5 +124,33 @@ def self.array_sort!(listing)
       end
       listings
     end
+
+    private_class_method def self.add_cloudfront_urls_for_listing_images(listings)
+      listing_images = ListingImage.all
+
+      listings.each do |listing|
```
jimlin-sfgov, line 131: We can make this `each` block more compact like this:
```ruby
next unless listing['Listing_Images'].present?
listing['Listing_Images'] = set_cloudfront_url(
  listing['Listing_Images'],
  ListingImage.where(salesforce_listing_id: listing['Id'])
)
```

app/services/force/listing_service.rb
```diff
@@ -122,5 +124,33 @@ def self.array_sort!(listing)
...
+        multiple_listing_images = set_cloudfront_url(listing['Listing_Images'],
+                                                     listing_images_array)
+        listing['Listing_Images'] = multiple_listing_images
+      end
+      listings
+    end
+
+    private_class_method def self.set_cloudfront_url(sf_listing_images, cf_listing_images)
+      sf_listing_images.each do |li|
+        cf_listing_image = cf_listing_images.select do |cf_li|
```
jimlin-sfgov, line 147: If you follow the recommendation I made for `self.add_cloudfront_urls_for_listing_images`, then the `cf_listing_images` argument will be a chain-able `ActiveRecord` relation, which allows us to make this code more compact:
```ruby
cf_listing_image = cf_listing_images.where(raw_image_url: li['Image_URL']).first
```

jimlin-sfgov (changes requested):

app/services/cache_service.rb
cade-exygy: LastModifiedDate does update in every scenario I have tested. The issue is that the prev_cached_listing and fresh_listing both have the same LastModifiedDate, causing this check to always return true. Thanks for linking that ApexClasses, that is insigtful!

app/services/cache_service.rb
```diff
@@ -33,8 +34,12 @@ def cache_only_updated_listings
...
       end
 
-      cache_single_listing(fresh_listing) unless
-        listing_unchanged?(prev_cached_listing, fresh_listing)
+      # TODO: find a differnt way to determin if listing is unchanged. This function seems to always return true.
+      unless listing_unchanged?(prev_cached_listing, fresh_listing)
+        Rails.logger.info("Listing is changed #{prev_cached_listing['Id']}")
+      end
+
+      cache_single_listing(fresh_listing)
```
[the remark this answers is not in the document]
cade-exygy: Yes, do to the above issue with the listing_unchanged check

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+  cache_string ||= Digest::MD5.hexdigest(listing_id + raw_image_url)
+  cache_string
+end
+
+def get_image_name(listing_id, cache_string)
+  "#{listing_id}-#{cache_string}.jpg"
+end
+
+def get_remote_image_path(image_name)
+  "images/listings/#{image_name}"
```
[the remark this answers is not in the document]
cade-exygy: The service was throwing errors for some reason when using constants. I'll look into it as a lower priority.

app/services/force/listing_service.rb
```diff
@@ -122,5 +124,33 @@ def self.array_sort!(listing)
...
+          cf_li.raw_image_url == li['Image_URL']
+        end.first
+        url = cf_listing_image && ENV['CACHE_LISTING_IMAGES'].to_s.casecmp('true').zero? ? cf_listing_image.image_url : li['Image_URL']
+        li['imageURL'] = url
```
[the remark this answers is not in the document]
cade-exygy: agreed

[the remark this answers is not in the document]
cade-exygy: maybe something like `displayImageURL` or `clientImageURL`?

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+          create_or_update_listing_image(@listing_id, image_url, li_raw_image_url)
+        end
+      elsif !listing_image_current(@listing_id,
+                                   image_url) || ENV['FORCE_MULTIPLE_LISTING_IMAGE_UPDATE'].to_s.casecmp('true').zero?
+        create_or_update_listing_image(@listing_id, image_url, li_raw_image_url)
+      end
+    end
+    self
+  end
+end
```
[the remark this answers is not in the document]
cade-exygy: addressed

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+    uploaded
+  else
+    false
+  end
+end
+
+def resize_image(image_url, listing_id, tmp_image_path)
+  Dir.mkdir('tmp/images') unless Dir.exist?('tmp/images')
+  image = MiniMagick::Image.open(image_url)
+  p image
```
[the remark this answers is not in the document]
cade-exygy: addressed

app/services/multiple_listing_image_service.rb
```diff
@@ -0,0 +1,136 @@
...
+  # TMP_DIR = 'tmp/images'.freeze
+  # REMOTE_IMAGE_PATH = 'images/listings'.freeze
+  # IMAGE_WIDTH = 768
+
+  def initialize(listing_hash)
+    @errors = []
+    @listing = listing_hash
+    @listing_id = listing_hash['Id']
+    @listing_images = listing_hash['Listing_Images']
+    @resized_listing_images = FileStorageService.find('images/listings')
```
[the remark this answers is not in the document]
cade-exygy: addressed

app/services/multiple_listing_image_service.rb
cade-exygy: addressed

app/services/multiple_listing_image_service.rb
cade-exygy: addressed

app/services/multiple_listing_image_service.rb
cade-exygy: addressed

app/services/cache_service.rb
cade-exygy: addressed

app/services/force/listing_service.rb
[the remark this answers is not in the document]
cade-exygy: Changed to `displayImageURL`

app/services/force/listing_service.rb
cade-exygy: addressed

app/services/force/listing_service.rb
cade-exygy: addressed

cade-exygy:
@jimlin-sfgov Thank you for your great suggestions! I addressed the majority. I will test through again and ensure things are working as expected. Let me know if you notice anything else!

app/services/multiple_listing_image_service.rb
cade-exygy: ran into some syntax errors, will update

app/services/cache_service.rb
jimlin-sfgov: I was able to verify that the `listing_unchanged?` method is working correctly, both locally and for the PR app.

## Local

1. make sure `CACHE_SALESFORCE_REQUESTS='true'` is in the local .env file
2. enable caching in dev environment by running `bundle exec rails dev:cache`
3. in `config/environments/development.rb`, change the `config.cache_store` value to `config.cache_store = :memory_store`
4. Update a listing name
5. Open the rails console `bundle exec rails c` and run this snippet
    ```ruby
    Force::ListingService.listings(subset: 'browse').find {|l| l['Id'] == 'a0W4U00000IgshdUAB'}.slice('Id', 'Name', 'LastModifiedDate')
    ```
6. You'll see that the `Name` and `LastModifiedDate` values are the older cached ones

## Heroku App

1. Open the console and start `bash`: https://dashboard.heroku.com/apps/dahlia-webapp-pr-1936/access?web-console=dahlia-webapp-pr-1936
2. Open the rails console `bundle exec rails c` and run this to verify that caching is enabled: `Rails.application.config.action_controller.perform_caching`
3. Run the snippet in the rails console:
    ```ruby
    Force::ListingService.listings(subset: 'browse').find {|l| l['Id'] == 'a0W4U00000IgshdUAB'}.slice('Id', 'Name', 'LastModifiedDate')
    ```
4. You'll see that the `Name` and `LastModifiedDate` values are the older cached ones

cade-exygy: Awesome! I will take a look.

jimlin-sfgov:
Do you plan on writing tests to a `multiple_listing_image_service_spec.rb` file for this PR?

app/services/cache_service.rb
cade-exygy: I was able to run this once locally, but wen I ran the snippet, I saw the updated values and not the older ones. Now I cannot run `bundle exec rails c` without getting this error: `Could not find cache store adapter for dalli_store (cannot load such file -- active_support/cache/dalli_store) (RuntimeError)`

app/services/cache_service.rb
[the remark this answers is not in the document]
jimlin-sfgov: Did you update `config.cache_store` in the development config file?

app/services/cache_service.rb
jimlin-sfgov: Did you update config.cache_store in the development config file?

cade-exygy: I do see this behavior in heroku pr env though. I ran browse and it displays the cached results. I then ran the browse again but this time with the force: true param. The latest changes come through as expected.

cade-exygy: So we could separate the listing image processing and the force=true calls to the Listing Service.

Or inside `cache_single_listing` check if `listing_unchanged`, before running the Force calls. And then call `process_listing_images`

The most accurate approach would be to check if the Listing_Images is different between prev_cached_listings and fresh_listings.

Something like:

```
def listing_images_unchanged?([prev_cached_listing, fresh_listing])
    prev_cached_listing_images = prev_cached_listing.present? && prev_cached_listing['Listing_Images']
    fresh_listing_images = fresh_listing.present? && fresh_listing['Listing_Images']
    # Compare attributes of each listing image
end
```

Any other ideas on next steps for this piece?

jimlin-sfgov: Based on my findings, `listing_unchanged?` is working normally. So we can revert all changes made to the `cache_only_updated_listings` in this PR.

Not sure what else we need to do besides reverting, unless I'm missing something?

cade-exygy: I think we would still need to run the process_images for the Listing Images array. If adding a ListingImage to the array in salesforce does not trigger `listing_unchanged?` to be false, then we will not update the ListingImages for that Listing.

jimlin-sfgov: > If adding a ListingImage to the array in salesforce does not trigger listing_unchanged? to be false

Oh, did you confirm that? I only mentioned that as a hypothesis, I didn't actually do any manual testing for that behavior.

app/services/cache_service.rb
```diff
@@ -43,6 +46,13 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
       (prev_cached_listing['LastModifiedDate'] == fresh_listing['LastModifiedDate'])
   end
 
+  def listing_images_unchanged?(prev_cached_listing, fresh_listing)
```
cade-exygy, line 49: Need to ensure that `==` check is sufficient or if we need to manually check attributes

cade-exygy:
> Do you plan on writing tests to a `multiple_listing_image_service_spec.rb` file for this PR?

Yes, I would like to! I am okay with moving forward with the PR though if the priority is to push to full

config/environments/development.rb
```diff
@@ -57,7 +57,7 @@
     }
 
     # These numbers are from the recommended dalli config: https://devcenter.heroku.com/articles/memcachier#ruby
-    config.cache_store = :dalli_store, nil, {
+    config.cache_store = :memory_store, nil, {
```
jimlin-sfgov, line 60: If we want to change this to use `:memory_store`, then we should not pass any additional arguments.
```ruby
 config.cache_store = :memory_store
```
vs
```ruby
 config.cache_store = :memory_store, nil, {...}
```

cade-exygy: Ah good catch, that change persisted from my local env, didn't intend to push that up yet. Changing back for now.

app/services/cache_service.rb
```diff
@@ -33,8 +34,13 @@ def cache_only_updated_listings
         l['Id'] == fresh_listing['Id']
       end
 
-      cache_single_listing(fresh_listing) unless
-        listing_unchanged?(prev_cached_listing, fresh_listing)
+      cache_single_listing(fresh_listing) unless listing_unchanged?(prev_cached_listing,
```
cade-exygy, line 37: After testing some different options, the evidence so far shows that `process_listing_images` will not get called correctly. My guess is that somewhere else is refreshing the Salesforce cache before we set `prev_cached_listings`. I will spend the day trying a few other things but I think we should move forward with:

- Calling `cache_single_listing` only if unchanged.
- Call `process_listing_images` every time

cade-exygy: Logging for a0W4U00000IgshdUAB:

app/services/cache_service.rb
```diff
@@ -43,7 +47,34 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
       (prev_cached_listing['LastModifiedDate'] == fresh_listing['LastModifiedDate'])
   end
 
+  def listing_images_unchanged?(prev_cached_listing, fresh_listing)
+    prev_cached_listing_images = prev_cached_listing.present? && prev_cached_listing['Listing_Images']
+    fresh_listing_images = fresh_listing.present? && fresh_listing['Listing_Images']
+
+    return true if fresh_listing_images.blank?
```
jimlin-sfgov, line 54: Does this condition ever actually happen?

app/services/cache_service.rb
```diff
@@ -43,7 +47,34 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
...
+    return false if prev_cached_listing_images&.length != fresh_listing_images&.length
+
+    unchanged = true
+    fresh_listing_images.each_with_index do |li, idx|
+      unchanged = listing_image_unchanged?(prev_cached_listing_images[idx], li)
+      break unless unchanged
+    end
+    Rails.logger.info("Listing Images for  #{fresh_listing['Id']} is unchanged: #{unchanged}")
+    unchanged
+  end
```
jimlin-sfgov, line 67: There might be a slight performance hit but this method could be more flexible and easier to read if we compare arrays of key-value pairs.

```ruby
def foobar(fresh_listing_images, prev_cached_listing_images)
  foo = fresh_listing_images.map { |li| li.slice('Id', 'Image_URL') }
  bar = prev_cached_listing_images.map { |li| li.slice('Id', 'Image_URL') }
  Rails.logger.info("foo #{foo}, bar #{bar}")
  (foo - bar).empty? && (bar - foo).empty?
end
```

app/services/cache_service.rb
cade-exygy: Yes, if someone deletes the last listing image in the array, the array could be empty

app/services/cache_service.rb
cade-exygy: No we have checks in the MultipleListingImage service that check if we need to update before processing

app/services/cache_service.rb
```diff
@@ -51,22 +51,15 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
...
-    end
-    Rails.logger.info("Listing Images for  #{fresh_listing['Id']} is unchanged: #{unchanged}")
-    unchanged
+    fresh_li_slice = fresh_listing_images&.map { |li| li.slice('Id', 'Image_URL') }
+    prev_li_slice = prev_cached_listing_images&.map { |li| li.slice('Id', 'Image_URL') }
+    Rails.logger.info("fresh #{fresh_li_slice}, prev #{prev_li_slice}")
+    (fresh_li_slice - prev_li_slice).empty? && (prev_li_slice - fresh_li_slice).empty?
   end
 
   def listing_image_unchanged?(prev_cached_listing_li, fresh_li)
```
jimlin-sfgov, line 65: This method is now unused, it can be removed.

lib/tasks/listing_images.rake
```diff
@@ -1,9 +1,11 @@
 namespace :listing_images do
   desc 'Process Listing Images'
   task process_images: :environment do
+    p 'Running Process Listing Images'
```
jimlin-sfgov, line 4: Debugging line?

app/services/cache_service.rb
```diff
@@ -43,7 +50,21 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
...
 
+  def listing_images_unchanged?(prev_cached_listing, fresh_listing)
+    prev_cached_listing_images = prev_cached_listing&.dig('Listing_Images')
+    fresh_listing_images = fresh_listing_images&.dig('Listing_Images')
+
+    return true if fresh_listing_images.blank?
+
+    fresh_li_slice = fresh_listing_images&.map { |li| li.slice('Id', 'Image_URL') }
+    prev_li_slice = prev_cached_listing_images&.map { |li| li.slice('Id', 'Image_URL') }
+    Rails.logger.info("fresh #{fresh_li_slice}, prev #{prev_li_slice}")
```
jimlin-sfgov, line 61: This was copy-pasted from my sample code, I would adjust it to be more descriptive, and/or remove it if you see no benefit.

Also, for some reason I never saw this log when I was testing in the PR environment, not sure why.

jimlin-sfgov (changes requested):
I tested in the PR environment and it looks good. Just needs a little bit of cleanup and should be good to go.

app/services/cache_service.rb
```diff
@@ -43,7 +50,21 @@ def listing_unchanged?(prev_cached_listing, fresh_listing)
       (prev_cached_listing['LastModifiedDate'] == fresh_listing['LastModifiedDate'])
   end
 
+  def listing_images_unchanged?(prev_cached_listing, _fresh_listing)
+    prev_cached_listing_images = prev_cached_listing&.dig('Listing_Images')
+    fresh_listing_images = fresh_listing_images&.dig('Listing_Images')
```
jimlin-sfgov, line 55: This line used to be 
```ruby
fresh_listing_images = fresh_listing.present? && fresh_listing['Listing_Images']
```

What happened to using the `fresh_listing` argument? Now, it looks like `fresh_listing_images&.dig` will always be `nil` because it's undefined right?

jimlin-sfgov (approved):
Looks good to me 👍 

Not sure if you've already addressed these things from the PR description:
> More to come, ironing out a few things to be release ready

PR #1943, opened 2023-09-27
Hotfix: Add 750 Harrison Unit 604

cliu02:
Hotfix: This modifies the SRO logic for 750 Harrison Unit 604.

### DIFF
--- a/cypress/e2e/listingDetails/openListings.e2e.ts
+++ b/cypress/e2e/listingDetails/openListings.e2e.ts
@@ -29,7 +29,7 @@ function verifyListing(language, id, altPhotoText, title, address, applyButtonTe
 const testListings = {
   OPEN_RENTAL: {
     id: "a0W0P00000F8YG4UAN",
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/listings/ListingConstantsService.js.coffee
```diff
@@ -114,6 +114,7 @@ ListingConstantsService = () ->
     'a0W4U00000HlubxUAB': '1335 Folsom Street'
     'a0W4U00000KGFDWUA5': '1335 Folsom Street'
     'a0W4U00000KKtXyUAL': '750 Harrison'
+    'a0W4U00000IXVL8UAP': '750 Harrison Unit 604',
```
alulabeshue-sfgov, line 117: For 1335 Folsom there are multiple sales force listing IDs (keys) that are using the same name (value). If a line was added like `'a0W4U00000IXVL8UAP': '750 Harrison'` I think you would only need this change and to update the env variables for react.

PR #1956, opened 2023-10-03
DAH-1717: Lottery search accessibility 

cliu02:
[DAH-1717](https://sfgovdt.jira.com/browse/DAH-1717)

This PR includes Christina's additional [recommendations](https://exygy.slack.com/archives/CCV0019TN/p1695744890604189?thread_ts=1695066815.678589&cid=CCV0019TN ) after testing with JAWS.

[DAH-1717]: https://sfgovdt.jira.com/browse/DAH-1717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jimlin-sfgov:
It looks like all of the requested changes in the Jira ticket have been addressed by previously merged PRs.

I cannot access the Exygy Slack link, so I really have no idea what this PR is supposed to accomplish.

jimlin-sfgov:
I noticed that the rendered html contains nested `font` elements
Screenshot 2023-10-04 at 2 07 59 PM

I'm wondering if this is intended, and if it will affect screen readers.

Also, `<font>` has been deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font

cade-exygy:
> I noticed that the rendered html contains nested `font` elements Screenshot 2023-10-04 at 2 07 59 PM
> 
> I'm wondering if this is intended, and if it will affect screen readers.
> 
> Also, `<font>` has been deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font

Does this come from Google Translate? If so, we do not have much control over these elements being inserted in the DOM

cade-exygy (approved):
Looks good! Using the Xcode Accessibility Inspector, I can screen read the content.

jimlin-sfgov:
> Does this come from Google Translate

Possibly. If the screenreaders are working, then this is a non-issue.

jimlin-sfgov (approved):

PR #1971, opened 2023-10-16
DAH-1762 Add EMpowerSF url

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1762


cade-exygy (commented):
URL on DAHLIA works fine, but the link goes to an unavailable page on SF.gov

jimlin-sfgov:
@cade-exygy The page is still under construction, so content is not available at this time.

cade-exygy (approved):

PR #1974, opened 2023-10-18
DAH-1801 Use image array for Application Intro page

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1801

### DIFF
--- a/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
+++ b/app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
@@ -75,6 +75,12 @@ ShortFormApplicationController = (
 
   $scope.emailRegex = SharedService.emailRegex
 
+  $scope.propertyCardImageURL = ->
+    if _.isArray($scope.listing.Listing_Images)
+      return $scope.listing.Listing_Images[0].Image_URL
+    else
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/short-form/ShortFormApplicationController.js.coffee
```diff
@@ -75,6 +75,12 @@ ShortFormApplicationController = (
 
   $scope.emailRegex = SharedService.emailRegex
 
+  $scope.propertyCardImageURL = ->
+    if _.isArray($scope.listing.Listing_Images)
+      return $scope.listing.Listing_Images[0].Image_URL
```
alulabeshue-sfgov, line 80: The [directory](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/javascript/modules/listings/SharedHelpers.tsx#L70) and [details](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx#L19) page are both using the field `displayImageURL`. Looking at some responses it looks to be the same URL as `Image_URL` but to ensure we are using the same images I think we can use the same field.

alulabeshue-sfgov (approved):

PR #1976, opened 2023-10-19
DAH-1789 Lottery API workaround for React

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1789

The Rails part of the workaround is not in this PR because it has already been added to the `production` branch via hotfix.


app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchFormWithoutBuckets.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { useState } from "react"
```
jimlin-sfgov, line 1: This entire file is a copy of `ListingDetailsLotterySearchForm.tsx`, except the `lotteryBucketDetails` prop has been removed.

jimlin-sfgov:
After further investigation, we have found a better and more universal way to work around the API issue. Closing this PR.

PR #1988, opened 2023-10-31
Hotfix enable preview flag on react

cade-exygy:
JIRA: https://sfgovdt.jira.com/browse/DAH-1828

## Changes

- In listing-detail.tsx we added functionality to allow the parsing of the preview=true flag for the react version of the listing details page

## Test Instructions

- Go to /listings/a0W8H0000014LSAUA2?react=true

- In network tab, ensure that the api call does not have the force=true flag

- Go to /listings/a0W8H0000014LSAUA2?react=true&preview=true`

- Ensure that the api call has the force=true flag

- You can also update a listing and salesforce and run the above test to see if the salesforce cache updates properly.

### Open Question

- Is there a cleaner way to parse queryParams and pass them to the api call?

### DIFF
--- a/app/javascript/api/apiEndpoints.ts
+++ b/app/javascript/api/apiEndpoints.ts
@@ -1,4 +1,5 @@
-export const listing = (listingId) => `/api/v1/listings/${listingId}.json`
+export const listing = (listingId, preview = false) =>
+  `/api/v1/listings/${listingId}.json?force=${preview ? "true" : "false"}`
 export const listingPreferences = (listingId) => `/api/v1/listings/${listingId}/preferences`
 export const listingUnits = (listingId) => `/api/v1/listings/${listingId}/units`
 export const lotteryBuckets = (listingId) => `/api/v1/listings/${listingId}/lottery_buckets`
--- a/app/javascript/api/apiEndpoints.ts
+++ b/app/javascript/api/apiEndpoints.ts
@@ -1,5 +1,4 @@
-export const listing = (listingId, preview = false) =>
-  `/api/v1/listings/${listingId}.json?force=${preview ? "true" : "false"}`
+export const listing = (listingId) => `/api/v1/listings/${listingId}.json`
 export const listingPreferences = (listingId) => `/api/v1/listings/${listingId}/preferences`
 export const listingUnits = (listingId) => `/api/v1/listings/${listingId}/units`
 export const lotteryBuckets = (listingId) => `/api/v1/listings/${listingId}/lottery_buckets`
--- a/app/javascript/api/listingApiService.ts
+++ b/app/javascript/api/listingApiService.ts
@@ -19,8 +19,8 @@ type ListingPreferencesResponse = { preferences: RailsListingPreference[] }
 type ListingUnitsResponse = { units: RailsUnit[] }
 type ListingAmiChartsResponse = { ami: RailsAmiChart[] }
 
-export const getListing = async (listingId?: string): Promise<RailsListing> =>
-  get<ListingsResponse>(listing(listingId)).then(({ data }) => data.listing)
+export const getListing = async (listingId?: string, preview?: boolean): Promise<RailsListing> =>
+  get<ListingsResponse>(listing(listingId, preview)).then(({ data }) => data.listing)
 
 /**
  * Get the lottery buckets with rankings for the given listing
--- a/app/javascript/api/listingApiService.ts
+++ b/app/javascript/api/listingApiService.ts
@@ -19,8 +19,8 @@ type ListingPreferencesResponse = { preferences: RailsListingPreference[] }
 type ListingUnitsResponse = { units: RailsUnit[] }
 type ListingAmiChartsResponse = { ami: RailsAmiChart[] }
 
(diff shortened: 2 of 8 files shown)


app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx
```diff
@@ -53,7 +53,7 @@ const createImageCardProps = (listing: RailsListing) => {
         }
       : {
           props: {
-            iamgeUrl: fallbackImg,
```
cade-exygy, line 56: Included this commit to fix typo

app/javascript/pages/listings/listing-detail.tsx
```diff
@@ -74,10 +74,12 @@ const ListingDetail = () => {
 
   useEffect(() => {
     const path = getPathWithoutLanguagePrefix(router.pathname)
-    void getListing(path.split("/")[2]).then((listing: RailsListing) => {
+    const previewFlag = router.asPath.includes("preview=true")
```
jimlin-sfgov, line 77: If you were looking to make this a little cleaner, you could leverage [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get#examples) in our `Router.ts` file:

```js
// ...
get asPath(): string {
  return window.location.pathname + window.location.search
},
get search(queryParam: string): string | null {
  return new URLSearchParams(window.location.search).get(queryParam)
}
// ...
```

Although this is more code, and I'm not sure if it's worth implementing.

app/javascript/api/apiEndpoints.ts
```diff
@@ -1,4 +1,5 @@
-export const listing = (listingId) => `/api/v1/listings/${listingId}.json`
+export const listing = (listingId, preview = false) =>
+  `/api/v1/listings/${listingId}.json?force=${preview ? "true" : "false"}`
```
jimlin-sfgov, line 2: We should generalize the implementation of the `force=true` param so that it works for all relevant endpoints, similar to how it is done in the Angular code.

We can do this by modifying the `get` function in our `apiService.ts` file:

```js
export const get = <T>(url: string, config?: AxiosRequestConfig): Promise<AxiosResponse<T>> => {
  const foo = { ...config.params }
  if (window.location.search.includes('preview=true')) {
    foo.force = 'true'
  }
  return axios.get(url, {...config, params: foo} )
}
```

jimlin-sfgov: If you want to be a little less general, you could update the functions in `*ApiService.ts` files in a similar fashion.

jimlin-sfgov: also, if you want to know which endpoints support the flag in Angular, you can grep for `forceRecache` in the repo

app/javascript/pages/listings/listing-detail.tsx
```diff
@@ -74,6 +74,7 @@ const ListingDetail = () => {
 
   useEffect(() => {
     const path = getPathWithoutLanguagePrefix(router.pathname)
+
```
jimlin-sfgov, line 77: we should remove this to minimize diffs

jimlin-sfgov (commented):
Because we've modified the `get` function, some API calls are using the flag even if it is not supported, for example on the directory page:
Screenshot 2023-11-01 at 1 32 13 PM

For this specific endpoint it is okay because we filter out unsupported params in Rails: https://github.com/SFDigitalServices/sf-dahlia-web/blob/c52b2209822ef37ce2ff113b4a6e2673f49c67d8/app/controllers/api/v1/listings_controller.rb#L82

I'm not sure if there are any other API calls that don't support the `prevview=true` flag.

app/javascript/api/listingApiService.ts
```diff
@@ -19,8 +19,15 @@ type ListingPreferencesResponse = { preferences: RailsListingPreference[] }
...
-  get<ListingsResponse>(listing(listingId)).then(({ data }) => data.listing)
+export const getListing = async (listingId?: string): Promise<RailsListing> => {
+  let force = "false"
+  if (window.location.search.includes("preview=true")) {
+    force = "true"
+  }
+  return get<ListingsResponse>(listing(listingId), { params: { force } }).then(
+    ({ data }) => data.listing
+  )
+}
```
jimlin-sfgov, line 30: Did you check for other endpoints that need to support the flag?

I mentioned in a previous comment that we should check the Angular code to see which endpoints support the `preview=true` params, here are my results:
Screenshot 2023-11-02 at 9 46 07 AM

cade-exygy: This ticket is just for ListingDetails on React so that we can unblock the release. We can create a new ticket to handle any other endpoints or params

jimlin-sfgov (approved):
Looks good.

There is no functional difference, but all of our API calls will now have `?force=false` appended to them by default.

app/javascript/api/listingApiService.ts
jimlin-sfgov: > We can create a new ticket

When you create the ticket, please mark it as a higher priority. To my knowledge, the three other endpoints are all used on the Listing Details page.

PR #1992, opened 2023-11-02
DAH-1832 feat: filter out Veterans Preferences from ListingDetailsPre…

cade-exygy:
Jira: https://sfgovdt.jira.com/browse/DAH-1832
# Changes
- Filter our preferences to exclude any preferences with a `preferenceName` that includes "Veterans"

# Review Instructions
- Go to a listing with Veteran Preferences applied: /listings/a0W8H0000014Yo4UAE?react=true
- Scroll down to Lottery Preferences and ensure that no Veteran preference types appear
- Click View Lottery Results button and you should see Veteran Preferences here

Demo: https://www.loom.com/share/f92306c70d064134a27478a8050b54ba?sid=0eadfb80-115f-474c-9565-27b14c5b27ef


app/javascript/modules/listingDetails/ListingDetailsPreferences.tsx
```diff
@@ -36,7 +36,9 @@ export const ListingDetailsPreferences = ({ listingID }: ListingDetailsPreferenc
 
   useEffect(() => {
     void getPreferences(listingID).then((preferences) => {
-      setPreferences(preferences)
+      setPreferences(
+        preferences?.filter((preference) => !preference.preferenceName.includes("Veterans"))
```
jimlin-sfgov, line 40: We're using `includes` because there will be multiple preferences related to veterans, correct? Are we sure that the  "V" will be capitalized and that it will always be plural?

cade-exygy: Yeah that is correct. All of the names do contain specifically "Veterans", but I can add a lowercase to be more resilient.

jimlin-sfgov (approved):

PR #1994, opened 2023-11-07
DAH-1841 fix: use alt text from Salesforce

cade-exygy:
JIRA: https://sfgovdt.jira.com/browse/DAH-1841

# Changes
- For the listing directory image we now use the image description field from the Salesforce listing image object as alt text

# Review Instructions
- Go to /for-rent
- Scroll to a listing with multiple listing images (Casa Adelante)
- Inspect the cover image and ensure alt text is the same as what is set in Salesforce

### DIFF
--- a/app/javascript/modules/listings/SharedHelpers.tsx
+++ b/app/javascript/modules/listings/SharedHelpers.tsx
@@ -70,12 +70,16 @@ export const getImageCardProps = (listing: RailsListing, hasFiltersSet?: boolean
       ? listing.Listing_Images[0].displayImageURL
       : listing?.imageURL ?? fallbackImg
 
+  const imageDescription =
+    listing?.Listing_Images?.length > 0
+      ? listing.Listing_Images[0].Image_Description
+      : `${listing.Building_Name} Building`
   return {
     imageUrl: imageUrl,
     href: `/listings/${listing.listingID}`,
     tags: getTagContent(listing),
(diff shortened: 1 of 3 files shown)


chadbrokaw (approved):

app/javascript/modules/listings/SharedHelpers.tsx
```diff
@@ -70,12 +70,16 @@ export const getImageCardProps = (listing: RailsListing, hasFiltersSet?: boolean
       ? listing.Listing_Images[0].displayImageURL
       : listing?.imageURL ?? fallbackImg
 
+  const imageDescription =
+    listing?.Listing_Images?.length > 0
+      ? listing.Listing_Images[0].Image_Description
+      : `${listing.Building_Name} Building`
```
chadbrokaw, line 76: I know that this was like this before, but I'm wondering if this text should be translated?

cade-exygy: Yeah we could dynamic translate the salesforce alt text and provide translations for the fallback.

cade-exygy: Adding the `useTranslate()` hook call to a parent component ( I placed it in `RentDirectory` ) seems to work. But not sure if this is the best way? I can create a ticket. In Angular it looks like we are translating alt text successfully.

PR #1996, opened 2023-11-08
Hotfix- Change ListingImage error to an info log

cade-exygy:
# Changes
- "No listing images provided for listing" error is now an info statement

# Review
- In Papertrail ensure that this log does not throw an error: https://my.papertrailapp.com/systems/dahlia-prod-test/events


jimlin-sfgov (approved):
Looks good. I don't think Papertrail can distinguish between `Rails.logger.info` and `Rails.logger.error`, but the important thing is that we are no longer using the string `'ListingImageService error'`, which is what triggers our slack notifications.

PR #1999, opened 2023-11-09
Fix: Only pass force param if preview flag is present

cade-exygy:
# Changes
- Refactored logic for adding `force=true` to api requests

# Review 
- Go to a listing detail page in react
- Open network tab
- Add `preview=true` to url and ensure that the force param is applied
- Add `preview=false` and ensure that force param is not applied
- Update listing details in Salesforce and ensure changes are updated on React (Will only work on prod env until we merge production branch back into main)


jimlin-sfgov:
Is there a ticket for this PR? If not, can you explain the motivation for this PR, e.g. what are the benefits of this refactor?

Also, the branch name is prefixed with `hotfix` but the target branch is `main`, is this intentional?

cade-exygy:
Yeah it is small bug found by @alulabeshue-sfgov from testing the react force update. The intention and benefits is that we do not pass the `force=false` param in the api request when no preview param is present. This will make it more difficult for the public to see the force flag. Since it is not a high priority, we decided to push to main and wait for the next deployment. We should also start pushing hotfixes to the main branch first when possible.

jimlin-sfgov:
> it is small bug found by @alulabeshue-sfgov 

I can't find any written mentions of the bug, so I don't know what it is. I'll let @alulabeshue-sfgov review this one.

alulabeshue-sfgov:
@cade-exygy It looks like for the units endpoint the frontend is not correctly setting the force param.

Screenshot 2023-11-13 at 2 59 54 PM

Screenshot 2023-11-13 at 3 00 12 PM

cade-exygy:
> @cade-exygy It looks like for the units endpoint the frontend is not correctly setting the force param.
> 
> Screenshot 2023-11-13 at 2 59 54 PM Screenshot 2023-11-13 at 3 00 12 PM

Good catch, should be fixed now

alulabeshue-sfgov (approved):

cade-exygy:
> There are some failing tests. But once you address those everything else looks good.

e2e-production is failing probably due to merging into main? And e2e is awaiting the Phrase merge

PR #2002, opened 2023-11-14
Bring in new Phrase translations

chadbrokaw:
This brings in new Phrase translations that were lost during the release process. The code should now be in sync with Phrase


cade-exygy (approved):
Changes look consistent!

PR #2004, opened 2023-11-14
DAH-1856 fix: add translate class to ImageCard, fix tests

cade-exygy:
JIRA: https://sfgovdt.jira.com/browse/DAH-1856

# Changes
 - Pass `translate` class as a prop to `ImageCard` to translate listing image alt text coming from Salesforce
 - Updated test file to check if the alt text is different than the english version, which signifies that a translation occurred.

Related Bloom UIC change: https://github.com/bloom-housing/ui-components/pull/137

# Review
- Go to listings/a0W4U00000IgrACUAZ?react=true
- Inspect alt text for the image
- Switch to a different language
- Ensure alt text is translated


cypress/e2e/listingDetails/openListings.e2e.ts
```diff
@@ -63,44 +56,41 @@ describe("Listing Details for Open Listings", () => {
...
     it("displays in Spanish", () => {
-      verifyListing(
-        "es",
-        testListings.OPEN_RENTAL.id,
-        ALT_PHOTO_TEXT.es,
-        testListings.OPEN_RENTAL.title,
-        null,
-        APPLY_BUTTON_TEXT.es
-      )
+      const langPart = "en"
```
chadbrokaw, line 67: Should this be `es`?

chadbrokaw: Same for below

app/javascript/modules/listingDetails/ListingDetailsImageCard.tsx
```diff
@@ -71,7 +71,7 @@ export const ListingDetailsImageCard = ({ listing }: ListingDetailsImageCardProp
   const { fallbackUsed, props: imageCardProps } = createImageCardProps(listing)
 
   return (
-    <header className="image-card--leader">
+    <header className="image-card--leader translate">
```
chadbrokaw, line 74: Does adding this `translate` class force the machine translation of the tag content?

cade-exygy: good catch, yes. I can try applying the `translate` class to only the `figure class="image_card__inner" element (in Bloom UIC). I'll research how to do that, unless you know a good way to apply that class without modifying any Bloom UIC components?

chadbrokaw (approved):

PR #2005, opened 2023-11-15
DAH-1869 fix open house ordering

cade-exygy:
Jira: https://sfgovdt.jira.com/browse/DAH-1869

# Changes
- use `Date.getTime()` to sort the combined date and start time string for the event

# Review
- Go to a Sales listing: /listings/a0W0P00000HbrEQUAZ?react=true
- Ensure that the open houses are correctly ordered by date and start time
- You can compare with the Angular version: /listings/a0W0P00000HbrEQUAZ
- You can edit the open house list in Salesforce as well to test

# Notes
- We can push this fix to main first, and then either cherry pick or create a PR for production.


jimlin-sfgov (changes requested):
Can we avoid using `lodash` in React code? It is not an explicit dependency, and we don't use it anywhere else in the React code. We have been using javascript's built in `sort` function in the React code.

jimlin-sfgov (approved):
Looks good.

In the future we can look into using a time library like [dayjs]( https://github.com/iamkun/dayjs) instead of writing time calculations ourselves.

PR #2007, opened 2023-11-16
DAH-1867: Add rental paragraph break

cliu02:
[DAH-1867](https://sfgovdt.jira.com/browse/DAH-1867)

[DAH-1867]: https://sfgovdt.jira.com/browse/DAH-1867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/__tests__/components/__snapshots__/BeforeApplyingForSale.test.tsx.snap
+++ b/app/javascript/__tests__/components/__snapshots__/BeforeApplyingForSale.test.tsx.snap
@@ -105,41 +105,41 @@ exports[`BeforeApplyingForSale display Before Applying when type is directory 1`
       <li
         class="text-gray-750 primary-lighter-markup-link"
       >
-        <p>
+        <span>
           Meet our 
           <a
-            href="https://sfmohcd.org/homebuyer-program-eligibility"
+            href="https://sf.gov/determine-if-you-can-buy-affordable-housing-program"
             target="_blank"
           >
             income requirements
           </a>
-        </p>
+        </span>
       </li>
       <li
         class="text-gray-750 primary-lighter-markup-link"
       >
-        <p>
+        <span>
           Complete 
           <a
-            href="https://sfmohcd.org/homebuyer-application-requirements#education"
+            href="https://sf.gov/sign-complete-homebuyer-education"
             target="_blank"
           >
             homebuyer education
           </a>
-        </p>
+        </span>
       </li>
       <li
         class="text-gray-750 primary-lighter-markup-link"
       >
-        <p>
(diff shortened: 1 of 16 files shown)


app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
```diff
@@ -307,7 +307,12 @@ export const ListingDetailsEligibility = ({
         {isRental(listing) && (
           <ListSection
             title={t("listingsForRent.rentalAssistance.title")}
-            subtitle={t("listingsForRent.rentalAssitance.subtitle")}
+            subtitle={
+              <>
+                <div className="mb-4">{t("listingsForRent.rentalAssistance.info1")}</div>
+                {t("listingsForRent.rentalAssistance.info2")}
```
jimlin-sfgov, line 313: For consistency, please wrap this string in a `div` as well.

No need to address in this PR, but it doesn't feel right to have `div` elements nested inside the inline `span` element, maybe an accessibility audit will flag this in the future? For some reason the "Household Maximum Income" section never got flagged during an audit, it has several `divs`, an `ul`, and `li`s all nested in a `span` 😲

jimlin-sfgov (changes requested):

app/javascript/modules/listingDetails/ListingDetailsEligibility.tsx
jimlin-sfgov: Tangential to this PR, but putting it down before I forget:

This bloom component could do with some changes: https://github.com/bloom-housing/ui-components/blob/main/src/sections/ListSection.tsx

1. `<span>` might be better off replaced with `<p>`
2. to avoid nesting non-string subtitles, do something like this:
    ```js
    <hgroup>
      <h3 className="custom-counter__title">{props.title}</h3>
      {typeof props.subtitle == 'string' ? (
        <span className="custom-counter__subtitle">{props.subtitle}</span>
      ) : (
        <>
          props.subtitle
        </>
      )}
    </hgroup>
    ```

jimlin-sfgov: @cliu02 Yes, I mean the translation string, like this
```js
  <>
    <div className="mb-4">{t("listingsForRent.rentalAssistance.info1")}</div>
    <div>{t("listingsForRent.rentalAssistance.info2")}</div>
```

jimlin-sfgov (approved):

PR #2008, opened 2023-11-17
DAH-1878 fix: add safety checks before iteration

cade-exygy:
JIRA: https://sfgovdt.jira.com/browse/DAH-1878

# Changes
- Add null checks for when units or ami api response returns `null` or empty array

# Review
- On Full, this sales listing used to throw an error upon render: /listings/a0W4U00000IgshdUAB?react=true
- Ensure that this page renders without errors
- You can also following the Engineering Notes to mock an empty units response.


jimlin-sfgov (approved):
looks good!

PR #2010, opened 2023-11-21
DAH-1879 Fix radio buttons on Educator Screening Question

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1879


alulabeshue-sfgov (approved):
LGTM

PR #2011, opened 2023-11-22
DAH-1885 Job code error message for Screen Readers

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1885


jimlin-sfgov:
Thanks @cliu02 That is also what I'm seeing. The error message is not getting automatically announced, but this is the same behavior as some other pages in application forms.

I think we should just move to PA testing so we can get someone with JAWS to try it out.

PR #2014, opened 2023-11-28
DAH-1869 fix: use dayjs to handle various startTime formats

cade-exygy:
Jira: https://sfgovdt.jira.com/browse/DAH-1869

# Changes
- use dayjs to convert date string to Dayjs object
- use dayjs.diff to sort dates

# Review
-Go to a Sales listing: /listings/a0W4U00000KneiUUAR?react=true
-Ensure that the open houses are correctly ordered by date and start time
-You can compare with the Angular version: /listings/a0W4U00000KneiUUAR
-You can edit the open house list in Salesforce as well to test

### DIFF
--- a/app/javascript/api/listingApiService.ts
+++ b/app/javascript/api/listingApiService.ts
@@ -19,11 +19,11 @@ type ListingPreferencesResponse = { preferences: RailsListingPreference[] }
 type ListingUnitsResponse = { units: RailsUnit[] }
 type ListingAmiChartsResponse = { ami: RailsAmiChart[] }
 
-const forceRecache = (): boolean => window.location.search.includes("preview=true")
+const forceRecache = () => (window.location.search.includes("preview=true") ? { force: true } : {})
 
 export const getListing = async (listingId?: string): Promise<RailsListing> => {
   const httpConfig = { params: {} }
-  httpConfig.params = { force: forceRecache() }
+  httpConfig.params = forceRecache()
   return get<ListingsResponse>(listing(listingId), httpConfig).then(({ data }) => data.listing)
 }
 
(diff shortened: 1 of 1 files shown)


app/javascript/util/listingUtil.ts
```diff
@@ -181,11 +181,13 @@ const formatEventTime = (eventTime: string) => {
 
 export const getEventDateTime = (eventDate: string, eventTime: string) => {
   const startTime = eventTime?.includes(":") ? eventTime : `${formatEventTime(eventTime)}`
-  return new Date(`${eventDate} ${startTime}`)
+  const customParseFormat = require("dayjs/plugin/customParseFormat")
```
jimlin-sfgov, line 184: we should stick to `import` for consistency

jimlin-sfgov (changes requested):

jimlin-sfgov:
In the future, let's hold off on merging approved PRs until they pass PA testing. That way we avoid multiple PRs per ticket, like this one.

cade-exygy:
> In the future, let's hold off on merging approved PRs until they pass PA testing. That way we avoid multiple PRs per ticket, like this one.

Fair point. We may have to adjust how we do PA. I believe PA testing is primarily done on the full instance. Also, we will have to make some tweaks to ensure that the PR deployment environment has the same vars and plugins as Full.

jimlin-sfgov:
> I believe PA testing is primarily done on the full instance

Yeah, we will have to do some tweaks before fully moving to PR instances for PA testing. For PRs that do not rely on environment vars etc., I have been commenting in the Jira ticket with the PR instance URL for PA testers to use.

jimlin-sfgov (approved):
Looks good.

The November ones are out of order, but that is because of inconsistent user input in salesforce.

The custom parsing format here handles what salesforce users _usually_ use, which is the best we can do on the front end.

cade-exygy:
> Looks good.
> 
> The November ones are out of order, but that is because of inconsistent user input in salesforce.
> 
> The custom parsing format here handles what salesforce users _usually_ use, which is the best we can do on the front end.

Awesome! Moved ticket to PA.

PR #2015, opened 2023-11-28
Hotfix prod remove listing image log

cade-exygy:
My guess is if a listing doesn't have any value for the Photo URL field, the ListingImageService throws this error. So if a new listing only uses Listing Images, then this will happen.

Logs: https://my.papertrailapp.com/groups/4549952/events?q=%22ListingImageService+error%22

For example, this listing https://housing.sfgov.org/listings/a0W4U00000NlS7DUAV does not seen to have the Building_URL field.

### DIFF
--- a/app/services/listing_image_service.rb
+++ b/app/services/listing_image_service.rb
@@ -18,7 +18,7 @@ class ListingImageService
 
   def process_image
     unless raw_image_url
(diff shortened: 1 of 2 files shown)


jimlin-sfgov (approved):
looks good!

PR #2017, opened 2023-11-30
DAH-1720 provide contrast for filter dropdown text label

tallulahkay:
[DAH-1720](https://sfgovdt.jira.com/browse/DAH-1720)

Provides contrast for the filter selection text in the language `Select` component of the react housing counselors view:

Screenshot 2023-11-30 at 11 08 26 AM

[DAH-1720]: https://sfgovdt.jira.com/browse/DAH-1720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/pages/getAssistance/counselor-filter.scss
+++ b/app/javascript/pages/getAssistance/counselor-filter.scss
@@ -19,7 +19,7 @@ select {
     }
     select {
         border-color: var(--bloom-color-gray-500);
-        @apply text-gray-700;
+        @apply text-gray-750;
     }
     input::placeholder {
         @apply overflow-visible;


app/javascript/pages/getAssistance/counselor-filter.scss
```diff
@@ -19,7 +19,7 @@ select {
     }
     select {
         border-color: var(--bloom-color-gray-500);
-        @apply text-gray-700;
+        color: #555555;
```
jimlin-sfgov, line 22: We should use the built-in style values instead of hardcoded ones. In this case we just need to change it to `@apply text-gray-750`

jimlin-sfgov (changes requested):

jimlin-sfgov:
A few things about best practices for creating PRs (they're missing from the confluence docs, so we'll add them there):
- add a link to the Jira ticket in the PR description
- in the linked Jira ticket, fill out the "Review Instructions" section

app/javascript/pages/getAssistance/counselor-filter.scss
jimlin-sfgov: For component-specific stylesheets, we've been using `@apply` instead of `color` for text. I'm not sure why we do it one way or the other, but we should keep it consistent unless there is a good reason to change.

tallulahkay: thanks!

tallulahkay:
@jimlin-sfgov updated!

PR #2019, opened 2023-12-01
DAH-1869 fix open house ordering (#2005)

cade-exygy:
* DAH-1869 fix: sort open houses with sortBy

* DAH-1869 fix: snapshot tests

* DAH-1869 feat: remove lodash, add custom dateTime sort

* DAH-1869 handle incomplete startTime

* DAH-1869 fix: handle incomplete start time

* DAH-1869 fix: refactor to improve performance

* DAH-1869 add null check

* DAH-1869 fix: snapshot

Hotfix PR for prod. Related to PRs in DAH-1869

### DIFF
--- a/app/javascript/util/listingUtil.ts
+++ b/app/javascript/util/listingUtil.ts
@@ -12,6 +12,7 @@ import type {
   RailsAmiChartValue,
 } from "../api/types/rails/listings/RailsAmiChart"
 import dayjs from "dayjs"
+import customParseFormat from "dayjs/plugin/customParseFormat"
 import { RESERVED_COMMUNITY_TYPES, TENURE_TYPES, CUSTOM_LISTING_TYPES } from "../modules/constants"
 import { RailsListing } from "../modules/listings/SharedHelpers"
 import { LANGUAGE_CONFIGS, getCustomListingType, getReservedCommunityType } from "./languageUtil"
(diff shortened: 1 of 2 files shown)


app/javascript/util/listingUtil.ts
```diff
@@ -148,6 +149,26 @@ export const getEventTimeString = (listingEvent: ListingEvent) => {
...
+    const formattedHour = hour > 12 ? hour - 12 : hour
+    return `${formattedHour}:00 ${suffix}`
+  }
+  return ""
+}
+
+export const getEventDateTime = (eventDate: string, eventTime: string) => {
+  const startTime = eventTime?.includes(":") ? eventTime : `${formatEventTime(eventTime)}`
+  dayjs.extend(customParseFormat)
+  return dayjs(`${eventDate} ${startTime}`, "YYYY-MM-DD h:mmA")
```
jimlin-sfgov, line 165: This custom format looks like it can only handle `3:00PM`, not `3:00 PM`, but it does handle both for some reason.

cade-exygy: Yeah that is interesting. It looks like we use strict parsing or multiple formats to be more specific: https://day.js.org/docs/en/parse/string-format#list-of-all-available-parsing-tokens

PR #2039, opened 2023-12-19
DAH-1865 Handle Apostrophe Character Entity from Salesforce free text fields

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1865

Docs explaining the markdown-to-jsx option: https://github.com/quantizor/markdown-to-jsx?tab=readme-ov-file#optionsnamedcodestounicode

### DIFF
--- a/app/javascript/modules/listings/SharedHelpers.tsx
+++ b/app/javascript/modules/listings/SharedHelpers.tsx
@@ -1,8 +1,7 @@
 import React from "react"
-import Markdown from "markdown-to-jsx"
 import { ApplicationStatusType, StatusBarType, t } from "@bloom-housing/ui-components"
 import { areLotteryResultsShareable, getTagContent } from "../../util/listingUtil"
-import { localizedFormat } from "../../util/languageUtil"
+import { localizedFormat, renderInlineMarkup } from "../../util/languageUtil"
 import type RailsSaleListing from "../../api/types/rails/listings/RailsSaleListing"
 import type RailsRentalListing from "../../api/types/rails/listings/RailsRentalListing"
 import type { ListingEvent } from "../../api/types/rails/listings/BaseRailsListing"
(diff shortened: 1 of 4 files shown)


app/javascript/util/languageUtil.tsx
```diff
@@ -164,14 +164,18 @@ export const getSfGovUrl = (enLink: string, node?: number, path?: string) => {
  */
 export function renderMarkup(translatedString: string, allowedTags?: string) {
```
jimlin-sfgov, line 165: I grepped for the regex `renderMarkup\([^t]` and found that this function is used for free text fields in some places.

app/javascript/util/languageUtil.tsx
```diff
@@ -164,14 +164,18 @@ export const getSfGovUrl = (enLink: string, node?: number, path?: string) => {
...
   return (
-    <Markdown options={{ forceBlock: true }}>
+    <Markdown options={{ forceBlock: true, namedCodesToUnicode: { "#39": "\u0027" } }}>
       {stripMostTags(translatedString, allowedTags)}
     </Markdown>
   )
 }
 
 export function renderInlineMarkup(translatedString: string, allowedTags?: string) {
```
jimlin-sfgov, line 173: I grepped for the regex `renderInlineMarkup\([^t]` and found that this function is used for free text fields in some places.

package.json
```diff
@@ -25,7 +25,7 @@
     "cssnano": "^6.0.1",
     "dayjs": "^1.10.7",
     "dotenv": "2.0.0",
-    "markdown-to-jsx": "7.1.8",
+    "markdown-to-jsx": "7.3.2",
```
jimlin-sfgov, line 28: Updated the library because the older version had a bug related to the `namedCodesToUnicode` option.

jimlin-sfgov:

chadbrokaw (approved):
One question but LGTM

app/javascript/modules/listings/SharedHelpers.tsx
```diff
@@ -89,7 +89,9 @@ export const getEventNote = (listingEvent: ListingEvent) => {
     <div className="flex flex-col">
       {listingEvent.Venue && (
         <span className="links-space translate">
-          <Markdown>{listingEvent.Venue}</Markdown>
+          <Markdown options={{ namedCodesToUnicode: { "#39": "\u0027" } }}>
```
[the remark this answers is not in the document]
jimlin-sfgov: > Curious at to why this does not use the renderMarkup function defined below?

I think it should, I've made an update to do that.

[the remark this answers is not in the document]
jimlin-sfgov: > Should this options object, and the ones below, be extracted into some consts file so that we have better consistency?

I think we're okay for now, as long as we keep using the `renderMarkup` functions instead of `<Markdown>` directly. We may want to revisit this if we need to pass more options to `<Markdown>`.

PR #2043, opened 2023-12-20
DAH-1944 Fix console error on Application Form Demographics page related to Veterans Flag

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1944


jimlin-sfgov:

chadbrokaw (approved):

PR #2047, opened 2024-01-08
DAH-1974 Disable machine translation e2e tests

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1974


cade-exygy (approved):
Looks good!

PR #2051, opened 2024-01-11
Release 01 11 2024

alulabeshue-sfgov:


jimlin-sfgov (approved):
I've done a cursory check on main website functions by visiting all the types of pages, and also SCV listing applications, since that is in this release as well.

I've done more in depth testing on the Listings with Veteran preferences, since that will be under scrutiny for this release. Everything seems to be working well.

PR #2052, opened 2024-01-12
chore: DAH-1554 Rails 7 upgrade Part 1

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1554


config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,107 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file eases your Rails 7.0 framework defaults upgrade.
+#
+# Uncomment each configuration one by one to switch to the new default.
+# Once your application is ready to run with all new defaults, you can remove
+# this file and set the `config.load_defaults` to `7.0`.
```
jimlin-sfgov, line 7: Removal of this file will occur on the Rails 7 Upgrade Part 2 PR.

config/application.rb
```diff
@@ -7,12 +7,15 @@
 require_relative '../lib/rack_x_robots_tag'
 
 # Require the gems listed in Gemfile, including any gems
-# you've limited to :test, :development, or :production.
+# you've limited to :test, : elopment, or :production.
```
cade-exygy, line 10: nit: Was this auto generated? Just wondering why it is "elopment" vs "development"

cade-exygy (approved):
Testing criteria passes on PR env! I am curious if we should test the cache service as well on this PR env? If so, I can set that up and see if there are any issues.

config/application.rb
jimlin-sfgov: good catch, I must have accidentally deleted "dev" somehow

jimlin-sfgov:
> I am curious if we should test the cache service as well on this PR env?

@cade-exygy Yes that would be great.

PR #2058, opened 2024-01-19
feat: DAH-1999 Show Veteran Application Question only when Veteran Pref is present on Listing

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1999

### DIFF
--- a/app/assets/javascripts/shared/SharedService.js.coffee
+++ b/app/assets/javascripts/shared/SharedService.js.coffee
@@ -167,7 +167,12 @@ SharedService = ($http, $state, $window, $document) ->
       a
     ), []).join '&'
 
-  Service.showVeteransApplicationQuestion = $window.VETERANS_APPLICATION_QUESTION is 'true'
+  Service.showVeteransApplicationQuestion = (listing) ->
+    $window.VETERANS_APPLICATION_QUESTION is 'true' &&
+    !!listing.Listing_Lottery_Preferences &&
+    _.some(listing.Listing_Lottery_Preferences, (pref) ->
+      pref?.Lottery_Preference?.Name?.toLowerCase().includes("veteran")
+    )
 
   return Service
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/shared/SharedService.js.coffee
```diff
@@ -167,7 +167,12 @@ SharedService = ($http, $state, $window, $document) ->
       a
     ), []).join '&'
 
-  Service.showVeteransApplicationQuestion = $window.VETERANS_APPLICATION_QUESTION is 'true'
+  Service.showVeteransApplicationQuestion = (listing) ->
+    $window.VETERANS_APPLICATION_QUESTION is 'true' &&
+    !!listing.Listing_Lottery_Preferences &&
+    _.some(listing.Listing_Lottery_Preferences, (pref) ->
+      pref?.Lottery_Preference?.Name?.toLowerCase().includes("veteran")
+    )
```
alulabeshue-sfgov, line 175: thought: With this change we may no longer need the env variable, similar to the listing details page where we just check the preferences and not the env variable.

jimlin-sfgov: Yes, I think in the (near) future we can phase out the env var. For now, I still like that we can disable the Veteran question with the env var, in case something goes wrong.

alulabeshue-sfgov: Makes sense. We have a [ticket](https://sfgovdt.jira.com/browse/DAH-1816) to go back and clean up feature toggles like the react pages, I've added this to that list to be cleaned up.

alulabeshue-sfgov (approved):
LGTM! Validated that veteran question appears only for listings that have veteran preferences. Also validated that even when a listing has veteran preferences if `VETERANS_APPLICATION_QUESTION` is set to false then the veteran question will not appear.

PR #2064, opened 2024-01-25
fix: DAH-2012 webappid set to null during account updates

alulabeshue-sfgov:
[DAH-2012](https://sfgovdt.jira.com/browse/DAH-2012)

[DAH-2012]: https://sfgovdt.jira.com/browse/DAH-2012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/account_controller.rb
+++ b/app/controllers/api/v1/account_controller.rb
@@ -6,22 +6,14 @@ class Api::V1::AccountController < ApiController
 
   def my_applications
     applications = map_listings_to_applications(current_user_applications)
-    render json: { applications: applications }
+    render json: { applications: }
   end
 
   def update
     contact = account_params
     contact[:contactID] = current_user.salesforce_contact_id
     contact[:webAppID] = current_user.id
-    salesforce_contact = Force::AccountService.create_or_update(
-      webAppID: contact[:webAppID],
-      contactId: contact[:contactID],
-      email: contact[:email],
-      firstName: contact[:firstName],
-      middleName: contact[:middleName],
-      lastName: contact[:lastName],
-      DOB: contact[:DOB],
-    )
+    salesforce_contact = Force::AccountService.create_or_update(contact.as_json)
     Emailer.account_update(current_user).deliver_later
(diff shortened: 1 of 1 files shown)


jimlin-sfgov:
For me, I've found the email confirmation step during account creation to not work in PR apps and the Full app. Which would make testing this PR difficult.

A workaround is to search through the logs in Papertrail to find the confirmation token, and manually enter the confirmation page URL into the browser, e.g.:
`https://dahlia-webapp-pr-2064.herokuapp.com/api/v1/auth/confirmation?config=default&confirmation_token=<token_found_in_the_logs>`

Screenshot 2024-01-26 at 3 39 27 PM

app/controllers/api/v1/account_controller.rb
```diff
@@ -6,22 +6,14 @@ class Api::V1::AccountController < ApiController
...
-    salesforce_contact = Force::AccountService.create_or_update(
-      webAppID: contact[:id],
-      contactId: contact[:contactID],
-      email: contact[:email],
-      firstName: contact[:firstName],
-      middleName: contact[:middleName],
-      lastName: contact[:lastName],
-      DOB: contact[:DOB],
-    )
+    salesforce_contact = Force::AccountService.create_or_update(contact.as_json)
```
cade-exygy, line 16: This looks good! Just a question: Is there any data from contact that we would not want to pass to the AccountService? I cannot think of any, and I don't think there would be security/privacy concerns since the call to AccountService should be internal.

cade-exygy (approved):
Testing criteria passes. Created an account and updated, webAppId=2 persisted.

alulabeshue-sfgov:
> For me, I've found the email confirmation step during account creation to not work in PR apps and the Full app. Which would make testing this PR difficult.

I created a follow up [ticket](https://sfgovdt.jira.com/browse/DAH-2022) to look into this.

PR #2025, opened 2023-12-06
DAH-1807 fix: remove logs, listing_image_service, refactor

cade-exygy:
DAH-1807 

Refactor Listing Image Caching to no longer use ListingImageService (and only use MultipleListingImageService).

PR #2069, opened 2024-01-30
DAH-1812 fix: cache if nothing exists in the cache

tallulahkay:
DAH-1812

Caches if there is nothing already in the cache. I'd like to include this in the release but if folks think we should wait until the next release we can. A workaround would be forcing a recache on prod after the release.

See comment on ticket for more context.

### DIFF
--- a/app/services/cache_service.rb
+++ b/app/services/cache_service.rb
@@ -6,12 +6,12 @@ class CacheService
     Rails.logger.info('CacheService Started')
     # Marshal.load(Marshal.dump(cached_listings)) is a hack to make a deep clone of cached_listings
     # so that prev_cached_listings isn't pointing to the same object as fresh_listings
-    cached_listings = Force::ListingService.listings(subset: 'browse')
+    cached_listings = Force::ListingService.listings(subset: 'browse', cache_only: true)
     @prev_cached_listings = Marshal.load(Marshal.dump(cached_listings))
 
     @fresh_listings = Force::ListingService.listings(subset: 'browse', force: true)
 
-    if opts[:refresh_all]
+    if opts[:refresh_all] || @prev_cached_listings.nil?
       cache_all_listings
     else
       cache_only_updated_listings
--- a/app/services/force/listing_service.rb
+++ b/app/services/force/listing_service.rb
@@ -12,7 +12,8 @@ module Force
       params[:type] = attrs[:type] if attrs[:type].present?
       params[:ids] = attrs[:ids] if attrs[:ids].present?
       force = attrs[:force].present? ? attrs[:force] : false
-      get_listings(params:, force_recache: force)
+      cache_only = attrs[:cache_only].present? ? attrs[:cache_only] : false
+      get_listings(params:, force_recache: force, cache_only:)
     end
 
     # get listings with eligibility matches applied
@@ -108,12 +109,15 @@ module Force
       end
     end
 
-    private_class_method def self.get_listings(params: {}, force_recache: false)
+    private_class_method def self.get_listings(params: {}, force_recache: false, cache_only: false)
       params[:subset] ||= 'browse'
       Rails.logger.info("Calling self.get_listings with force: #{force_recache}")
 
       results = Request.new(parse_response: true)
-                       .cached_get('/ListingDetails', params, force_recache)
+                       .cached_get('/ListingDetails', params, force_recache, cache_only)
+
+      return if results.nil?
+
       results_with_cached_listing_images = add_cloudfront_urls_for_listing_images(results)
       add_image_urls(results_with_cached_listing_images)
     end
--- a/app/services/force/request.rb
+++ b/app/services/force/request.rb
@@ -24,7 +24,7 @@ module Force
       variable.to_s.casecmp('true').zero?
     end
 
-    def cached_get(endpoint, params = nil, force = false)
+    def cached_get(endpoint, params = nil, force = false, cache_only = false)
       key = "#{endpoint}#{params ? '?' + params.to_query : ''}"
       force = ActiveModel::Type::Boolean.new.cast(force)
 
@@ -38,6 +38,8 @@ module Force
         "running cached_get for #{endpoint} with force set to #{force_refresh}",
       )
       @cache.fetch(key, force: force_refresh, expires_in: expires_in) do
+        next if cache_only
+
         get(endpoint, params)
       end
     end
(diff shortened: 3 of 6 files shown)


jimlin-sfgov:
I'm getting deja vu from reading the review instructions in Jira. Are they the same instructions as the ones that were used for [this PR](https://github.com/SFDigitalServices/sf-dahlia-web/pull/2025)?

app/services/force/request.rb
```diff
@@ -38,6 +38,8 @@ def cached_get(endpoint, params = nil, force = false)
         "running cached_get for #{endpoint} with force set to #{force_refresh}",
       )
       @cache.fetch(key, force: force_refresh, expires_in: expires_in) do
```
jimlin-sfgov, line 40: question: I'm not following how the code changes in this PR will solve this issue:
> If there isn’t already a cache (i.e. the cache is expired or it’s a new deploy) then the caching doesn’t work

Can't we just pass in `true` for the `force:` option here? It's what triggers a cache write, [docs here](https://api.rubyonrails.org/v7.0.0/classes/ActiveSupport/Cache/Store.html#method-i-fetch)

jimlin-sfgov (commented):
issue: It's not clear to me how the changes in this PR fix the stated problem. Some comments on how the code works would be helpful, or we could do a PR review pairing session.

app/services/cache_service.rb
```diff
@@ -6,12 +6,12 @@ def prefetch_listings(opts = {})
...
     # Marshal.load(Marshal.dump(cached_listings)) is a hack to make a deep clone of cached_listings
     # so that prev_cached_listings isn't pointing to the same object as fresh_listings
-    cached_listings = Force::ListingService.listings(subset: 'browse')
+    cached_listings = Force::ListingService.listings(subset: 'browse', cache_only: true)
     @prev_cached_listings = Marshal.load(Marshal.dump(cached_listings))
 
     @fresh_listings = Force::ListingService.listings(subset: 'browse', force: true)
 
-    if opts[:refresh_all]
+    if opts[:refresh_all] || @prev_cached_listings.nil?
```
jimlin-sfgov, line 14: Question: I missed this line when I reviewed the first time. This is the code that will addresses the main issue right? I'm still not clear on what the new `cache_only` argument does though.

app/services/force/request.rb
tallulahkay: Ok interesting point about how force triggers a cache write, but I think the caching we are missing is in cache_single_listing, which is separate

I named the param cache_only to indicate "grab only what's in the cache" so we can check if we need to move forward and do the cache_single_listing calls

tallulahkay:
thanks for your review jim! so yes, review instructions are the same, I just found a bug when doing them for the release review, which this addresses.

I think we need cache_only because we need to specifically compare what's in the cache to what's new. Previously, if there was nothing in the cache, we grabbed what's it's salesforce instead of just returning nil and compared what's in salesforce to what's in salesforce (causing no caching)

The way we have caching set up needs an initial state, like a base case

app/services/cache_service.rb
tallulahkay: @prev_cached_listings won't return as nil without the cache_only functionality

app/services/force/request.rb
tallulahkay: I am still pondering on this because I can't figure out why the built in caching doesn't do what we need it to do - will get back to you after some investigation

app/services/force/listing_service.rb
```diff
@@ -12,7 +12,8 @@ def self.listings(attrs = {})
       params[:type] = attrs[:type] if attrs[:type].present?
       params[:ids] = attrs[:ids] if attrs[:ids].present?
       force = attrs[:force].present? ? attrs[:force] : false
-      get_listings(params:, force_recache: force)
+      cache_only = attrs[:cache_only].present? ? attrs[:cache_only] : false
+      get_listings(params:, force_recache: force, cache_only:)
```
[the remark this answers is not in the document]
tallulahkay: I like this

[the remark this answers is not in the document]
tallulahkay: I'm implementing this change just have to test it for good measure

jimlin-sfgov:
@tallulahkay 

preface: this suggestion would be a pretty big change, very open to discussing this and/or not implementing it.

suggestion: I prefer not to add arguments to multiple methods just to pass a flag along the execution chain. I think a more readable way to fix the bug is to create a method that explicitly checks if the cache is empty or not. For example:

```ruby
# app/services/cache_service.rb
def prefetch_listings(opts = {})
  # ...
  if opts[:refresh_all] || Force::ListingService.listings_cache_is_empty?
  # ...
end

# app/services/force/listing_service.rb
def self.listings_cache_is_empty?
  params = { subset: 'browse' }
  key = "/ListingDetails?#{params.to_query}"
  Rails.cache.exists?(key)
end


```

jimlin-sfgov:
thought: I'm finding the bug description at odds with how Rails caching is supposed to work.

> if nothing is found in the cache, ___it just calls the salesforce api___

However, the documentation for `Rails.cache.fetch` clearly states that data will be written to the cache in the event of a cache miss:

> if a block has been passed, that block will be passed the key and executed in the event of a cache miss. The return value of the block ___will be written to the cache___ under the given cache key

p.s. Sorry for posting all these comments that are spaced apart, it's taking a while for me to re-familiarize myself with our caching system.

tallulahkay:
Re: the code update - I like your suggestion - I say we make a new ticket for this and move on with the release without it and implement what you suggested, but since it's a bigger change we should take it through the whole process instead of tacking it onto this release. For this release, we can just manually refresh the Rails cache in prod.

Re: The second comment.. Saying "it just calls the salesforce api" was unspecific, sorry. The rails cache.fetch calls the salesforce api and then caches the information at that call (you may notice the listings page update), but it doesn't move on through our cache_service (so the individual listing doesn't get a cache value), and so it doesn't cache everything else we need, which includes caching the individual listing in the rails cache along with other information surrounding the listings (like the images). It gets stuck in this loop of comparing what is returned from the salesforce api to itself, caching (rails) only the listing_s_ and not the individual listing. Tldr - using what is returned from the salesforce api as a means for comparison in cache_service is insufficient.

cade-exygy (commented):
Code changes make sense! Can we find some time to quickly pair to go over it just so I am sure I understand completely!

PR #2072, opened 2024-02-01
feat: DAH-2010: Upgrade to Typescript 5 and fortify types

chadbrokaw:
[DAH-2010](https://sfgovdt.jira.com/browse/DAH-2010)

[DAH-2010]: https://sfgovdt.jira.com/browse/DAH-2010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gemfile
```diff
@@ -23,7 +23,7 @@ gem "sprockets-rails"
 gem 'sprockets_uglifier_with_source_maps'
 # See https://github.com/rails/execjs#readme for more supported runtimes
 # gem 'therubyracer',  platforms: :ruby
-gem 'puma', '~> 5.6.8'
+gem 'puma', '~> 6.4.2'
```
jimlin-sfgov, line 26: question: Is this intentional? Seems unusual to fold a major version ruby gem change into a Node package update.

app/javascript/api/types/rails/listings/RailsUnit.d.ts
```diff
@@ -21,10 +21,10 @@ type RailsUnit = {
   Unit_Square_Footage?: number
   attributes: ListingAttributes
   isReservedCommunity: boolean
-  Price_Without_Parking?: any
-  Price_With_Parking?: any
-  HOA_Dues_Without_Parking?: any
-  HOA_Dues_With_Parking?: any
+  Price_Without_Parking?: number | string // Eventually these get transformed from numbers into price range strings, to be used in the pricing table
```
jimlin-sfgov, line 24: thought: this seems like a code smell indicating that we're going against the data immutability principle of React.

app/javascript/hooks/useScript.tsx
```diff
@@ -35,6 +36,7 @@ const useScript = (src) => {
         script.addEventListener("error", setAttributeFromEvent)
       } else {
         // Grab existing script status from attribute and set to state.
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
jimlin-sfgov, line 39: suggestion: This entire file is written in a style that predates Typescript, and is used only to facilitate our hacky machine translation process. We can just ignore the entire file by adding [// @ts-nocheck](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#-ts-nocheck-in-typescript-files) to the top of the file/

app/javascript/modules/listingDetails/ListingDetailsUnitAccordions.tsx
```diff
@@ -45,7 +45,18 @@ const getTableData = (units: RailsUnit[]) => {
...
+  availability: number
+  minSqFt: number
+  maxSqFt: number
+}
+
+type PropertyType = {
+  [key: string]: UnitType
+}
+
+const sortUnits = (units: RailsUnit[]): PropertyType => {
```
jimlin-sfgov, line 59: suggestion: We can use the [record utility](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) to save a few lines and remove the need for the `PropertType` type, e.g.
```js
const sortUnits = (units: RailsUnit[]): Record<RailsUnit["Unit_Type"], UnitType> => {
```

app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -226,8 +227,12 @@ export const getListingCards = (listings, directoryType, stackedDataFxn, hasFilt
     )
   })
 
-export const openListingsView = (listings, directoryType, stackedDataFxn, filtersSet?) =>
-  listings.length > 0 && getListingCards(listings, directoryType, stackedDataFxn, filtersSet)
+export const openListingsView = (
+  listings: RailsListing[],
+  directoryType: DirectoryType,
+  stackedDataFxn: (listing: RailsRentalListing) => Record<string, StackedTableRow>[],
```
jimlin-sfgov, line 233: question: I'm seeing `stackedDataFxn` appear a lot in a other places as an untyped argument. Is it worth the effort to define this type in a more accessible place so that we can use it in more places?

Gemfile
chadbrokaw: Thank you for calling that out! This snuck in because I was getting a bunch of bundler issues with 5.6.8, upgrading to 6.4.2 solved those (and brought us in line with Partners). I extracted this into its own PR (#2073) so the separation is clearer

app/javascript/util/routeUtil.ts
```diff
@@ -45,7 +45,12 @@ export const getNewLanguagePath = (
   currentPath: string | undefined,
   newLanguagePrefix: string,
   queryString: string | undefined
-): string => getLocalizedPath(currentPath, toLanguagePrefix(newLanguagePrefix), queryString)
+): string => {
+  if (Object.values(LanguagePrefix).includes(newLanguagePrefix as LanguagePrefix)) {
+    return getLocalizedPath(currentPath, newLanguagePrefix as LanguagePrefix, queryString)
+  }
+  return getLocalizedPath(currentPath, toLanguagePrefix(undefined), queryString)
```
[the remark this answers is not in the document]
chadbrokaw: I'm not entirely sure what you mean here? Do you mean changing line 52 to be:
```suggestion
  return getLocalizedPath(currentPath, LanguagePrefix.English, queryString)
```

app/javascript/api/types/rails/listings/RailsUnit.d.ts
chadbrokaw: Sigh, I agree... however I prefer this over `any`. With that said, this should be refactored in another ticket to remove the looseness of this type, unless you see an easy way to do it?

app/javascript/hooks/useScript.tsx
chadbrokaw: I like this idea but we would also have to attach an eslint ignore for the ts comment itself. What if we just applied eslint rules to the file as needed? So the top of the file would have:
```
/* eslint-disable @typescript-eslint/no-explicit-any */
/* eslint-disable @typescript-eslint/no-unsafe-argument */
```

As an aside, this file could use a thorough refactor to be better in line with Typescript principles.

app/javascript/modules/listingDetails/ListingDetailsUnitAccordions.tsx
chadbrokaw: 😍 Amazing!

app/javascript/__tests__/modules/listingDetails/ListingDetailsUnitAccordions.test.tsx
```diff
@@ -31,8 +31,8 @@ describe("ListingDetailsUnitAccordion", () => {
           fetchedUnits: true,
           fetchingAmiCharts: false,
           fetchedAmiCharts: false,
-          fetchingAmiChartsError: null,
-          fetchingUnitsError: null,
+          fetchingAmiChartsError: undefined,
+          fetchingUnitsError: undefined,
```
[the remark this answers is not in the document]
chadbrokaw: I'm not 100% sure, but I think what is happening is that the useReducer is not provided the ContextProps type, so the type checker is being more lenient with the starting state. 

I will look into this more.

app/javascript/api/types/rails/listings/RailsUnit.d.ts
jimlin-sfgov: I don't see an easy way, otherwise I would have offered a suggestion. I can create a ticket to look into the complexities of certain functions like `matchSharedUnitFields`.

app/javascript/hooks/useScript.tsx
jimlin-sfgov: > What if we just applied eslint rules to the file as needed?

Sounds good!

> this file could use a thorough refactor

Might not be worth the time. This file is only used for our current machine translation process, which is on track to be replaced.

app/javascript/api/types/rails/listings/RailsUnit.d.ts
chadbrokaw: Okay great, thank you!

app/javascript/hooks/useScript.tsx
chadbrokaw: > Might not be worth the time. This file is only used for our current machine translation process, which is on track to be replaced.

Good point! 

Okay, I will apply the file wide eslint rules. Thanks for the suggestion!

jimlin-sfgov: fyi this eslint comment can be removed, I see that you removed the other ones already.

jimlin-sfgov (approved):
great job! this looks good to merge when once we merge https://github.com/SFDigitalServices/sf-dahlia-web/pull/2073

PR #2075, opened 2024-02-06
fix: DAH-2052 Remove log message about missing index method

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2052


chadbrokaw (approved):
Thank you for taking care of this! 🎉

alulabeshue-sfgov:
It also looks like this was [fixed](https://github.com/newrelic/newrelic-ruby-agent/issues/2048) in the newrelic ruby agent. We are currently using 9.0.0 according to the papertrail logs, if we update to 9.3.0 we may not need this empty function.

jimlin-sfgov:
> if we update to 9.3.0 we may not need this empty function.

Good find, I'll create another infra ticket to specify the gem version for `newrelic_rpm` so we can update it to the latest version.

jimlin-sfgov:
Closing this PR since the issue is not urgent, and the suggestion to update the newrelic gem instead is a better approach.

PR #2079, opened 2024-02-07
Fix: add google signing key step for circleci

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2089


tallulahkay (approved):
This lgtm! No idea why this became an issue, but this fix can't hurt and we should merge to unblock the pipeline

PR #2081, opened 2024-02-12
fix: DAH-2068: Stabilize ListingDetailsEligibility Test

chadbrokaw:
[DAH-2068](https://sfgovdt.jira.com/browse/DAH-2068)

[DAH-2068]: https://sfgovdt.jira.com/browse/DAH-2068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tallulahkay:
This doesn't fix the flakiness for me - for me, it has to do with my env variable setup. I think we should be setting environment variables in the test setup instead of using local vars.

tallulahkay (approved):
Ignore my last comment - this works! I must have had some error locally

PR #2082, opened 2024-02-12
DAH-1947 prevent duplicate db entries

tallulahkay:
DAH-1947

The ticket says to delete stale db records. I updated the code so that, moving forward, we query the listing images by listing id and raw image url, and update the image urls if they have changed. This way, we can have minimal db entries (hopefully just 1 per image per listing) and they will stay up to date. 

I also removed the populating of image url for the whole listing because we don't really use it anymore. (We now use displayImageUrl).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,6 +194,8 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: ./reports/junit/
       - store_test_results:
           path: ./reports/junit/
+      - store_artifacts:
+          path: app/javascript/test-coverage
   unit-rspec:
     <<: *defaults
     steps:
@@ -204,6 +206,8 @@ jobs:
       - run: RAILS_ENV=test bundle exec rails db:create
       - run: RAILS_ENV=test bundle exec rails db:test:prepare
       - run: bundle exec rake spec
+      - store_artifacts:
+          path: coverage
   unit-angular:
     <<: *defaults
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ yarn-debug.log*
 .DS_Store
 
 # Test output
+coverage
 app/javascript/test-coverage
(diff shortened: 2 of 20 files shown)


alulabeshue-sfgov (changes requested):
issue: It looks like the imageUrl in used in the [angular](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/assets/javascripts/short-form/templates/g1-confirmation.html.slim#L8) application during /confirmation

Screenshot 2024-02-13 at 4 12 31 PM

tallulahkay:
@alulabeshue-sfgov I removed the image url stuff because it wasn't required for the ticket - was just trying to do some related cleanup but I can make another ticket to remove the use of image url

cade-exygy (approved):
Changes look good! Going to manually run the jobs on the PR env now

cade-exygy:
Testing on the PR env: It looks like some images are being picked up and cached correctly, but I noticed that the images for this listing are not getting cached: https://dahlia-webapp-pr-2082.herokuapp.com/listings/a0W8H0000014LS9UAM

tallulahkay:
> Testing on the PR env: It looks like some images are being picked up and cached correctly, but I noticed that the images for this listing are not getting cached: https://dahlia-webapp-pr-2082.herokuapp.com/listings/a0W8H0000014LS9UAM

@cade-exygy I think the rails cache is never being updated really because it wasn't initialized. We rely on the prefetch daily call to initialize the cache - this was my mistake I should have prepared the environment properly before anyone looked at it

tallulahkay:
@cade-exygy which is what I was trying to solve here: https://github.com/SFDigitalServices/sf-dahlia-web/pull/2069... it's not really a problem in full and prod because of prefetch_daily (only noticeable after release and before prefetch_daily job is run)

cade-exygy:
Looks like this is caching images correctly!

PR #2089, opened 2024-02-20
feat: remove force update flag

cade-exygy:
https://sfgovdt.jira.com/browse/DAH-1997?atlOrigin=eyJpIjoiODM1MTMxODBjZjNhNDhmN2I1ZGQ5YTllZjJkYmI1M2YiLCJwIjoiaiJ9

Removed the FORCE_MULTIPLE_LISTING_IMAGE_UPDATE flag which overrides the check determining if we write to the Postgres table.

Testing steps:
Follow typical listing image flow
Ensure the rake job is set in the scheduler and the proper env variables are set in heroku
Update a listing image and wait for the job to run
Ensure that the image is cached and the database table is updated correctly.


tallulahkay (approved):
great!

PR #2096, opened 2024-02-28
DAH-2122 cypress e2e fixtures

tallulahkay:
DAH-2122 

Implements fixtures for cypress e2e tests so that they're no longer hitting salesforce against the prod. (In some cases, they will always use fixtures, like in rentalListings.e2e.ts where it's specifically testing the words appearing)


alulabeshue-sfgov:
Looks good! Would we still have a way to run full integration tests when needed? In our last meeting we also discussed having something like contract tests against salesforce api in replace of full integration tests.

cypress/plugins/index.js
```diff
@@ -18,4 +18,11 @@
 module.exports = (_on, _config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+  _on("task", {
+    log(message) {
+      console.log(message)
+
+      return null
+    },
+  })
```
jimlin-sfgov, line 27: question: is this left over from debugging?

cypress/fixtures/listingDetails.json
```diff
@@ -0,0 +1,239 @@
...
+		],
+		"attributes": {
+			"type": "Listing",
+			"url": "/services/data/v35.0/sobjects/Listing/a0W8H000000AmpKUAS"
+		},
+		"Name": "Test - Closed with Lottery Results (Do Not Modify)",
+		"Realtor_Commission_Unit": "percent",
+		"Allows_Realtor_Commission": false,
+		"Parking_Information": "Onsite parking",
+		"Tenure": "New rental",
```
jimlin-sfgov, line 55: suggestion (non-blocking): update the name of this file to indicate that it has details for a _rental_ listing.

tallulahkay: for sure! I also included that it is a completed listing

tallulahkay:
@alulabeshue-sfgov yes!! I updated it so that it only uses fixtures if it's in a production environment when it's beneficial to hit full for testing

cypress/e2e/listingDetails/completedLotteryListing.e2e.ts
```diff
@@ -51,7 +37,18 @@ describe("Listing Details for Completed Lottery Listing", () => {
   //   // there is a rogue loading issue beyond the scope of this story
   //   cy.wait(6000)
   // })
-
+  beforeEach(() => {
+    if (Cypress.env("salesforceInstanceUrl") === "https://sfhousing.my.salesforce.com") {
```
jimlin-sfgov, line 41: question: Is this conditional set up as intended? Equality means that we are running with prod env vars, so it means that the fixtures will be used when running the production tests.

cypress/e2e/listingDetails/completedLotteryListing.e2e.ts
```diff
@@ -1,13 +1,5 @@
...
-    id: "a0W8H000000AmpKUAS",
-    lotteryNumber: "01150047",
-  },
-  COMPLETED_LOTTER_PROD: {
-    id: "a0W4U00000IhGZcUAN",
-    lotteryNumber: "01171676",
-  },
-}
+const listingId = "a0W8H000000AmpKUAS"
+const lotteryNumber = "01150047"
```
jimlin-sfgov, line 2: question: Are these the intended values? They are the full environment values, and probably won't work when running prod tests.

cypress/e2e/listingDetails/completedLotteryListing.e2e.ts
tallulahkay: yes! Because we don't want to be hitting prod salesforce anymore, we want to use fixtures for prod

cypress/e2e/listingDetails/completedLotteryListing.e2e.ts
tallulahkay: Prod tests will use fixtures

jimlin-sfgov (approved):
great job!

PR #2103, opened 2024-03-08
chore: DAH-2132 update vcr recordings for rspec tests

tallulahkay:
DAH-2132

Updates VCR snapshots so they align with current endpoints.


chadbrokaw (approved):
I wonder if we should create a recurring task of some kind so that we are regularly doing this? Either way, looks great, thank you!

PR #2110, opened 2024-03-22
fix: DAH-2199 Custom Preferences whitelist

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2199


alulabeshue-sfgov (approved):
Looks good

PR #2112, opened 2024-03-25
fix: DAH-2036 Application bar accessibility 

cliu02:
[DAH-2036](https://sfgovdt.jira.com/browse/DAH-2036)

[DAH-2036]: https://sfgovdt.jira.com/browse/DAH-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/listings/ListingConstantsService.js.coffee
+++ b/app/assets/javascripts/listings/ListingConstantsService.js.coffee
@@ -123,6 +123,11 @@ ListingConstantsService = () ->
     'a0W4U00000Nlb5lUAB', # 400 China Basin
   ]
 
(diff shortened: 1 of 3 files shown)


jimlin-sfgov (changes requested):
issue: I can no longer use the tab key to select nav bar items that __are__ clickable. You can see this behavior when you complete at least one section in the application, which makes that section's nav item clickable.

jimlin-sfgov (approved):
looks good!

PR #2118, opened 2024-03-27
test: DAH-2175 Add tests for ListingDetail component

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2175


tallulahkay (approved):
nice! 100% is great

PR #2123, opened 2024-03-28
fix: DAH-1298 allow searching without leading 0

tallulahkay:
DAH-1298


jimlin-sfgov (approved):
looks good!

app/javascript/modules/listingDetailsLottery/ListingDetailsLotterySearchForm.tsx
```diff
@@ -36,7 +36,12 @@ export const ListingDetailsLotterySearchForm = ({
   )
 
   const onSubmit = (data: { lotterySearchNumber: string }) => {
-    const lotterySearchNumber = data.lotterySearchNumber
+    let lotterySearchNumber = data.lotterySearchNumber
+
+    if (lotterySearchNumber.length < 8) {
+      lotterySearchNumber = "0".repeat(8 - lotterySearchNumber.length) + lotterySearchNumber
```
fwextensions, line 42: [caniuse](https://caniuse.com/mdn-javascript_builtins_string_repeat) is showing only the most recent Chrome on Android supporting `repeat()`, which seems odd.  Also Opera Mobile.  So could do the tried and true `("00000000" + lotterySearchNumber).slice(-8)` to pad it.

cade-exygy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#browser_compatibility

Shows a different result (all compatible) but we should test to ensure

PR #2128, opened 2024-04-02
fix: DAH-2221 Use lottery_status fallback for Listing Directory Page

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2221


app/javascript/modules/listings/DirectoryHelpers.tsx
```diff
@@ -339,7 +339,7 @@ export const sortListings = (
         additional.push(listing)
       }
     } else {
-      if (areLotteryResultsShareable(listing)) {
+      if (isLotteryCompleteDeprecated(listing)) {
```
jimlin-sfgov, line 342: We swant to use `isLotteryCompleteDeprecated` but this function handles both the new Salesforce field as well as the deprecated one. Once we are sure that Salesforce has fully transitioned to the new field, we can replace all instances of `isLotteryCompleteDeprecated` with `isLotteryComplete`, which will be tracked in this ticket: https://sfgovdt.jira.com/browse/DAH-1802

tallulahkay (approved):
This looks good! fwiw, I think the `Publish_Lottery_Results` field is safe to be fully removed now

PR #2131, opened 2024-04-08
Hotfix: DAH-2229 Application Submission Logging

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2229


tallulahkay (approved):
I see the log when I apply! 

Also also - I am looking into why the legacy e2e tests don't pass, and believe I've found the reason, so I think this is good to merge regardless of those tests

PR #2142, opened 2024-04-16
Angular/DAH-2140 sf cleanup job after test run

tallulahkay:
DAH-2141

Used a workaround from this circle discussion thread: https://discuss.circleci.com/t/workaround-run-jobs-sequentially-regardless-of-the-outcome-of-the-required-jobs/40807


.circleci/config.yml
```diff
@@ -241,6 +241,22 @@ jobs:
...
+    steps:
+      - setup
+      - start-server
+      - run: bundle exec rails tests:e2e_cleanup
+  waiter:
+    docker: 
+      - image: cimg/node:current
+    steps:
+      - run: |
+          while [[ $(curl --location --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CCI_TOKEN"| jq -r '.items[]|select(.name != "waiter")|.status' | grep -c "running") -gt 0 ]]
```
jimlin-sfgov, line 255: question: How did you set `$CCI_TOKEN` in CircleCI? I though this would be a per-user token, but it seems to work for me even though I didn't set one.

jimlin-sfgov (commented):
question: Does this PR automate data clean up when running the E2E tests locally?

.circleci/config.yml
tallulahkay: I put a token I generated in my account in the environment variables - this reminds me that we will need to put one in the prod contexts - adding to the engineering notes

tallulahkay:
> question: Does this PR automate data clean up when running the E2E tests locally?

Angular tests never run locally (they don't work) so it doesn't super apply here, but it's a relevant question for the future and for the partners app

So two answers: 
1. Salesforce will have a cron job to run the cleanup job, so any data will get cleaned up at least once a day
2. It's written in the partners ticket to include running the test locally - it will probably be tacked onto the command to run the job, and if the job gets interrupted then the cleanup rake job can be run manually

jimlin-sfgov (approved):
copied this branch to test it in a separate CircleCI pipeline, works as described 👍

PR #2145, opened 2024-04-17
fix: DAH-2267 Fix the Fix for 'Reset and Start Fresh' option

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2267


tallulahkay (approved):
👍

PR #2146, opened 2024-04-17
DAH-2261 update new design for when a lottery number is entered

tallulahkay:
DAH-2261

Note: pushed new translation strings to phrase


app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.scss
```diff
@@ -1,4 +1,5 @@
 .lottery-results-row {
+  align-items: center;
```
jimlin-sfgov, line 2: question: did you confirm this styling change with designers? I can't find any mention of it.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
```diff
@@ -23,13 +23,20 @@ export const ListingDetailsLotteryResultsRow = ({ bucket }: LotteryResultsRowPro
           {bucket.preferenceName === "generalLottery"
             ? t("lottery.generalPool")
             : defaultIfNotTranslated(
-                `listings.lotteryPreference.${bucket.preferenceName}.title`,
-                bucket.preferenceName
+                `listings.lotteryPreference.${bucket.preferenceName
+                  .replace("Tier 1 ", "")
+                  .replace("Tier 2 ", "")}.title`,
+                bucket.preferenceName.replace("Tier 1 ", "").replace("Tier 2 ", "")
```
jimlin-sfgov, line 29: issue: dynamic translation keys need comments per our [guidelines](https://miro.com/app/board/uXjVMzMekug=/?moveToWidget=3458764569463367056&cot=14).

Not written down anywhere but: we have previously discussed how dynamic keys can make debugging etc. harder, and try to avoid it when possible.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
```diff
@@ -23,13 +23,20 @@ export const ListingDetailsLotteryResultsRow = ({ bucket }: LotteryResultsRowPro
...
         </Heading>
+        {bucket.preferenceName.includes("Tier 1")
+          ? t("listings.lotteryPreference.sfusd.educators.shortTitle")
+          : (bucket.preferenceName.includes("Tier 2")
+          ? t("listings.lotteryPreference.sfusd.allOtherEmployees.shortTitle")
+          : t("listings.lotteryPreference.sfusd.generalPublic.shortTitle"))}
         {bucket.preferenceName !== "generalLottery" && (
           <p className="mb-1 text-gray-950 text-sm">
-            {t("lottery.upToXUnitsAvailable", { units: bucket.unitsAvailable })}
+            {t("lottery.upToXUnitsAvailable", { units: bucket.unitsAvailable ?? 0 })}
```
jimlin-sfgov, line 39: question: Have we discussed if this is the best way to handle blank values? I could be wrong, but blank values shouldn't happen in Production and usually indicate a data issue in Salesforce.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
```diff
@@ -23,13 +23,20 @@ export const ListingDetailsLotteryResultsRow = ({ bucket }: LotteryResultsRowPro
...
             : defaultIfNotTranslated(
-                `listings.lotteryPreference.${bucket.preferenceName}.title`,
-                bucket.preferenceName
+                `listings.lotteryPreference.${bucket.preferenceName
+                  .replace("Tier 1 ", "")
+                  .replace("Tier 2 ", "")}.title`,
+                bucket.preferenceName.replace("Tier 1 ", "").replace("Tier 2 ", "")
               )}
         </Heading>
+        {bucket.preferenceName.includes("Tier 1")
```
jimlin-sfgov, line 32: question: Are we sure this condition is sufficient? There won't new preferences in the near future that also have `"Tier 1"` in their string name?

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.scss
tallulahkay: I did not! Without this, now the the block of text is bigger, it will look like: 
Screenshot 2024-04-19 at 9 40 32 AM
which I assumed is something Yindi would flag - I'll comment on the ticket to ask

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
tallulahkay: I was approaching this as a stopgap design, anticipating a much bigger redesign later before any new preferences get added, but maybe that's a risky assumption

Maybe I can branch off the way I did with the pr for before the lottery number is entered, and make a component that will only be used for educators?

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
tallulahkay: We have not! I agree - won't happen in production

Is that a design question ie "If there are blank values, what should it look like"? Or we can just remove this change to line 39 and continue assuming there won't be blank values

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
tallulahkay: Ah, thanks I'll add a comment

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.scss
tallulahkay: yindi said top aligned is fine, so I'll remove the scss

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
jimlin-sfgov: > maybe would be better to have an educator specific ListingDetailsLotteryResultsRow

That's my personal preference (minimizing inline logic), but I'll leave that decision up to you.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
jimlin-sfgov: The current implementation is actually better for detecting data issues, even if it is ugly. We've caught things in the past because `up to %{units}` is really easy to spot.

app/javascript/modules/listingDetailsLottery/ListingDetailsLotteryResultsRow.tsx
jimlin-sfgov: I haven't looked at the other PR, but componentization is always good!

tallulahkay:
@jimlin-sfgov I did some componentization so the new changes only apply to educator listings

jimlin-sfgov (approved):
great job!

jimlin-sfgov:
@tallulahkay This is not directly related to this PR but: one of your other merged PRs added`cypress/downloads/downloads.html`

PR #2147, opened 2024-04-18
DAH-2272 run angular jobs for all non-prod jobs

tallulahkay:
DAH-2272


jimlin-sfgov (approved):
lgtm! I think the testing of the main branch will have to happen after the merge, because these config changes won't be in effect until then?

PR #2155, opened 2024-05-02
chore: DAH-2296 refactor and test leasing agent support

alulabeshue-sfgov:
[DAH-2296](https://sfgovdt.jira.com/browse/DAH-2296)

[DAH-2296]: https://sfgovdt.jira.com/browse/DAH-2296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/api/v1/listing_interest_controller.rb
```diff
@@ -24,17 +25,17 @@ def confirmation
       end
 
       listing_id = listing_id_map(b)
```
tallulahkay, line 27: could you add comments here explaining what r, x, and e mean?

config/routes.rb
```diff
@@ -37,7 +37,7 @@ def matches?(_)
           get 'eligibility' => 'listings#eligibility'
         end
       end
-      get 'trk' => 'email#confirmation'
+      get 'trk' => 'listing_interest#index'
```
tallulahkay, line 40: I actually like confirmation better - I think it's clearer

config/routes.rb
```diff
@@ -37,7 +37,7 @@ def matches?(_)
           get 'eligibility' => 'listings#eligibility'
         end
       end
-      get 'trk' => 'email#confirmation'
```
tallulahkay, line 40: what is trk?

app/controllers/api/v1/listing_interest_controller.rb
```diff
@@ -24,17 +25,17 @@ def confirmation
...
-      redirect_to "/confirming_email?listing=#{listing_id}&response=#{r}"
+      redirect_to "/listing_interest?listing=#{listing_id}&response=#{r}"
```
tallulahkay, line 28: can you also explain b, a, r, s?

tallulahkay (commented):
This looks good - just wondering if there's a way to avoid hard coding the listings? can we generate a listingIdToNameMap by calling an endpoint?

config/routes.rb
alulabeshue-sfgov: It is short for tracking. Email templates in salesforce campaign monitor have a size limit for links and the token itself was getting long so we decided to use shorter names in the api like `trk` for tracking and the param `t` for token.

config/routes.rb
alulabeshue-sfgov: Ok. I was thinking it might be confused with the email confirmations we have today for when a user signs up for an account.

tallulahkay: listing_interest#confirmation?

tallulahkay (approved):
Love the comments!

PR #2159, opened 2024-05-07
DAH-2078 Link Preview with React Helmet

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2078


tallulahkay:
@jimlin-sfgov have you tested in full? I'm not convinced this is going to change any behavior - when I was testing it, a lot of preview sites saw the tags and showed them, but then it didn't actually work in practice

jimlin-sfgov:
@tallulahkay You're right, link previews are inconsistent across tools and platforms. Additionally, the prerender token is getting removed from the review app. I'll test some more and update the instructions

jimlin-sfgov:
Closing this PR, there is a bug where link previews don't work on X. We'll have to address that bug before attempting to implement this PR's ticket.

PR #2162, opened 2024-05-10
DAH-2312 do not show base preferences for SCV 1

tallulahkay:
DAH-2312


alulabeshue-sfgov (approved):
LGTM! Validated in the PR app that educator one listing is not showing the base preference for both T1 and T2 preferences. Automated tests also look good.

PR #2163, opened 2024-05-13
fix: DAH-2313 run for all jobs

tallulahkay:
DAH-2313

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -288,7 +288,8 @@ workflows:
           <<: *prod_and_release_jobs
           name: e2e-production-legacy
       - waiter:
-          <<: *angular_jobs
+          requires:
+            - build
       - cleanup-test-data:
           requires:
             - waiter


tallulahkay:
@jimlin-sfgov I'm not sure how to test this

.circleci/config.yml
```diff
@@ -288,7 +288,8 @@ workflows:
           <<: *prod_and_release_jobs
           name: e2e-production-legacy
       - waiter:
-          <<: *angular_jobs
+          requires:
+            - build
```
jimlin-sfgov, line 292: question: What's the difference between this and adding a second `waiter` in the array? e.g.

```yaml
- waiter:
  <<: *angular_jobs
- waiter:
  <<: **prod_and_release_jobs
```

tallulahkay: the requires: build is just something that was missing before

removing the <<: <job> just means it will run always instead of specifically for both types of jobs (those are our only two jobs)

jimlin-sfgov:
> @jimlin-sfgov I'm not sure how to test this

@tallulahkay You can make a hotfix branch with these changes (but not actually merge it), prod and hotfix branches should have the same pipeline

tallulahkay:
@jimlin-sfgov cool, looks like it works: https://app.circleci.com/pipelines/github/SFDigitalServices/sf-dahlia-web/4871/workflows/dfb55b92-764d-4bbc-ac5f-7b7922948af0

jimlin-sfgov (approved):
looks like its working 👍

PR #2176, opened 2024-05-22
fix: update google translate bundle

cade-exygy:
Fixes CircleCI error on main: https://app.circleci.com/pipelines/github/SFDigitalServices/sf-dahlia-web/4951/workflows/2a502181-9669-476d-bc33-2aac8c11337a/jobs/31977

Merge into main to allow build to happen. However, ruby tests will fail due to the faraday version change.

Test by pulling locally and running bundle install. Then ensure app builds


Gemfile.lock
```diff
@@ -197,7 +197,7 @@ GEM
       faraday (>= 1.0, < 3.a)
     google-cloud-errors (1.4.0)
     google-cloud-translate-v2 (0.5.0)
-      faraday (>= 0.17.3, < 3.a)
+      faraday (>= 0.17.3, < 2.a)
```
jimlin-sfgov, line 200: suggestion: Manually updating the lock file isn't a great idea, and probably why the deployment failed. You could try to explicitly set the faraday gem version in the gemfile. All of our gems seem to support faraday v1 as a dependency, so you can try to use version 1.10.3.

jimlin-sfgov:
@cade-exygy It must be your lucky day, google just updated the gem dependency, so we shouldn't need to do anything now: https://github.com/googleapis/google-cloud-ruby/releases/tag/google-cloud-translate-v2%2Fv0.5.1

Gemfile
```diff
@@ -61,6 +61,7 @@ gem 'easypost', '>=3.0.1'
 gem 'StreetAddress', github: 'Exygy/street-address', require: 'street_address'
 
 # google translation
+gem 'faraday'
```
jimlin-sfgov, line 64: suggestion: we don't actually need this anymore, but it shouldn't make a difference

jimlin-sfgov (approved):
good to go

PR #2200, opened 2024-06-17
feat: geocode param from Street to SingleLine

alulabeshue-sfgov:
[DAH-2376](https://sfgovdt.jira.com/browse/DAH-2376)

## Change
When migrating to the new GIS server we were getting incorrect responses, for example entering a san francisco address was returning a coordinate in south san francisco. After investigating DT recommended we migrate to using `SingleLine` rather than the `Street` parameter. The argument is that `SingleLine` will better parse the format we pass in such as `1550+Mission+St`

[DAH-2376]: https://sfgovdt.jira.com/browse/DAH-2376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jimlin-sfgov (approved):
lgtm

PR #2203, opened 2024-06-27
feat: DAH-2372 auth account boilerplate

cade-exygy:
https://sfgovdt.jira.com/browse/DAH-2372

Review Instructions

- Go to http://localhost:3000/sign-in?react=true
- Go to http://localhost:3000/create-account?react=true
- Go to http://localhost:3000/forgot-password?react=true
- Go to http://localhost:3000/reset-password?react=true


app/assets/javascripts/shared/SharedService.js.coffee
```diff
@@ -75,7 +75,15 @@ SharedService = ($http, $state, $window, $document) ->
...
-
+    'dahlia.sign-in':
+      buildUrl: (state, params) -> Service._addLanguageAndParamsToUrl(params.lang, "/sign-in")
+      shouldRailsRoute: (isFirstLoad) -> !isFirstLoad && $window.ACCOUNT_INFORMATION_PAGES_REACT is "true"
+    'dahlia.create-account':
+      buildUrl: (state, params) -> Service._addLanguageAndParamsToUrl(params.lang, "/create-account")
+      shouldRailsRoute: (isFirstLoad) -> !isFirstLoad && $window.ACCOUNT_INFORMATION_PAGES_REACT is "true"
+    'dahlia.forgot-password':
+      buildUrl: (state, params) -> Service._addLanguageAndParamsToUrl(params.lang, "/forgot-password")
+      shouldRailsRoute: (isFirstLoad) -> !isFirstLoad && $window.ACCOUNT_INFORMATION_PAGES_REACT is "true"
```
chadbrokaw, line 86: Nice! Can you change these so that they are using different env variables (also adding /reset-password) in there. I realize this wasn't super clear in the ticket, so I changed the description/AC to make that clearer

cade-exygy: Ah nice catch, I knew I was forgetting something. What env variable should we use? Or do we want a new one?

cade-exygy: Pushed those changes, having some trouble testing reset password, maybe pair quick tomorrow?

chadbrokaw (commented):
Thank you for making those changes! Two questions:
- When I go to the react version of Reset Password I just get a blank screen with no text?
- Would you mind adding a quick test for both the reset password page?

chadbrokaw (approved):
Looking at the Code Climate errors, I'm not sure that we can do anything about the coverage errors as most of the code added in this PR, with the exception of the new React pages, is more at the infrastructure level and would be difficult to test.

PR #2208, opened 2024-07-02
DAH-2283 Listen to Salesforce Events POC

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2283


spec/services/event_subscriber_translate_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      Force::EventSubscriberTranslateService.new.listen_and_process_events
+    end
+
+    # How to test subscription events? Maybe the Faye gem's test suite has clues:
+    #   https://github.com/faye/faye/blob/master/spec/ruby/client_spec.rb
+    # it 'detects events from Salesforce and translates changed fields' do
+    #   Force::EventSubscriberTranslateService.new.listen_and_process_events
+    # end
+  end
+end
```
jimlin-sfgov, line 47: note: testing `eventmachine` code is difficult. I'm not sure it's appropriate to stub so much stuff. I also don't know how to mock a subscriber event.

jimlin-sfgov:
@cade-exygy Eagerly awaiting your feedback 🙂

spec/services/event_subscriber_translate_service_spec.rb
cade-exygy: I will look into this. Do you think this should be a blocker for merge?

jimlin-sfgov: This isn't a blocker for a merge if this just a POC that will not be used in production. I can add a coverage exception in the config file.

This will become a blocker later when we transition from a POC to code that runs in production.

cade-exygy: This test should work for testing a event coming through!

jimlin-sfgov: running an EM loop to test an EM loop, I would have never thought of that

cade-exygy (approved):
Looks good, I have consistently been receiving platform events when testing.

PR #2209, opened 2024-07-02
feat/DAH-2263 cache translations with listings

tallulahkay:
DAH-2263


app/services/google_translation_service.rb
```diff
@@ -7,6 +7,38 @@ class TranslationError < StandardError; end
...
+      target[:translation].each_with_index do |value, i|
+        field = keys[i].to_sym
+        return_value[field] ||= {}
+        return_value[field][target[:to].to_sym] = value
+      end
+    end
+    # alternatively, we could splat translations into the listing object
+    # so they aren't nested under the translations key, but instead under each key: {**listing, **return_value}
+    # gets complicated with nested translations (like listing image descriptions)
+    listing[0][:translations] = return_value
```
cade-exygy, line 30: We may also run into the issue of overwriting previous set translations. For example if someone updates "Field A", then we will hit this service to translate that field. Then if someone updates "Field B" we hit this service again but set the translations object to just the result of "Field B".

Using this pattern, we could just do like an update (not exactly this, but similar concept): `listing[0][:translations] = [...translations, return_value[field_to_update]`

And we could also consider batching field updates by setting up a queue.

tallulahkay: I fixed the overwriting issue you mentioned and added a test case that covers it

app/services/google_translation_service.rb
```diff
@@ -19,9 +20,49 @@ def translate(text, to)
...
     results.map(&:text)
   end
+
+  def transform_translations_for_caching(listing_id, keys, translations)
+    listing = Force::ListingService.cached_listing(listing_id)
+    prev_cached_translations = listing[0][:translations] || {}
+    # keys can come from updated_values.keys in the event_subscriber_translate_service
+    # they will be in the same order as the translations because the translation service
+    # uses the values from that object this would make the assumption that every value
+    # comes back with 1 translation for each translation language (we would have a problem
```
tallulahkay, line 51: Even if google doesn't return a translation for a language, the api will still return a field so this is not a possibility

app/services/google_translation_service.rb
```diff
@@ -19,9 +20,49 @@ def translate(text, to)
...
+    # if it came back with just 2 in the translation array but were expecting 3 fields)
+    return_value = {}
+    translations.each do |target|
+      target[:translation].each_with_index do |value, i|
+        field = keys[i].to_sym
+        return_value[field] ||= {}
+        return_value[field][target[:to].to_sym] = value
+      end
+    end
+    # alternatively, we could splat translations into the listing object
```
tallulahkay, line 61: can remove comment

tallulahkay:
@cade-exygy I updated based on our conversation!

jimlin-sfgov (approved):
lgtm. Did you want a more rigorous review from @cade-exygy, who has been leading the machine translation project?

tallulahkay:
@jimlin-sfgov I think your review is enough! Cade can look at is as a PA tester

cade-exygy:
Loos good!

PR #2211, opened 2024-07-08
hotfix: DAH-2416 add invalid character validation

cade-exygy:
Changes related to the issues documented in this ticket: https://sfgovdt.jira.com/browse/DAH-2416

Added a check for the firstName, lastName, and middleName parameters coming from the create account flow:
If any of those fields contains 'http' or 'www', we render an error.

Review Instructions
- Go to create an account
- In First Name, Middle Name, and Last Name fields: try to add a url and see if you can create an account successfully.
- Finally, remove any invalid characters and ensure you can create an account.
- I added a quick unit test, hopefully that passes.

We will need to review our validation practices going forward. I think there are some opportunities for refactoring here. This should address part of the issue (next we need to look at API limiting), but I am curious what other validations we may want to add here.

Additionally, I think we should also add validation checks to the Salesforce code.

### DIFF
--- a/app/controllers/overrides/registrations_controller.rb
+++ b/app/controllers/overrides/registrations_controller.rb
@@ -43,7 +43,7 @@ module Overrides
     end
 
     def includes_url_characters(value)
-      value.include?('www') || value.include?('http')
+      value.include?('www') || value.include?('http') || value.include?('.')
     end
 
     def setup_resource_for_create
--- a/app/controllers/overrides/registrations_controller.rb
+++ b/app/controllers/overrides/registrations_controller.rb
@@ -38,7 +38,7 @@ module Overrides
     end
 
     def name_fields_have_invalid_characters?
-      includes_url_characters(params[:contact][:firstName]) || includes_url_characters(params[:contact][:lastName]) || includes_url_characters(params[:contact][:middleName])
+      includes_url_characters(params[:contact][:firstName]) || includes_url_characters(params[:contact][:lastName])
     end
 
     def includes_url_characters(value)
--- a/app/controllers/overrides/registrations_controller.rb
+++ b/app/controllers/overrides/registrations_controller.rb
@@ -4,19 +4,20 @@ module Overrides
     # method copied from original gem; refactored to please Rubocop
     def create
       setup_resource_for_create
+      return render_create_error if name_fields_have_invalid_characters?
+
       if resource_class.devise_modules.include?(:confirmable) && !@redirect_url
         return render_create_error_missing_confirm_success_url
       end
 
       # if whitelist is set, validate redirect_url against whitelist
-      if DeviseTokenAuth.redirect_whitelist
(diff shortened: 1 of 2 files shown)


app/controllers/overrides/registrations_controller.rb
```diff
@@ -4,19 +4,20 @@ class RegistrationsController < DeviseTokenAuth::RegistrationsController
...
       if resource_class.devise_modules.include?(:confirmable) && !@redirect_url
         return render_create_error_missing_confirm_success_url
       end
 
       # if whitelist is set, validate redirect_url against whitelist
-      if DeviseTokenAuth.redirect_whitelist
-        unless DeviseTokenAuth.redirect_whitelist.include?(@redirect_url)
-          return render_create_error_redirect_url_not_allowed
-        end
+      if DeviseTokenAuth.redirect_whitelist && !DeviseTokenAuth.redirect_whitelist.include?(@redirect_url)
```
cade-exygy, line 14: This is just a result of Ruby auto formatting

app/controllers/overrides/registrations_controller.rb
```diff
@@ -36,6 +37,14 @@ def error_checks_for_create
       # success redirect url is required
     end
 
+    def name_fields_have_invalid_characters?
+      includes_url_characters(params[:contact][:firstName]) || includes_url_characters(params[:contact][:lastName]) || includes_url_characters(params[:contact][:middleName])
```
jimlin-sfgov, line 41: suggestion: Concatenate values, e.g.
```ruby
includes_url_characters(params[:contact].values.join)
```

cade-exygy: I decided we don't need to include `middleName` because it is not included in the email template, and since it is an optional field.

app/controllers/overrides/registrations_controller.rb
jimlin-sfgov: note: This formatting is causing the diff coverage check to complain, we can ignore.

app/controllers/overrides/registrations_controller.rb
fwextensions: A string like `bitcoin.ru/get-rich` would pass the checks, but might still be something that would be auto-linked in an email client.

cade-exygy: We can add the `.` check back if we think that is needed.

jimlin-sfgov:
@cade-exygy Target branch should be `production`, if this is truly a hotfix.

fwextensions:
Can we ignore the code coverage for now?

cade-exygy:
> @cade-exygy Target branch should be `production`, if this is truly a hotfix.

@alulabeshue-sfgov thoughts on this in regards to what you have suggested as a change in how we handle hotfixes (Eg. push to main always)

We can cherrypick this commit to the prod branch

fwextensions (approved):
Ship it

PR #2216, opened 2024-07-22
DAH-2465 Add optional resend notice to email template

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2465


chadbrokaw (approved):
Code looks good. Followed review instructions and was able to get an email delivered in all languages

PR #2217, opened 2024-07-23
DAH-2434 Remove name from email layout

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2434

### DIFF
--- a/spec/mailers/emailer_spec.rb
+++ b/spec/mailers/emailer_spec.rb
@@ -25,9 +25,9 @@ describe Emailer, type: :mailer do
 
     it 'renders the body' do
       VCR.use_cassette('emailer/submission_confirmation') do
-        name = "#{@params[:firstName]} #{@params[:lastName]}"
+        # name = "#{@params[:firstName]} #{@params[:lastName]}"
         text = 'Thanks for applying. We have received your application'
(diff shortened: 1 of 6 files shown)


spec/mailers/emailer_spec.rb
```diff
@@ -98,7 +98,8 @@
       mail = Emailer.account_update(@record)
 
       VCR.use_cassette('emailer/account_update') do
-        name = 'Hello Jane Doe,'
+        # name = 'Hello Jane Doe,'
```
tallulahkay, line 101: Why leave the old functionality in comments in all these tests?

jimlin-sfgov: We'll need them later for the follow up ticket: https://sfgovdt.jira.com/browse/DAH-2471

lmk if you prefer to just remove lines for this kind of situation

tallulahkay: All good to leave them then! Could you include a todo code comment that references that ticket?

cade-exygy: I am okay with the comments in this case

tallulahkay (approved):
Nice!

PR #2218, opened 2024-07-24
DAH-2466 More Logging to recover from incidents

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2466


tallulahkay (approved):
Works flawlessly!

PR #2219, opened 2024-07-25
fix/DAH-2479 update spacing

tallulahkay:
DAH-2479


chadbrokaw (approved):
Nice! The only thing that I noticed was that the padding around the form sections is 24px when in the Jira it seems like it should be 32px? 
Screenshot 2024-07-29 at 1 52 12 PM

tallulahkay:
@chadbrokaw it's 32px when you add both the paddings:
Screenshot 2024-07-29 at 6 47 14 PM
Screenshot 2024-07-29 at 6 47 22 PM

PR #2220, opened 2024-07-25
fix: DAH-2490 use address instead of singleline

alulabeshue-sfgov:
[DAH-2490](https://sfgovdt.jira.com/browse/DAH-2490)

[DAH-2490]: https://sfgovdt.jira.com/browse/DAH-2490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alulabeshue-sfgov:
> Alameda address I tested returned "boundary_match":null
Thanks for reviewing @cliu02! I agree this is not a blocker, I've reached out to DT about some Alameda addresses that are not getting geocodes.

PR #2221, opened 2024-07-29
feat: Change Account Info API

chadbrokaw:
[DAH-2385](https://sfgovdt.jira.com/browse/DAH-2385)

[DAH-2385]: https://sfgovdt.jira.com/browse/DAH-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/overrides/registrations_controller.rb
+++ b/app/controllers/overrides/registrations_controller.rb
@@ -4,7 +4,7 @@ module Overrides
     # method copied from original gem; refactored to please Rubocop
     def create
       setup_resource_for_create
-      return render_create_error if name_fields_have_invalid_characters? || !AccountValidationService.valid_dob?(account_params[:DOB])
+      return render_create_error if AccountValidationService.name_fields_have_invalid_characters?(params[:contact]) || !AccountValidationService.valid_dob?(account_params[:DOB])
 
       if resource_class.devise_modules.include?(:confirmable) && !@redirect_url
         return render_create_error_missing_confirm_success_url
@@ -35,15 +35,6 @@ module Overrides
 
     def error_checks_for_create
       # success redirect url is required
-    end
-
-    def name_fields_have_invalid_characters?
-      contact_names = [params[:contact][:firstName], params[:contact][:lastName]]
-      contact_names.any? { |name| includes_url_characters(name) }
-    end
-
-    def includes_url_characters(value)
-      value.include?('www') || value.include?('http') || value.include?('.')
     end
 
     def setup_resource_for_create
--- a/app/controllers/overrides/registrations_controller.rb
+++ b/app/controllers/overrides/registrations_controller.rb
@@ -4,7 +4,7 @@ module Overrides
     # method copied from original gem; refactored to please Rubocop
     def create
       setup_resource_for_create
-      return render_create_error if name_fields_have_invalid_characters?
+      return render_create_error if name_fields_have_invalid_characters? || !AccountValidationService.valid_dob?(account_params[:DOB])
 
       if resource_class.devise_modules.include?(:confirmable) && !@redirect_url
         return render_create_error_missing_confirm_success_url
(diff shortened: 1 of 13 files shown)


app/controllers/overrides/registrations_controller.rb
```diff
@@ -4,7 +4,7 @@ class RegistrationsController < DeviseTokenAuth::RegistrationsController
     # method copied from original gem; refactored to please Rubocop
     def create
       setup_resource_for_create
-      return render_create_error if name_fields_have_invalid_characters?
+      return render_create_error if name_fields_have_invalid_characters? || !AccountValidationService.valid_dob?(account_params[:DOB])
```
cade-exygy, line 7: We can probably add name_fields_have_invalid_characters and any future validation to the AccountValidationService

app/javascript/api/authApiService.ts
```diff
@@ -42,3 +42,15 @@ export const updatePassword = async (
     passwordConfirmation: passwordConfirmation,
     token: token,
   }).then(({ data }) => data.accessToken)
+
+export const updateNameOrDOB = async (user: User): Promise<User> => {
```
cade-exygy, line 46: nit: It looks email can be updated in this function too? If so, consider renaming the function to reflect that. `updateUserProfile`?

app/javascript/pages/account/account-settings.tsx
```diff
@@ -123,35 +124,54 @@ const PasswordSection = ({ user, setUser }: SectionProps) => {
...
+  setUser: React.Dispatch<User>,
+  setLoading: React.Dispatch<boolean>
+) => {
+  return apiUpdateNameOrDOB(newUser)
+    .then((profile) => {
+      saveProfile(profile)
+      setUser(newUser)
+    })
+    .catch((error) => {
+      // TODO: In the case that a user's DOB is invalid, this is a snippet of the AxiosError that will be returned
```
cade-exygy, line 139: Can we include the logic now or does this have to wait for the future error banner ticket?

app/services/account_validation_service.rb
```diff
@@ -0,0 +1,10 @@
+class AccountValidationService
```
cade-exygy, line 1: nit: RuboCop extension has some formatting suggestions

cade-exygy (approved):
Test criteria passes!

app/javascript/api/authApiService.ts
chadbrokaw: Updating the email will need to be done in a different function as the endpoint for updating email is different (https://docs.google.com/document/d/18gurkqTOat8X55NHTFOoGsE16JqSqji71XwPTlcYlNI/edit#heading=h.t9yx58fn9p4l)

app/javascript/pages/account/account-settings.tsx
chadbrokaw: The banner logic will be implemented in a future ticket

app/controllers/overrides/registrations_controller.rb
chadbrokaw: I moved it to AccountValidationService, can you review that in the PR app to make sure everything is working properly?

app/services/account_validation_service.rb
chadbrokaw: Thanks for flagging, I fixed the errors

app/javascript/api/authApiService.ts
cade-exygy: Ah okay, so will the email field in this function be removed after that? Or is the email required to be sent alongside the name and DOB?

app/controllers/overrides/registrations_controller.rb
cade-exygy: Looks good!

app/javascript/api/authApiService.ts
chadbrokaw: Based on my testing I don't think that the email field is required, but it is included in the AngularJS version of this call so I included it as well. I figured it would be worthwhile to include just in case it is needed somewhere that I didn't see

PR #2222, opened 2024-07-30
feat: DAH-2343 Account Settings banners frontend

cliu02:
[DAH-2343](https://sfgovdt.jira.com/browse/DAH-2343)

[DAH-2343]: https://sfgovdt.jira.com/browse/DAH-2343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/pages/account/account-settings.tsx
+++ b/app/javascript/pages/account/account-settings.tsx
@@ -125,10 +125,9 @@ const EmailSection = ({ user, setUser, handleBanners }: SectionProps) => {
         }
         setUser(newUser)
         handleBanners("emailBanner")
-        // TODO(DAH-2343): Inform the user that they will need to verify their new email address
       })
       .catch((error) => {
-        // TODO(DAH-2343): Inform the user that an error has occurred
+        // TODO: Inform the user that an error has occurred
         // In the case that the email is malformed, the error will have code 422
         console.log(error)
       })


app/javascript/pages/account/account-settings.tsx
```diff
@@ -258,10 +341,33 @@ const AccountSettings = ({ profile }: { profile: User }) => {
...
-            <DateOfBirthSection user={user} setUser={setUser} />
-            <EmailSection user={user} setUser={setUser} />
-            <PasswordSection user={user} setUser={setUser} />
+            {banners.nameUpdateBanner && (
+              <span className="mb-8">
+                <UpdateBanner />
+              </span>
+            )}
+            {banners.nameSavedBanner && <SavedBanner />}
+            <NameSection user={user} setUser={setUser} handleBanners={handleBanners} />
```
chadbrokaw, line 350: I wonder if we should pass the banners as a prop to their respective sections so that it is clearer which banner goes with each section?

app/javascript/pages/account/account-settings.tsx
```diff
@@ -173,8 +210,13 @@ const NameSection = ({ user, setUser }: SectionProps) => {
     handleSubmit,
   } = useForm({ mode: "all" })
 
+  const onChange = () => {
+    handleBanners("nameUpdateBanner")
+  }
+
   const onSubmit = async (data: { firstName: string; middleName: string; lastName: string }) => {
     setLoading(true)
+    handleBanners("nameSavedBanner")
```
chadbrokaw, line 219: Should this be passed to the updateNameOrDOB function as a prop so that it can happen in the `.then` block? This way if there is an error we aren't showing the saved banner

app/javascript/pages/account/account-settings.tsx
```diff
@@ -211,8 +254,13 @@ const DateOfBirthSection = ({ user, setUser }: SectionProps) => {
     watch,
   } = useForm({ mode: "all" })
 
+  const onChange = () => {
+    handleBanners("dobUpdateBanner")
+  }
+
   const onSubmit = async (data: { dob: DOBFieldValues }) => {
     setLoading(true)
+    handleBanners("dobSavedBanner")
```
chadbrokaw, line 263: Same question as above about passing this to the updateNameOrDOB function

chadbrokaw (commented):
Looks good! Looking at the Figma and this, I see two things:
- On the DOB section, there is not a gap between the banner and the top border
- In the designs, when there is a banner in the name section, the top border for the name section should go away

Would you also mind adding some tests to cover the banners?

app/javascript/pages/account/account-settings.tsx
chadbrokaw:
```suggestion
            <NameSection user={user} setUser={setUser} handleBanners={handleBanners} banners={<>
            {banners.nameUpdateBanner && (
              <span className="mb-8">
                <UpdateBanner />
              </span>
            )}
            {banners.nameSavedBanner && <SavedBanner />}
            </>} />
```

or even

```suggestion
            <NameSection user={user} setUser={setUser} handleBanners={handleBanners} banners={banners} />
```

Or something like that. What do you think?

chadbrokaw: Ah, good point! Okay, that makes sense

chadbrokaw:
> > On the DOB section, there is not a gap between the banner and the top border
> 
> Thank you for reviewing @chadbrokaw! Could you clarify which DOB banner you're referencing? Here's what I see on Figma: Screenshot 2024-07-31 at 3 16 11 PM

Sorry that was unclear! In the designs there is a gap between the DOB banner and the top border, but I'm not seeing that in the PR Deployment
Screenshot 2024-08-01 at 8 42 20 AM

app/javascript/pages/account/account-settings.tsx
```diff
@@ -220,7 +281,14 @@ const DateOfBirthSection = ({ user, setUser }: SectionProps) => {
...
     }
 
-    await updateNameOrDOB(newUser, saveProfile, setUser, setLoading)
+    await updateNameOrDOB(
+      newUser,
+      saveProfile,
+      setUser,
+      setLoading,
+      handleBanners,
+      "dobSavedBanner"
```
chadbrokaw, line 290: Nice! I think this can be simplified more so that it is just passed as one function. Something like 
```suggestion
     () => handleBanners("dobSavedBanners")
```
and the type on the `updateNameOrDOB` could be: `bannersCallback?: () => void`

app/javascript/pages/account/account-settings.tsx
```diff
@@ -14,16 +14,46 @@ import FormSubmitButton from "./FormSubmitButton"
 import PasswordFieldset from "./PasswordFieldset"
 import NameFieldset from "./NameFieldset"
 import DOBFieldset from "./DOBFieldset"
+import "./account-settings.scss"
```
tallulahkay, line 17: does putting this here bubble down to all the components inside?

app/javascript/pages/account/account-settings.tsx
```diff
@@ -89,6 +124,7 @@ const EmailSection = ({ user, setUser }: SectionProps) => {
           email,
         }
         setUser(newUser)
+        handleBanners("emailBanner")
         // TODO(DAH-2343): Inform the user that they will need to verify their new email address
```
chadbrokaw, line 128: Would you mind removing this comment for me?

chadbrokaw (approved):
Looks great!

PR #2224, opened 2024-07-31
DAH-2471 Name Fields for confirmed-account Emails

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2471


chadbrokaw (approved):
Looks good!

PR #2228, opened 2024-08-05
DAH 2389 create account react

tallulahkay:
DAH-2389


app/javascript/pages/account/create-account.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  return (
+    <CreateAccountFormSection>
+      <NameFieldset
+        register={register}
+        errors={errors}
+        defaultFirstName={null}
+        defaultMiddleName={null}
+        defaultLastName={null}
+      />
```
chadbrokaw, line 39: Would if be better to make these default fields optional? That way we aren't having to pass `null`

app/javascript/pages/account/create-account.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <Button className="uppercase" type="button" onClick={signInRedirect}>
+        {t("label.signIn")}
+      </Button>
+    </Card.Section>
+  )
+}
+
+const onSubmit = () => {
+  console.log("hello")
+}
```
chadbrokaw, line 126: For this, how will we get the form values to this submit function?

chadbrokaw: Relatedly, how will errors be propagated up to the `CreateAccount` component? 

Would it be better to use one large "useForm" hook instead of making multiple mini forms?

chadbrokaw (commented):
Looks good! I noticed three potential design differences when comparing to the Figma:

- "We ask for date of birth to verify your age. You must be 18 or older to apply for housing." should be 13px font size
- "We only use your email to contact you about an application you submitted or to help you reset your password." should be 13px font size
- The "Already have an account?" header at the bottom needs to have a serif font and be bigger

app/javascript/pages/account/create-account.tsx
tallulahkay: oooh yeah

app/javascript/pages/account/create-account.tsx
tallulahkay: Yes I didn't think this through - thank you! Made that update, so lmk if that looks better

chadbrokaw (approved):
Nice! Only remaining thing is the "We use this information to help you fill in your application. You can update these details at any time." text should be 13px font

tallulahkay:
> Nice! Only remaining thing is the "We use this information to help you fill in your application. You can update these details at any time." text should be 13px font

Yindi said that it actually should be 14px, so I'm going to merge!

PR #2226, opened 2024-08-01
fix: DAH-2512 Remove use of deprecated lottery field in Angular condition logic

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2512

PR #2229, opened 2024-08-06
test: DAH-2529 Fix angular unit tests

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2529


tallulahkay (approved):
ty for fixing this!

alulabeshue-sfgov:
1. The next time someone runs rspec again the fixtures are being reverted. Do we need to prevent these from updating or prevent engs from committing these changes if it breaks tests?
2. The [regex](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/.circleci/config.yml#L31) did not match to your branch when it should have. The angular-unit tests should have ran as part of the original [PR](https://github.com/SFDigitalServices/sf-dahlia-web/pull/2226).

jimlin-sfgov:
@alulabeshue-sfgov 

 > The next time someone runs rspec again the fixtures are being reverted. Do we need to prevent these from updating or prevent engs from committing these changes if it breaks tests?

We need to prevent engs from committing the changes. The root cause is that Angular fixtures are auto-generated from VCRs (see `spec/support/jasmine_macros.rb`), but our VCRs are not in a good state... I tried re-recording all of our VCRs and got a bunch of Rspec failures. We will have to look into this later.

> The [regex](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/.circleci/config.yml#L31) did not match to your branch when it should have. The angular unit tests should have ran as part of the original https://github.com/SFDigitalServices/sf-dahlia-web/pull/2226.

Yeah, that one is on me. I forget that I need to prepend 'angular' to the branch name when I haven't worked on Angular-related PRs in a while.

PR #2235, opened 2024-08-07
DAH-2534 fix: Account Settings banner functionality

cliu02:
[DAH-2534](https://sfgovdt.jira.com/browse/DAH-2534)

[DAH-2534]: https://sfgovdt.jira.com/browse/DAH-2534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/pages/account/account-settings.tsx
+++ b/app/javascript/pages/account/account-settings.tsx
@@ -97,11 +97,13 @@ const UpdateForm = ({
 interface SectionProps {
   user: User
   setUser: React.Dispatch<User>
-  handleBanners: (banner: string) => void
+  handleBanners?: (banner: string) => void
 }
 
-const EmailSection = ({ user, setUser, handleBanners }: SectionProps) => {
+const EmailSection = ({ user, setUser }: SectionProps) => {
   const [loading, setLoading] = useState(false)
+  const [emailUpdateBanner, setemailUpdateBanner] = useState(false)
+  const [emailBanner, setemailBanner] = useState(false)
 
   const {
     register,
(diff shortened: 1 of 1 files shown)


tallulahkay (changes requested):
Could you do a structural update where the logic for the banners is moved into the section components? 

I think the switch statement and banner state object at the top level is getting a little unwieldy. Each component can have it's own state that manages their respective banners and maintain the same functionality.

For example in the `NameSection` component, there would be these additional state variables:
```
const [nameUpdatedBanner, setNameUpdatedBanner] = useState(false)
const [nameSavedBanner, setNameSavedBanner] = useState(false)
```
onchange would call:
```
const onChange = () => {
  setNameUpdatedBanner(true)
  setNameSavedBanner(false)
}
```
the update call would look like this:
```   
await updateNameOrDOB(newUser, saveProfile, setUser, setLoading, () => setNameSavedBanner(true))
```
This logic would go in the return clause:
```
{nameUpdatedBanner && <UpdateBanner />}
{nameSavedBanner && (
  <span className="mt-8">
    <SavedBanner />
  </span>
)}
```

And then `handleBanners` wouldn't have to be passed into each section

Let me know what you think!! Happy to pair on this

tallulahkay:
> Thank you for reviewing @tallulahkay! It's not currently structured at the section component level because each Seeds Card section has width padding and the banners are styled to be at the top card level, so component styling would need to shuffle around.

@cliu02 They would still be outside the seeds Card.Section:
In NameSection:

```
return 
<>
      {nameUpdatedBanner && <UpdateBanner />}
      {nameSavedBanner && (
        <span className="mt-8">
          <SavedBanner />
        </span>
      )}
      <UpdateForm onSubmit={handleSubmit(onSubmit)} loading={loading}>
        <NameFieldset
          register={register}
          errors={errors}
          defaultFirstName={user?.firstName ?? null}
          defaultMiddleName={user?.middleName ?? null}
          defaultLastName={user?.lastName ?? null}
          onChange={onChange}
        />
      </UpdateForm>
</>
```

tallulahkay (approved):
This looks and works amazing!! thank you for making that change I know it was a big one

PR #2238, opened 2024-08-08
DAH-2388: feat: front end email validation

chadbrokaw:
[DAH-2388](https://sfgovdt.jira.com/browse/DAH-2388)

[DAH-2388]: https://sfgovdt.jira.com/browse/DAH-2388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fwextensions:
Can't tell from the diff, but does this component show an error while the user is typing into an empty field?  The current Angular page's behavior is to not show an error until the user tabs away.  That way, they don't get an error just because they haven't finished typing yet.  What I see in the account creation PR is that you get an error message as soon as you start typing in a blank field, which isn't correct.

chadbrokaw:
> Can't tell from the diff, but does this component show an error while the user is typing into an empty field? The current Angular page's behavior is to not show an error until the user tabs away. That way, they don't get an error just because they haven't finished typing yet. What I see in the account creation PR is that you get an error message as soon as you start typing in a blank field, which isn't correct.

Yes, the component will show an error as the user is typing. However that behavior is determined when the form handler is instantiated with the useForm hook (e.g. https://github.com/SFDigitalServices/sf-dahlia-web/blob/c774cfd9ac1577d52410b7258c567e7b384898a5/app/javascript/pages/account/account-settings.tsx#L108-L112) and not in this component.

I will start a thread in Slack to discuss this more

PR #2244, opened 2024-08-14
feat: DAH-2280 return translations from listing service

cade-exygy:
## Description

Here is a demo video: https://www.loom.com/share/b6a0a8b2e0ab43c3a4c4cf32e9193aec?sid=332789a7-6f57-4a59-95e2-a0db5f3c4204

Listing Service now returns cached translations.

#Jira Ticket
DAH-2280

### Version Control

- [x] branch name begins with `angular` if it contains updates to Angular code
- [x] branch name contains the Jira ticket number
- [x] PR name follows `type: TICKET-NUMBER Description` format, e.g. `feat: DAH-123 New Feature`

### Code quality

- [x] [the set of changes is small](https://google.github.io/eng-practices/review/developer/small-cls.html#what-is-small)
- [x] [if the set of changes cannot be small, reviewers have been forewarned](https://google.github.io/eng-practices/review/developer/small-cls.html#cant)
- [x] code meets test coverage thresholds
- [x] code is properly formatted
- [x] code is linted
- [x] code irrelevant to the ticket is not modified e.g. changing indentation due to automated formatting
- [x] automated tests pass
- [x] if the code changes the UI, it matches the UI design exactly

### Review instructions

- [x] instructions specify which environment(s) it applies to
- [x] instructions have already been performed at least once
- [x] instructions can be followed by PA testers
- [x] instructions specify if it can only be followed by an engineer

### Request review

- [x] PR has `needs review` label
- [x] Use `Housing Eng` group to automatically assign reviewers, and/or assign specific engineers
- [x] If time sensitive, notify engineers in Slack

Locally (Engineers only since we don't run the subscriber on start):
Make sure you have the env variables set:
GOOGLE_PROJECT_ID
GOOGLE_TRANSLATE_KEY

- run memcached
- run `yarn start`
- run `bundle exec rake translate_test:subscribe`
- run `tail -f log/development.log`
- Update a listing in salesforce, edit Credit History if you want to see the value appear on the listing details page. Credit History is just the field I used as example to ease testing. See #2249 
- You should receive an event and see those values get translated in your logs
- Go to that listing and view the network object. You should see those updated translations
- If you scroll down to Credit History, you should see that updated value.

PR #2255, opened 2024-08-21
feat: DAH-2451 cache service gets translations on listing translations cache miss

cade-exygy:
## Description

When we hit the listing detail service, we return the listing translations alongside the listing. If the listing translations do not exist, we need to call the translation service to grab them and cache them.

## Jira ticket
DAH-2451
relies on #2244 being merged

### Version Control

### Code quality

- [x] [if the set of changes cannot be small, reviewers have been forewarned](https://google.github.io/eng-practices/review/developer/small-cls.html#cant)
- [x] code meets test coverage thresholds
- [x] code is properly formatted
- [x] code is linted
- [x] automated tests pass

### Request review

### Review instructions

- [x] instructions can be followed by PA testers
- [x] instructions specify if it can only be followed by an engineer

On localhost:
run memcached
Go to a listing
Check logs for `CacheService.processTranslations` being called
Inspect in Network Tab the Listing response and ensure the translation object is populated.

If you refresh, you should not see this call happen again because the `cache.fetch()` call should hit successfully.


tallulahkay (approved):
This works well!! I love how small this ticket is - made it really easy to review 🤩 

I worry we are clogging up the logs because the translation object logs are so big.. maybe we can shorten to just success or failure messages, or just log the keys or somethin

tallulahkay:
What if this went in first? Would we hit the google translation api too much?

tallulahkay:
@cade-exygy I found an issue with this!! if the cache is not running, it doesn't return the translations

cade-exygy:
> @cade-exygy I found an issue with this!! if the cache is not running, it doesn't return the translations I think it needs to return translations whether or not it successfully caches them in `cache_listing_translations`

Nice catch! I put that fix in #2244 to keep these changes focused on the listing service

tallulahkay:
This doesn't _quite_ work because if there are translations for a listing, but incomplete ones (like for only one field, for ex) then it doesn't get the translations. We need to do some sort of check to make sure the cache has all the translations we need

tallulahkay (approved):
lgtm after the test is updated!

As soon as we merge this, we will start hitting the google cloud translate api - is full prepared for that?

spec/services/cache_service_spec.rb
```diff
@@ -134,6 +134,8 @@
       # mock cached listing
       listing['Listing_Other_Notes'] = 'Test Notes'
       listing['Realtor_Commission_Info'] = 'Test Commission Info'
+
+      puts listing
```
jimlin-sfgov, line 138: nit: I'm assuming this was for debugging, we should remove them from tests.

jimlin-sfgov (approved):
LG

one nit suggestion

PR #2283, opened 2024-09-06
feat: DAH-2556 check translations cache LastModifiedDate

cade-exygy:
## Description

This PR adds methods to compare the LastModifiedDate of the listing to the LastModifiedDate of the translations cache. Current behavior is when the listing modified date is newer than the translation cache, reprocess translations. If the listing modified date is older than the translations, refresh the listing.

Includes refactoring that breaks down existing functionality into smaller methods. 

# DAH-2556

### Version Control

### Code quality

- [x] [if the set of changes cannot be small, reviewers have been forewarned](https://google.github.io/eng-practices/review/developer/small-cls.html#cant)
- [ ] code meets test coverage thresholds
- [x] code is properly formatted
- [x] code is linted
- [x] automated tests pass

### Review instructions

- [ ] instructions can be followed by PA testers
- [x] instructions specify if it can only be followed by an engineer

### Request review

## Review

For engineers locally:
Run `memcached`
Run `yarn start`
Run `bundle exec rake translate:subscribe`

Load a listing detail page -> since a listing will initially have no translations cache, this should trigger the translations to be processed. 

Refresh page -> The listing's translation cache should be valid -> no translations call

Update a field in Salesforce -> Event processor updates cache -> LastModifiedDates should match when the listing cache is refreshed next

Refresh page -> (May refresh listing) -> Should not reprocess translations

Kill the Eventsubscriber process -> Update a field in salesforce -> No translation call

Refresh page -> LastModifiedDates should be different -> translations are reprocessed


app/services/force/listing_service.rb
```diff
@@ -53,7 +54,10 @@ def self.listing(id, opts = {})
       listing
     end
 
-    def self.translations_valid?(listing_translations)
+    def self.translations_valid?(listing_translations, listing_last_modified)
+      return false unless listing_translations[:LastModifiedDate] == listing_last_modified
```
jimlin-sfgov, line 58: question: This seems to duplicate the method `CacheService#listing_unchanged?`, which gets run every 10 minutes via the `prefetch` job. Does this make the `prefetch` job redundant now? I.e. do we still need a background process to update caches for all modified listings every 10 minutes, when we are now doing it on every visit to a listing details page?

app/services/google_translation_service.rb
```diff
@@ -27,8 +27,12 @@ def translate(text, to)
     end
   end
 
-  def cache_listing_translations(listing_id, keys, translations)
-    translations = transform_translations_for_caching(listing_id, keys, translations)
+  def cache_listing_translations(listing_id, keys, translations, last_modified)
+    Force::Request.new(parse_response: true).cached_get(
+      "/ListingDetails/#{CGI.escape(listing_id)}", nil, true
+    )
```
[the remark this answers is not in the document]
cade-exygy: Adding it here would guarantee that the listing is refreshed anytime we run translations for a listing. (Either by event trigger or a full process call). 

This solves the issue of https://sfgovdt.jira.com/browse/DAH-2575, as well as ensures the last modified date of the listing is immediately correct for when the listing service is hit.

Testing through the different cases in the review instructions may provide some clarity to my above open question.

app/services/force/listing_service.rb
cade-exygy: This is a good question and is related to the open question in the description. 

This function compares the listing LastModifiedDate and the translations cache LastModifiedDate. 

This returns false when:
- A listing has been updated but the event subscriber failed to update the translations.

- With the force=true call in the google_translate_service, we should always refresh the listing object when we update translations. So in theory, the LastModifiedDate of the listing should be guaranteed to be accurate for the purposes of the comparison.

If this function is false, then we process translations -> force update the listing -> translate all fields.

So yes, if the event subscriber is working -> then everytime we update a listing -> we translate and force update the listing. We could probably remove the prefetch job in this case.

If the event subscriber is down for some reason, then we would not update translations or force refresh the listing. 
If we disable the prefetch job under this scenario: The listing wouldn't reflect the new changes until a manual force refresh. And we wouldn't catch the disagreement on the listing service call.

We could remove the force=true call from the google translate service, and then this check in the listing detail would return false due to the translations being newer. (Unless the prefetch job gets it first). The intended behavior would be to refresh the listing. Then the dates should match.

This ticket primarily aims to resolve the potential complications of the event subscriber failing. We do plan on implementing event replays so that is a future consideration. 

Hopefully this provides some insight into the potential paths forward. We could keep the prefetch job and introduce some redundancy that would cover all of the potential scenarios.

We could remove the force=true call from the google_translation service, and just let the listing_service be responsible for updating the listing cache. Perhaps that would simplify things for us. 

Maybe we should make the LastModifiedDate comparison have two outcomes: if the listing date is older than the translations -> refresh the listing. If the translation date is older than the listing -> refresh the translations.

app/services/google_translation_service.rb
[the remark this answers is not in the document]
tallulahkay: Based on our convo in the translations sync, can we remove the force update listing call here?

app/services/force/listing_service.rb
```diff
@@ -29,36 +31,70 @@ def self.eligible_listings(filters)
...
+      add_image_urls(results_with_cached_listing_images).first
+    end
 
-      if ::UNLEASH.is_enabled? 'GoogleCloudTranslate'
-        listing_translations = @cache.fetch("/ListingDetails/#{id}/translations") do
-          Rails.logger.info("Nothing in cache for Listing #{id} translations")
-          {}
-        end
+    def self.get_listing_translations(listing)
+      return unless ::UNLEASH.is_enabled? 'GoogleCloudTranslate'
```
tallulahkay, line 51: I think this flag should be outside the function call to avoid confusion

app/services/force/listing_service.rb
```diff
@@ -29,36 +31,70 @@ def self.eligible_listings(filters)
...
       @cache = Rails.cache
-      endpoint = "/ListingDetails/#{CGI.escape(id)}"
+      endpoint = "#{CACHE_KEY_PREFIX}#{CGI.escape(id)}"
       force = opts[:force] || false
       Rails.logger.info("Calling self.listing for #{id} with force: #{force}")
+
       results = Request.new(parse_response: true).cached_get(endpoint, nil, force)
+      listing = process_listing_images(results)
+
+      listing['translations'] = get_listing_translations(listing) || {}
```
tallulahkay, line 41: Unleash feature flag check should wrap this

app/services/force/listing_service.rb
```diff
@@ -29,36 +31,70 @@ def self.eligible_listings(filters)
...
+      )
+        Rails.logger.info("Translations are not valid for #{listing_id}")
+        return CacheService.new.process_translations(listing)
+      end
 
-        listing['translations'] = listing_translations || {}
+      if listing_is_outdated?(listing_translations[:LastModifiedDate],
+                              listing['LastModifiedDate'])
+        Rails.logger.info("Listing is outdated for #{listing_id}")
+        refresh_listing_cache(listing_id)
```
tallulahkay, line 65: I thought we discussed not refreshing the listing cache in the translations flow?

cade-exygy: We removed it from the translation service. We do want to refresh the listing when we know that the listing is outdated.

tallulahkay: That wasn't my understanding. I thought we wanted to avoid updating the listing cache on page load - if we update on page load than the cache is moot.

cade-exygy: I discussed this above: 

> Maybe we should make the LastModifiedDate comparison have two outcomes: if the listing date is older than the translations -> refresh the listing. If the translation date is older than the listing -> refresh the translations.

But it is a fair question. Relates to this ticket as well: https://sfgovdt.jira.com/browse/DAH-2575

We can wait for the 10 minutes: which means english values will be out of date until a listing recache if we update translations via an event. If we have an event miss, then we would have to wait the 10 minutes to resolve those differences as well.

tallulahkay: We discussed in a meeting and decided to move forward with updating the listing cache!

tallulahkay (approved):
This looks good!! Do you think it's possible to get the test coverage up?

PR #2304, opened 2024-09-17
DAH-2722 fix: check for nil before parsing LastModifiedDate

cade-exygy:
## Description

If `Time.parse` receives `nil`, it throws an `ArgumentError`
Check for `nil` before comparing `LastModifiedDate`

## DAH-2722

https://sfgovdt.jira.com/

### Version Control

### Code quality

### Review instructions

### Request review

## Review

Run `memcached`
Run `yarn start`

Once loaded, visit several listing detail pages. Ensure that the page loads.
Check logs to see if the outdated checks are working properly

### DIFF
--- a/app/services/force/listing_service.rb
+++ b/app/services/force/listing_service.rb
@@ -3,6 +3,8 @@
 module Force
   # encapsulate all Salesforce Listing querying functions
   class ListingService
+    class InvalidLastModifiedDateError < StandardError; end
+
     CACHE_KEY_PREFIX = '/ListingDetails/'
     DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
     # get all open listings or specific set of listings by id
@@ -74,15 +76,17 @@ module Force
     end
 
     def self.translations_are_outdated?(cache_last_modified, listing_last_modified)
-      return true if cache_last_modified.nil? || listing_last_modified.nil?
-
       parse_time(cache_last_modified) < parse_time(listing_last_modified)
+    rescue InvalidLastModifiedDateError => e
+      Rails.logger.error("Error checking if translations are outdated: #{e.message}")
+      true
     end
 
     def self.listing_is_outdated?(cache_last_modified, listing_last_modified)
-      return true if cache_last_modified.nil? || listing_last_modified.nil?
-
       parse_time(cache_last_modified) > parse_time(listing_last_modified)
+    rescue InvalidLastModifiedDateError => e
+      Rails.logger.error("Error checking if translations are outdated: #{e.message}")
+      true
     end
 
     def self.fetch_listing_translations_from_cache(listing_id)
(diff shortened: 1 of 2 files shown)


app/services/force/listing_service.rb
```diff
@@ -74,10 +74,14 @@ def self.translations_invalid?(listing_translations)
     end
 
     def self.translations_are_outdated?(cache_last_modified, listing_last_modified)
+      return true if cache_last_modified.nil? || listing_last_modified.nil?
```
tallulahkay, line 77: can these checks happen before going into the functions? 

ie 
```
if cache_last_modified.nil?
  refresh_listing_cache(listing_id)
end

if listing_last_modified.nil? 
  refresh_listing_cache(listing_id)
end

File: app/services/force/listing_service.rb [hunk 2]
```diff
@@ -74,10 +74,14 @@ def self.translations_invalid?(listing_translations)
...
+
       parse_time(cache_last_modified) < parse_time(listing_last_modified)
     end
 
     def self.listing_is_outdated?(cache_last_modified, listing_last_modified)
+      return true if cache_last_modified.nil? || listing_last_modified.nil?
```
tallulahkay, line 83: can these checks happen outside of the functions? What do we actually want to happen if the cache_last_modified is nil or the listing_last_modified is nil?

cade-exygy: We would want to rerun translations so that the LastModifiedDate is populated. If the listing LastModifiedDate is nil, then it would make since to refresh the listing. That would be a rare case, and it would only happen if the listing is undefined for some reason, which should already be handled.

I think we should keep these checks within these functions

tallulahkay:
Not sure if this change is needed - flushing the cache solved this problem in full

cade-exygy:
> Not sure if this change is needed - flushing the cache solved this problem in full

Even if this is an edge case, we should still handle the `nil` case since it did cause the service to error out. If bad test data causes issues like this, imo we still need to handle the case that results in the error.

tallulahkay:
@cade-exygy it's a relic of transitioning from translations not having LastModifiedDate as they are already stored in the cache to expecting them having LastModifiedDate. In the edge case that only existed in full, we only need the one check for if LastModifiedDate on the translations cache is nil

app/services/force/listing_service.rb
tallulahkay: If they stay in the function, then it is confusing. What if the listing_is_outdated function is run and cache_last_modified is nil but listing_last_modified is not nil. Then we are automaticlaly returning that listing_is_outdated is true, but that isn't the case

cade-exygy: I think if any parameter of the comparison is invalid, we should default to returning `true` for invalid. 
I added more explicit error handling to illustrate that. 
This will report the error but not prevent the page from loading. And should ensure that the translations will be reprocessed if needed.

app/services/force/listing_service.rb
```diff
@@ -75,10 +77,16 @@ def self.translations_invalid?(listing_translations)
...
       parse_time(cache_last_modified) < parse_time(listing_last_modified)
+    rescue InvalidLastModifiedDateError => e
+      Rails.logger.error("Error checking if translations are outdated: #{e.message}")
+      true
     end
 
     def self.listing_is_outdated?(cache_last_modified, listing_last_modified)
       parse_time(cache_last_modified) > parse_time(listing_last_modified)
+    rescue InvalidLastModifiedDateError => e
+      Rails.logger.error("Error checking if translations are outdated: #{e.message}")
```
tallulahkay, line 88: Should this be "Error checking if listing is outdated"?

tallulahkay (approved):

PR #2313, opened 2024-09-19
fix: DAH-2724 Updated DOB Validation Checks

chadbrokaw:
## Description

This PR updates the DOB validation on the front and backend to reject any date before 1/1/1900. The PR also makes it so that the DOB fields reject any non-numeric characters.

## Jira ticket

[DAH-2724](https://sfgovdt.jira.com/browse/DAH-2724)

### Version Control

### Code quality

- [ ] if the code changes the UI, it matches the UI design exactly

### Review instructions

### Request review

[DAH-2724]: https://sfgovdt.jira.com/browse/DAH-2724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/pages/account/components/DOBFieldset.tsx
+++ b/app/javascript/pages/account/components/DOBFieldset.tsx
@@ -90,17 +90,12 @@ const validateNumber = (required: boolean, value: string, maxValue: number, erro
 
 const validateAge = (month: string, day: string, year: string) => {
   if (year.length < 4) return "dob:invalid"
-  if (dayjs(`${month}/${day}/${year}`, "M/D/YYYY").valueOf() > dayjs().valueOf())
-    return "dob:invalid"
-  if (
-    dayjs(`${month}/${day}/${year}`, "M/D/YYYY").valueOf() <
-    dayjs().subtract(117, "years").valueOf()
-  )
-    return "dob:invalid"
-  if (
-    dayjs(`${month}/${day}/${year}`, "M/D/YYYY").valueOf() > dayjs().subtract(18, "years").valueOf()
-  )
-    return "dob:age"
+  const enteredDate = dayjs(`${month}/${day}/${year}`, "M/D/YYYY")
+  const minDate = dayjs("1/1/1900", "M/D/YYYY")
+
+  if (enteredDate.valueOf() > dayjs().valueOf()) return "dob:invalid"
+  if (enteredDate.valueOf() < minDate.valueOf()) return "dob:invalid"
+  if (enteredDate.valueOf() > dayjs().subtract(18, "years").valueOf()) return "dob:age"
   return true
 }
 
@@ -166,6 +161,7 @@ const DateField = ({
         },
       }}
       inputProps={{ maxLength: fieldInfo[fieldKey].maxLength }}
+      type="number"
       register={register}
       onChange={onChange}
     />
--- a/app/javascript/pages/account/components/DOBFieldset.tsx
+++ b/app/javascript/pages/account/components/DOBFieldset.tsx
@@ -93,8 +93,13 @@ const validateAge = (month: string, day: string, year: string) => {
   const enteredDate = dayjs(`${month}/${day}/${year}`, "M/D/YYYY")
   const minDate = dayjs("1/1/1900", "M/D/YYYY")
 
-  if (enteredDate.valueOf() > dayjs().valueOf()) return "dob:invalid"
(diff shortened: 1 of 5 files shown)


tallulahkay (commented):
I was able to enter only "e" - wonder if e is considered a number.... 

Screenshot 2024-09-19 at 12 02 47 PM

tallulahkay:
also, reminder to include the test environment in review instructions!

chadbrokaw:
> I was able to enter only "e" - wonder if e is considered a number....
> 
> Screenshot 2024-09-19 at 12 02 47 PM

Good catch! You are right, it looks like the letter e can be used here to represent an exponent. I wonder what your thoughts are on this? If we wanted to prevent this we would need to modify Bloom UIC components

chadbrokaw:
> @chadbrokaw would it be fixed by moving over to the seeds field?

There is not a Seeds field currently. When they build out the Seeds field tho my hope is they will expose things like the onKeyPress callback so that we can block things like this

app/javascript/pages/account/components/DOBFieldset.tsx
```diff
@@ -166,6 +161,7 @@ const DateField = ({
         },
       }}
       inputProps={{ maxLength: fieldInfo[fieldKey].maxLength }}
+      type="number"
```
jimlin-sfgov, line 164: issue: `e`, `-`, `+`, and `.` can be entered. Use [pattern attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern). Also use `aria-label` attribute for a11y.

tallulahkay: I remember discussing the character issue in a meeting and, as this would require an update to the bloom component, I thought we decided not to worry about this right now. We should create a follow-up ticket and request bloom uic to be updated.

+1 to the aria-label

tallulahkay: @chadbrokaw you can pass in a pattern though inputProps

tallulahkay: @jimlin-sfgov did you try implementing this? You can plug a pattern into the inputProps, but it doesn't seem to prevent the key from being entered. I think we should address this in another ticket

+1 to the aria label though

tallulahkay (approved):
Looks good, besides the ability to enter the few symbols. I think we can find a workaround that doesn't involve updating bloom components, but let's address in another ticket

app/javascript/pages/account/components/DOBFieldset.tsx
chadbrokaw: Good catch! Please correct me if I'm misunderstanding, but the pattern attribute does not prevent characters from being entered, so adding that would be redundant with the validation check being run through React Hook Form?

jimlin-sfgov: > did you try implementing this?

No I didn't, thanks for calling me out!

jimlin-sfgov (approved):
LG

app/javascript/pages/account/components/DOBFieldset.tsx
```diff
@@ -90,17 +90,17 @@ const validateNumber = (required: boolean, value: string, maxValue: number, erro
...
-  if (dayjs(`${month}/${day}/${year}`, "M/D/YYYY").valueOf() > dayjs().valueOf())
-    return "dob:invalid"
+  const enteredDate = dayjs(`${month}/${day}/${year}`, "M/D/YYYY")
+  const minDate = dayjs("1/1/1900", "M/D/YYYY")
+
   if (
-    dayjs(`${month}/${day}/${year}`, "M/D/YYYY").valueOf() <
-    dayjs().subtract(117, "years").valueOf()
-  )
+    !enteredDate.isValid() ||
```
chadbrokaw, line 97: @tallulahkay I found that a user can input 19+2 or 19e2 for example and that number would pass through validation. I added this isValid callback to ensure that we block any of those types of inputs. Would you mind taking a look at this?

tallulahkay: Cool!!

tallulahkay (approved):
This works great!! I approve again :)

PR #2323, opened 2024-09-25
feat: DAH-2723 update rtr preference cards

tallulahkay:
## Description

1. Hard codes translations for Right to Return - Hunters View preference cards because human translations won't be available in time for Hunter's view. 
2. Adds English strings for Right to Return - Protrero (those human translations should be back in time).
3. Adds node link for new read more link for right to return (both preferences).
4. Adds both preferences to list of preferences to allow for display.

NOTE: Translations might change. I pushed the english ones up to phrase but not the other languages.
ANOTHER NOTE: diff-coverage is bad because it is just that one link - I don't think this urgently needs to be tested

## Jira ticket

DAH-2723

### Version Control

### Code quality

### Review instructions

### Request review


jimlin-sfgov:
> diff-coverage is bad because it is just that one link - I don't think this urgently needs to be tested

Agree that it's not urgent, but for reference we would want to add an additional test in `ListingDetailsPreferences.test.tsx` that references a new data file with the preference in `.../data/RailsListingPreferences/`, and then update snapshots.

app/javascript/modules/listingDetails/ListingDetailsPreferences.tsx
```diff
@@ -15,6 +15,7 @@ const determineDescription = (
   NRHPDistrict?: string
 ) => {
   if (customPreferenceDescription) {
+    // todo: Use GoogleCloudTranslate here
```
jimlin-sfgov, line 18: question: is there a ticket we can reference in this comment, so we don't forget to implement the todo?

jimlin-sfgov (approved):
LG with one non-blocking question/suggestion

PR #2380, opened 2024-10-26
chore: DAH-2888 Auto-Translation logging

jimlin-sfgov:
## Description

Remove redundant logs and add useful information to existing ones.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-2888

### Version Control

### Code quality

### Review instructions

### Request review

### DIFF
--- a/app/services/force/event_subscriber_translate_service.rb
+++ b/app/services/force/event_subscriber_translate_service.rb
@@ -80,8 +80,8 @@ module Force
     def subscribe_to_listing_updates
       @faye_client.subscribe('/data/Listing__ChangeEvent') do |platform_event|
         logger(
-          'New Salesforce event via /data/Listing__ChangeEvent: ' \
-          "#{platform_event.inspect}",
+          "New Salesforce event via '/data/Listing__ChangeEvent': " \
+          "#{platform_event.to_json}",
         )
         event = parse_event(platform_event)
         translate_and_cache(event)
@@ -91,10 +91,10 @@ module Force
     def translate_and_cache(event)
       translations = translate_event_values(event.listing_id, event.updated_values)
       if translations.empty?
-        logger("No translations for event: #{event.inspect}")
+        logger("No translations for event: #{event.to_json}")
         return []
       end
-      logger("Caching translations for event: #{event.inspect}")
+      logger("Caching translations for event: #{event.to_json}")
 
       @translation_service.cache_listing_translations(
         event.listing_id,
@@ -151,7 +151,7 @@ module Force
     end
 
     def extract_updated_values(changed_fields, event)
-      logger("Extracting updated values: #{changed_fields.inspect}")
+      logger("Extracting updated values: #{changed_fields.to_json}")
 
       listing_field_names_salesforce = ServiceHelper.listing_field_names_salesforce
       filtered_fields = changed_fields.select do |field|
--- a/app/services/force/event_subscriber_translate_service.rb
+++ b/app/services/force/event_subscriber_translate_service.rb
@@ -90,24 +90,17 @@ module Force
 
     def translate_and_cache(event)
       translations = translate_event_values(event.listing_id, event.updated_values)
-      logger(
-        "Event Translations: #{translations.inspect}",
-      )
       if translations.empty?
-        logger(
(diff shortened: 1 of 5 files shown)


jimlin-sfgov:
Still haven't figured out how to properly stub an error in the eventmachine loop, which is why diff coverage is below threshold after modifying logs. I think it's okay to ignore, unless others think otherwise.

app/services/force/event_subscriber_translate_service.rb
```diff
@@ -137,14 +123,19 @@ def process_event_values(listing_id, values)
           text_to_translate.push(listing[domain_key])
         end
       end
-      text_to_translate
+      [text_to_translate, values.keys]
```
cade-exygy, line 126: **question (non-blocking):** : `values.keys` doesn't change within this function, do we need to return it here or can we grab this value from the caller block?

app/services/force/event_subscriber_translate_service.rb
```diff
@@ -137,14 +123,19 @@ def process_event_values(listing_id, values)
...
       languages = %w[ES ZH TL]
-      text_to_translate = process_event_values(listing_id, values)
-      @translation_service.translate(text_to_translate, languages)
+      text_to_translate, keys_of_text = process_event_values(listing_id, values)
+      @translation_service.translate(
+        text: text_to_translate,
+        targets: languages,
+        caller_method: "#{self.class.name}##{__method__}",
+        listing_id:,
+        field_names: keys_of_text,
```
cade-exygy, line 137: **question (non-blocking):** This could just be `values.keys` right?

app/services/force/event_subscriber_translate_service.rb
jimlin-sfgov: Good point, this is redundant code, happy to remove unnecessary return values and arguments

app/services/force/event_subscriber_translate_service.rb
jimlin-sfgov: This is duplicate of the above comment, which I agree with

app/services/google_translation_service.rb
```diff
@@ -14,32 +16,43 @@ def initialize
     @cache = Rails.cache
   end
 
-  def translate(text, to)
+  def translate(text:, field_names:, targets:, caller_method:, listing_id: nil)
```
jimlin-sfgov, line 19: Adding a bunch of args here just for better logs isn't ideal, but I think it's the most quickest way to get useful logging information.

Later when we do some more refactoring, we can do a more thorough cleanup of logging.

tallulahkay: Sure, can you add that as a todo in a comment above the method, that these extra params are only for logging?

app/services/google_translation_service.rb
```diff
@@ -73,9 +88,42 @@ def transform_translations_for_caching(listing_id, keys, translations, last_modi
     return_value
   end
 
-  def google_translation_logger(message, error: false)
+  def log_text_to_translate(caller_method, targets, text, field_names, listing_id)
```
tallulahkay, line 91: Couldn't these two be one method, with a param for the log (that's the only difference?)

tallulahkay: These actually also don't have to be pulled out into methods, bc they are only used once each

app/services/google_translation_service.rb
tallulahkay: On second thought - I don't think we should do this. Just log this information before translate is called. So the pattern of the logs could be:
"Calling translate"
"Attempting translation"
"Translation successful"

tallulahkay (commented):
Just added some suggestions. The logs look good but I don't think it's necessary to do them in this way, as it's a little convoluted.

app/services/google_translation_service.rb
jimlin-sfgov: > Just log this information before translate is called

Do you mean moving the logs to the methods that call `translate`, e.g the `translate_event_values` and `process_translations` methods?

tallulahkay: yes!

tallulahkay (approved):
Thank you!!

PR #2418, opened 2024-11-15
feat: DAH-2566 redirect from create account to sign in

tallulahkay:
## Description

Redirects from create account to sign in after a successful registration. 

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-2566

### Version Control

### Code quality

- [x] if the changes include human translations, follow the [human translations process](https://sfgovdt.jira.com/l/cp/XS1KpvE4)

### Review instructions

### Request review


tallulahkay:
I couldn't figure out that last bit of tests - I asked @chadbrokaw to take a look to see if he could help

app/javascript/pages/account/create-account.tsx
```diff
@@ -163,8 +163,8 @@ const onSubmit = (setError: (name: string, error: ErrorOption) => void) => (data
   }
   createAccount(userData, contactData)
     .then(() => {
-      // TODO: Redirect to Sign In page with modal message
-      console.log("Created an account.")
+      window.sessionStorage.setItem("redirect", "true")
```
cade-exygy, line 166: **question:**  I am curious to hear the reason behind using session storage. 

Is it possible to use the router pattern for handling the redirect?

```
useEffect(() => {
    const path = getPathWithoutLanguagePrefix(router.pathname)
    void getListing(path.split("/")[2]).then((listing: RailsSaleListing) => {
      if (!listing) {
        router.push("/")
      }
      setListing(listing)
    })
  }, [router, router.pathname])
```

jimlin-sfgov: I think session storage was used to mimic the way Angular was doing it? However, I think it's more "standard" to just use a query param, e.g. `window.location.href = '/sign-in?redirect=true'`

In terms of using the `window` object vs the router pattern, I'm not sure which is better. The router pattern looks cleaner, but the router we're using comes from Bloom's ui-components library, which we're trying to move away from. Also, we already use the `window` object in a lot of other places to navigate between pages.

tallulahkay: @jimlin-sfgov nailed it - I did it to mimic angular, and we use the window object all over the code base already.

I also don't think it's great for someone to be able to get there by just entering `/sign-in?redirect=true`. This way, they only see that modal when we want them to.

tallulahkay: Also, window.replace() removes the previous page from the history, so a user can't use the back button to go back to the create account form. Does router.push have a way to do this?

I think a better way to implement this functionality is with react router but we aren't configured to do that

cade-exygy: Okay, thanks for the explanation! I think this will be okay for now. I might look into refactoring it into a custom hook so it is more maintainable as we add redirect logic to other pages.

cade-exygy: I would love to use react router, but we have had some issues in the past with it due to our Rails routing. React Router is great for SPAs. But since we are kinda forced into a non-SPA paradigm custom routers (like NavigationContext) are typically what is recommended I think.

jimlin-sfgov: I also think this is fine as is, but our page navigation system in general needs cleanup. Funny how we need to refactor things before the whole rewrite thing is even completed.

app/javascript/__tests__/pages/create-account.test.tsx
```diff
@@ -100,6 +100,80 @@ describe("<CreateAccount />", () => {
...
+          const newUrl = new URL(href, base)
+          this._href = newUrl.href
+        } catch {
+          this._href = href
+        }
+      },
+      get: function () {
+        return this._href || "http://dahlia.com"
+      },
+    })
```
jimlin-sfgov, line 136: thought: this is a lot of setup that doesn't get re-used. We should extract it out if we need ot use it for future tests.

app/javascript/pages/account/create-account.tsx
```diff
@@ -163,8 +163,8 @@ const onSubmit = (setError: (name: string, error: ErrorOption) => void) => (data
...
+      window.location.replace("/sign-in")
```
jimlin-sfgov, line 167: question: This is sending me to the Angular page (I'm testing the review app), what do I need to do to get redirected to the React page?

jimlin-sfgov (approved):
LGTM

PR #2467, opened 2024-12-23
test: DAH-2084 Implement Github Actions

chadbrokaw:
## Description

This PR adds in Github Actions that check if the pr title is correctly formatted and, if applicable, if the localization files are properly sorted.

To test this PR, please try to commit improperly sorted localization files and try to change the title of this PR to something incorrect.

### Version Control

### Code quality

- [x] if the changes include human translations, follow the [human translations process](https://sfgovdt.jira.com/l/cp/XS1KpvE4)

### Review instructions

### Request review


tallulahkay (approved):
This looks good! I wonder if the translations action would make more sense as a pre-commit hook - is that even possible? what do you think?

chadbrokaw:
> This looks good! I wonder if the translations action would make more sense as a pre-commit hook - is that even possible? what do you think?

I'm pretty sure it's possible! I looked into both and went with the Github Action because I get annoyed when I'm prevented from commiting something in progress that might fail commit checks at the time but will be correct by the time I merge. Do you have a strong preference? I'm good with either spot

tallulahkay:
> > This looks good! I wonder if the translations action would make more sense as a pre-commit hook - is that even possible? what do you think?
> 
> I'm pretty sure it's possible! I looked into both and went with the Github Action because I get annoyed when I'm prevented from commiting something in progress that might fail commit checks at the time but will be correct by the time I merge. Do you have a strong preference? I'm good with either spot

No strong preference so leave as is!

.github/workflows/translations-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          node-version: "16"
+
+      - name: Install dependencies
+        run: npm install --global jsonlint-cli
+
+      - name: Validate and check sorting of translation files
+        run: |
+          for file in app/assets/json/translations/react/*.json; do
+            echo "Checking $file"
+            jsonlint-cli $file
```
jimlin-sfgov, line 37: question: did you check if `jsonlint-cli` sorts things the same way that the utility configured in our`Gruntfile.js` does? I also wonder if the other utilities in our Gruntfile actually get used, like `grunt-json-remove-fields` and `i18next-scanner`

chadbrokaw: `jsonlint-cli` checks to make sure that the JSON is formatted correctly, `jq` is what does the sorting. When I tested jq against grunt translations they both resulted in the same sort, so I think the two work the same.

I also agree on the gruntfile! I think there are definitely some utilities we aren't using. I can create a ticket to remove unused utilities (though part of me wonders if we should just wait to remove those until we are completely moved away from angular?)

jimlin-sfgov (approved):
LGTM

PR #2662, opened 2025-07-14
fix: DAH-3440 Connection Issue Login Redirect Bug

josh-chou:
## Description

Fixes a redirect to the login page when user attempts to view  listings

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-3440

### Version Control

### Code quality

### Review instructions

### Request eng review

### DIFF
--- a/app/javascript/api/apiService.ts
+++ b/app/javascript/api/apiService.ts
@@ -5,7 +5,8 @@ import { setAuthHeaders, getHeaders, isTokenValid } from "../authentication/toke
 
 // Use this function for authenticated calls
 const createAxiosInstance = (): AxiosInstance => {
-  if (!getHeaders()) {
+  const headers = getHeaders()
+  if (!headers || Object.keys(headers).length === 0) {
     throw new Error("Unauthorized")
   }
   if (!isTokenValid()) {


josh-chou:
> 
> I wonder if there is something going on in this function that could also be contributing to this issue? https://github.com/SFDigitalServices/sf-dahlia-web/blob/DAH-3440-login-redirect-issue/app/javascript/authentication/context/UserProvider.tsx#L32
> 
> See the catch block on line 40

So what is happening is that angular code is writing an empty object {} to auth_headers when there is nobody signed in.  Then in the UserProvider code you linked above, deep in getProfile the getHeaers call passes, but fails the isTokenValid check, so a TokenExpired is thrown instead of Unauthorized.

This PR implements a light handed patch to recognize the {} as an Unauthorized error, avoiding the redirect.  There are two other more fundamental problems which contributed to this bug but is not directly addressed:
1. UserProvider should not be performing a redirect because not every page requires a user be logged in.
2. Angular should not be writing {} to the auth_header local storage.  I could not find references to auth_header in the Angular code, and it looks like all auth_header interaction is handled via ng-token-auth.

josh-chou:
> That makes sense, thanks for the breakdown. How do you think we should handle the remaining fundamental problems. Unfortunately I think the bug I described above is still critical behavior. Should we try and address it in this PR? Or should we do that in another urgent PR?

This patch should resolve the case you described with a signed out session by preventing the login redirect altogether.  Your bug also describes an incorrect redirect after login, which is not addressed.

There seems to be several related issues around this area.  I think we should create a separate Jira ticket for each and track them separately.

PR #2882, opened 2026-04-30
fix: DAH-4091 Upload URL per application

cliu02:
## Description

Use updated appId for backend and continue receiving potential applicationNumber for emails from initial I2A pilot. 

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4091

### Version Control

### Code quality

### Code conventions

### Review instructions

### Request eng review

### DIFF
--- a/app/controllers/invite_to_controller.rb
+++ b/app/controllers/invite_to_controller.rb
@@ -15,12 +15,6 @@ class InviteToController < ApplicationController
         url: application['uploadURL'],
       )
     end
-    if decoded_params['applicationNumber'].present? && decoded_params['applicationNumber'] != 'null'
-      application = Force::ShortFormService.get(decoded_params['applicationNumber'])
-      @invite_to_props = @invite_to_props.merge(
-        url: application['uploadURL'],
-      )
-    end
     record_response(decoded_params)
     render 'invite_to'
   end
--- a/app/controllers/invite_to_controller.rb
+++ b/app/controllers/invite_to_controller.rb
@@ -9,8 +9,8 @@ class InviteToController < ApplicationController
     decoded_params ||= params
     @invite_to_props = props(decoded_params)
     # Get file upload URL for application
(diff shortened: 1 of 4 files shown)


app/services/dahlia_backend/message_service.rb
```diff
@@ -117,7 +117,7 @@ def prepare_submission_fields_invite_to_response(application, listing, deadline,
                                                   application_number, app_id, action)
       return nil unless application && listing
 
-      if application_number.blank? && action.present?
+      if action.present?
```
cliu02, line 120: note: the backend no longer sends the application_number

app/controllers/invite_to_controller.rb
```diff
@@ -9,14 +9,8 @@ def index
...
     # Get file upload URL for application
-    if decoded_params['appId'].present? && decoded_params['appId'] != 'null'
-      application = Force::ShortFormService.get(decoded_params['appId'])
-      @invite_to_props = @invite_to_props.merge(
-        url: application['uploadURL'],
-      )
-    end
-    if decoded_params['applicationNumber'].present? && decoded_params['applicationNumber'] != 'null'
-      application = Force::ShortFormService.get(decoded_params['applicationNumber'])
+    if decoded_params['appId'].present? || decoded_params['applicationNumber'].present?
```
cliu02, line 12: note: old I2A emails still send applicationNumber, so it is used as a fallback to appId

app/services/dahlia_backend/message_service.rb
alulabeshue-sfgov: **question:** what about old links? would we still need to rely on application_number

alulabeshue-sfgov (approved):

app/services/dahlia_backend/message_service.rb
cliu02: @alulabeshue-sfgov correct me if I'm wrong, but this check is to determine which payload to send to the backend for recording the response.

PR #2934, opened 2026-06-08
fix: DAH-0000 Update account settings e2e tests

cliu02:
## Description

Gets the account settings buttons using its corresponding input instead of the string. This allows for the different strings between old and new account pages (like "Update" vs. "Save name").

To review, ensure circleCI e2e tests pass with the account layout flag turned on and off. 

### Version Control

### Code quality

### Code conventions

### Review instructions

### Request eng review

### DIFF
--- a/cypress/e2e/accounts/accountSettings.e2e.ts
+++ b/cypress/e2e/accounts/accountSettings.e2e.ts
@@ -33,7 +33,7 @@ describe("Account Settings", () => {
       "/api/v1/account/update",
       userObjectGenerator({ firstName: "Jane", lastName: "Smith" })
     ).as("updateName")
-    cy.get('button[type="submit"]').contains("Update").first().click()
+    cy.get('input[name="firstName"]').closest("form").find('button[type="submit"]').click()
     cy.wait("@updateName").its("response.statusCode").should("eq", 200)
     // Commenting out the below line because the page is not updating with the success message consistently in the e2e test
     // cy.contains("Your changes have been saved.")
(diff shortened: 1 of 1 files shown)


jimlin-sfgov:
thought: although the updated queries are more flexible, I feel that they go against [best practices](https://testing-library.com/docs/queries/about/#priority) laid out by RTL, which I think apply to these specific e2e tests.
> your test should resemble how users interact with your code... Outside of forms, text content is the main way users find elements.

cliu02:
> thought: although the updated queries are more flexible, I feel that they go against [best practices](https://testing-library.com/docs/queries/about/#priority) laid out by RTL, which I think apply to these specific e2e tests.
> 
> > your test should resemble how users interact with your code... Outside of forms, text content is the main way users find elements.

That makes sense @jimlin-sfgov! 

"Outside of forms" - I wonder if the account settings page doesn't apply to this text content recommendation then, since it's a form, so matching to inputs may be more reasonable? If we want to match to text, we could create a followup ticket for after the toggle is flipped on to update these lines to target "Save X" directly. Happy to hear other suggestions or ideas!

jimlin-sfgov:
@cliu02 For forms, users still look at text to identify inputs, from the same link: 
> When navigating through a website form, users find elements using label text.

A follow-up ticket would work. I'm wondering if we are missing other potential follow up tickets, as we work through the redesign?

cliu02:
@jimlin-sfgov If there are specific follow-up tickets besides cleaning up old code, like amending these e2e tests, I think it makes sense to create them as needed. Jira ticket: https://sfgovdt.jira.com/browse/DAH-4165

jimlin-sfgov (approved):
LGTM

PR #2987, opened 2026-07-20
feat: DAH-4237 record the invite-to response on the client to filter email-scanner bot clicks

fwextensions:
## Summary

Email-security scanners (SafeLinks, Proofpoint, etc.) **prefetch every link in an email with a GET**. Our invite-to "Yes / No / Waitlist" links record the applicant's response *on that GET*, so scanner prefetches silently corrupt Salesforce state and send spurious confirmation emails.

Before changing any of that, we need to know how big the problem is. **This PR only measures it.** The page now runs a client-side check for whether a human is actually looking at it — JS ran, the page is genuinely visible, and either a real interaction happened or the page was foregrounded for ~2s — and reports the result to a **log-only** endpoint that records nothing.

Server-side recording on the GET is unchanged in every mode, so turning this on in production carries no behavior risk for applicants or leasing agents. Diffing "GETs that recorded" against "GETs that also logged a human-verified line" tells us, on live traffic, what share of clicks are bots and whether any real humans fail detection.

Gated behind Unleash flag `temp.webapp.inviteToClientRecording`. Flag off = byte-identical to today.

<details>
<summary>Why the detection filters bots, the two modes, and changes</summary>

### Why this distinguishes humans from scanners

Stop treating a bare GET as evidence of a human. The log call fires only after clearing, in order:

1. **JavaScript must run.** The GET renders inert HTML; React fires the call on mount. Most scanners issue a plain GET and never execute page JS, so they never reach it.
2. **The page must be genuinely visible.** We wait for `document.visibilityState === "visible"` plus evidence of a paint (two nested `requestAnimationFrame` callbacks). Headless/sandbox scanners that *do* run JS usually render hidden or backgrounded and fail here. The two-frame trick is a heuristic, not a spec guarantee — that's written into the code comment, and it's why the gates are layered rather than relied on individually.
3. **A human signal must occur.** We fire on the first real interaction (`pointermove` / `pointerdown` / `keydown` / `scroll`) **or** after ~2s of continuous foreground visibility (dwell). A person trips this immediately; an automated fetch that closes on DOM load does not.

Each gate is something a scanner is unlikely to satisfy but a human satisfies without noticing.

### Modes (Unleash flag `temp.webapp.inviteToClientRecording`)

Resolved **server-side** in the controller and passed to React as `clientRecordingMode`. The flag is read as a plain boolean; no variant is consulted.

| Mode | Flag | Server records on GET? | Client behavior |
|------|------|------------------------|-----------------|
| `off` | disabled | **Yes** (legacy) | inert |
| `shadow` | enabled | **Yes** (unchanged) | runs the detection and calls the **log-only** endpoint `POST /api/v1/next-steps/log-human-verified` — records nothing, logs `act` / `trigger` / `elapsedMs` |

There is deliberately **no `on` (client-records) mode**. Recording from the client has to be reworked once #2993 lands, since that PR changes how `record-response` authorizes callers (a signed token rather than trusting the request body). Adding it back is tracked in **DAH-4275**.

There is also deliberately **no `sessionStorage` dedup guard**: the log call writes nothing, so there's nothing to double-record, and a guard would suppress the repeat loads and refreshes we want to see in the logs. Durable duplicate handling belongs in the backend service and is tracked in **DAH-4276**.

### Changes

- `InviteToController#index` resolves `clientRecordingMode` from the flag and passes it to the React page. It still calls `record_response` on the GET in both modes.
- New `useAutoRecordInviteToResponse` hook implements the visibility + paint + interaction-or-dwell gating. It owns its own enable checks (`mode`, `isTest`, `documentsPath`, and a complete unexpired payload) rather than taking a precomputed `enabled` flag.
- New `POST /api/v1/next-steps/log-human-verified` endpoint for the telemetry: logs only, 400 on malformed input, values `.inspect`'d to prevent log forging from an unauthenticated endpoint.
- RSpec + Jest coverage for both modes and every detection gate.

Verified end-to-end on a review app (scanner `curl` vs. real browser click). See [`docs/plans/defer-invite-response-recording.md`](docs/plans/defer-invite-response-recording.md) for design and rollout notes.

### Rollout

`shadow` on staging → `shadow` on production for ~1–2 weeks → review the bot/human ratio → then decide whether to move recording to the client (after #2993) or keep it server-side. Since recording behavior is identical in both modes, the flag is safe to leave on and instantly revertible.

</details>

### DIFF
--- a/app/controllers/api/v1/invite_to_response_controller.rb
+++ b/app/controllers/api/v1/invite_to_response_controller.rb
@@ -32,6 +32,40 @@ class Api::V1::InviteToResponseController < ApiController
     render json: { error: 'Submit response error' }, status: :internal_server_error
   end
 
+  # Shadow-mode endpoint: records nothing to Salesforce or the backend. It only logs
+  # that the client-side human-detection judged this page view to be a real human, so
+  # we can compare against the server-side GET recording during the shadow rollout.
+  # Safe to call repeatedly.
+  def log_human_verified
+    params.expect(record: %i[type deadline appId listingId act trigger elapsedMs])
+    type, deadline, application_id, listing_id, act, trigger, elapsed_ms =
+      params[:record].values_at(:type, :deadline, :appId, :listingId, :act, :trigger,
+                                :elapsedMs)
+
+    # Values come from an unauthenticated public endpoint, so .inspect each one to
+    # neutralize newline/control-character log forging and make nil/blank values legible.
+    Rails.logger.info(
+      'InviteToResponseController#log_human_verified: ' \
+      'human-verified click (shadow, not recorded) ' \
+      "type=#{type.inspect}, " \
+      "listingId=#{listing_id.inspect}, " \
+      "deadline=#{deadline.inspect}, " \
+      "appId=#{application_id.inspect}, " \
+      "act=#{act.inspect}, " \
+      "trigger=#{trigger.inspect}, " \
+      "elapsedMs=#{elapsed_ms.inspect}",
+    )
+    render json: { success: true }, status: :ok
+  rescue ActionController::ParameterMissing => e
+    # Malformed client request, not a server fault - keep it out of 5xx error rates.
+    Rails.logger.warn("Log human-verified click bad request: #{e.message}")
+    render json: { error: 'Log human-verified click bad request' }, status: :bad_request
+  rescue StandardError => e
+    Rails.logger.error("Log human-verified click error: #{e.message}")
+    render json: { error: 'Log human-verified click error' },
+           status: :internal_server_error
+  end
+
   private
 
   def deadline_has_passed?(deadline)
--- a/app/controllers/api/v1/invite_to_response_controller.rb
+++ b/app/controllers/api/v1/invite_to_response_controller.rb
@@ -32,9 +32,10 @@ class Api::V1::InviteToResponseController < ApiController
     render json: { error: 'Submit response error' }, status: :internal_server_error
   end
 
-  # Shadow-mode endpoint: records nothing to Salesforce or the backend. It only logs
-  # that the client-side human-detection judged this page view to be a real human, so
-  # we can compare against the server-side GET recording during the shadow rollout.
+  # Shadow-mode endpoint: records nothing - it never calls sf-dahlia-backend, so no
+  # Salesforce state changes and no applicant email is sent. It only logs that the
+  # client-side human-detection judged this page view to be a real human, so we can
+  # compare against the server-side GET recording during the shadow rollout.
   # Safe to call repeatedly.
   def log_human_verified
     params.expect(record: %i[type deadline appId listingId act trigger elapsedMs])
--- a/app/controllers/invite_to_controller.rb
+++ b/app/controllers/invite_to_controller.rb
@@ -1,5 +1,7 @@
 # Invite to X controller
 class InviteToController < ApplicationController
+  CLIENT_RECORDING_FLAG = 'temp.webapp.inviteToClientRecording'
+
   before_action :ignore_head_requests
 
   def index
@@ -18,7 +20,25 @@ class InviteToController < ApplicationController
         schedulingUrl: application['leaseupAppointmentSchedulingURL'],
       )
     end
-    record_response(decoded_params)
+    # 'on'     - client is the source of truth; skip server-side recording entirely.
+    # 'shadow' - keep recording server-side (unchanged behavior) but let the client run
+    #            its human-detection in parallel and log-only, so we can measure it
+    #            against real traffic before flipping to 'on'.
+    # 'off'    - legacy behavior: record server-side on GET.
+    if client_recording_mode == 'on'
+      # Resolve act/appId the same way props() does, so legacy links carrying
+      # 'response'/'applicationNumber' don't log misleading nils.
+      act = decoded_params['act'] || decoded_params['response']
+      app_id = decoded_params['appId'] || decoded_params['applicationNumber']
+      Rails.logger.info(
+        'InviteToController#index: *NOT* recording server-side, deferring to client ' \
+        "act=#{act.inspect}, " \
+        "appId=#{app_id.inspect}, " \
+        "deadline=#{decoded_params['deadline'].inspect}",
+      )
+    else
+      record_response(decoded_params)
+    end
     render 'invite_to'
   end
 
@@ -57,8 +77,26 @@ class InviteToController < ApplicationController
     {
       assetPaths: static_asset_paths,
       urlParams: url_params,
+      clientRecordingMode: client_recording_mode,
       submitPreviewLinkTokenParam: encode_token(url_params.except(:act, :response)),
     }.compact
+  end
+
+  # Resolves the rollout state of the client-side recording feature from the
+  # 'temp.webapp.inviteToClientRecording' Unleash flag:
+  #   flag disabled                       -> 'off'    (legacy server-side recording)
+  #   flag enabled, variant 'shadow'      -> 'shadow' (server records; client detects + logs)
+  #   flag enabled, any other/no variant  -> 'on'     (client records; server does not)
+  def client_recording_mode
+    return @client_recording_mode if defined?(@client_recording_mode)
+
+    @client_recording_mode =
+      if Rails.configuration.unleash.is_enabled?(CLIENT_RECORDING_FLAG)
+        variant = Rails.configuration.unleash.get_variant(CLIENT_RECORDING_FLAG)
+        variant&.name == 'shadow' ? 'shadow' : 'on'
+      else
+        'off'
+      end
   end
 
   def record_response(decoded_params)
--- a/app/controllers/invite_to_controller.rb
+++ b/app/controllers/invite_to_controller.rb
@@ -20,25 +20,12 @@ class InviteToController < ApplicationController
         schedulingUrl: application['leaseupAppointmentSchedulingURL'],
       )
     end
-    # 'on'     - client is the source of truth; skip server-side recording entirely.
-    # 'shadow' - keep recording server-side (unchanged behavior) but let the client run
-    #            its human-detection in parallel and log-only, so we can measure it
-    #            against real traffic before flipping to 'on'.
-    # 'off'    - legacy behavior: record server-side on GET.
-    if client_recording_mode == 'on'
-      # Resolve act/appId the same way props() does, so legacy links carrying
-      # 'response'/'applicationNumber' don't log misleading nils.
-      act = decoded_params['act'] || decoded_params['response']
-      app_id = decoded_params['appId'] || decoded_params['applicationNumber']
-      Rails.logger.info(
-        'InviteToController#index: *NOT* recording server-side, deferring to client ' \
-        "act=#{act.inspect}, " \
-        "appId=#{app_id.inspect}, " \
-        "deadline=#{decoded_params['deadline'].inspect}",
-      )
-    else
-      record_response(decoded_params)
-    end
+    # Recording always happens server-side on GET for now. When the flag is enabled the
+    # client additionally runs its human-detection in parallel and logs only ('shadow'),
+    # so we can measure bot vs. human clicks against real traffic. Moving the recording
+    # to the client is deferred until
+    # https://github.com/SFDigitalServices/sf-dahlia-web/pull/2993 lands.
+    record_response(decoded_params)
     render 'invite_to'
   end
 
@@ -84,19 +71,14 @@ class InviteToController < ApplicationController
 
   # Resolves the rollout state of the client-side recording feature from the
   # 'temp.webapp.inviteToClientRecording' Unleash flag:
-  #   flag disabled                       -> 'off'    (legacy server-side recording)
-  #   flag enabled, variant 'shadow'      -> 'shadow' (server records; client detects + logs)
-  #   flag enabled, any other/no variant  -> 'on'     (client records; server does not)
+  #   flag disabled -> 'off'    (client hook inert)
+  #   flag enabled  -> 'shadow' (server records on GET; client detects humans, logs only)
+  # There is deliberately no 'on' (client-records) state yet - see PR #2993.
   def client_recording_mode
     return @client_recording_mode if defined?(@client_recording_mode)
 
     @client_recording_mode =
-      if Rails.configuration.unleash.is_enabled?(CLIENT_RECORDING_FLAG)
-        variant = Rails.configuration.unleash.get_variant(CLIENT_RECORDING_FLAG)
-        variant&.name == 'shadow' ? 'shadow' : 'on'
-      else
-        'off'
-      end
+      Rails.configuration.unleash.is_enabled?(CLIENT_RECORDING_FLAG) ? 'shadow' : 'off'
   end
 
   def record_response(decoded_params)
--- a/app/javascript/hooks/useAutoRecordInviteToResponse.ts
+++ b/app/javascript/hooks/useAutoRecordInviteToResponse.ts
@@ -1,0 +1,213 @@
+import { useEffect } from "react"
+import { recordResponse, logHumanVerifiedClick } from "../api/inviteToApiService"
+import { isDeadlinePassed } from "../util/listingUtil"
+
+const INTERACTION_EVENTS = [
+  "pointermove",
+  "pointerdown",
+  "touchstart",
+  "keydown",
+  "scroll",
+] as const
+
+export type ClientRecordingMode = "off" | "shadow" | "on"
+
+interface UseAutoRecordInviteToResponseArgs {
+  enabled: boolean
+  // "on" fires the real recordResponse; "shadow" fires a log-only call (server still records on GET).
+  mode?: ClientRecordingMode
+  act?: "yes" | "no" | "contact" | "submit" | "appointment"
+  appId?: string
+  listingId?: string
+  deadline?: string
+  type?: string
+  dwellMs?: number
+}
+
+const AUTO_RECORD_ACTS = new Set(["yes", "no", "contact"])
+
+// Shadow and on modes use distinct guard keys so a session that later flips from shadow to on
+// still records for real (the shadow call must not block the real one).
+const guardKey = (appId: string, act: string, mode: ClientRecordingMode) =>
+  mode === "shadow" ? `i2x-shadow-${appId}-${act}` : `i2x-recorded-${appId}-${act}`
+
+const isGuardSet = (key: string) => {
+  try {
+    return sessionStorage.getItem(key) === "true"
+  } catch {
+    return false
+  }
+}
+
+const setGuard = (key: string) => {
+  try {
+    sessionStorage.setItem(key, "true")
+  } catch {
+    // Ignore private-mode / storage-disabled failures; server-side dedup is the backstop.
+  }
+}
+
+/**
+ * Automatically records an applicant's response to an invite-to-apply/interview email link
+ * once the page has genuinely been seen by a human, gating out email-scanner-bot prefetches.
+ *
+ * In `mode: "on"` it fires the real `recordResponse`; in `mode: "shadow"` it fires a log-only
+ * `logHumanVerifiedClick` (the server still records on GET) so the detection can be measured
+ * against live traffic. Fires exactly once when all of the following hold:
+ *  - `enabled` is true (caller is responsible for computing this, including the
+ *    client-recording mode, `act`, `isTest`, `documentsPath`, and deadline checks)
+ *  - the document is visible
+ *  - two animation frames have elapsed while visible (a real paint happened)
+ *  - either a first user interaction occurs, or `dwellMs` of continuous visibility elapses
+ *    (whichever comes first)
+ */
+export const useAutoRecordInviteToResponse = ({
+  enabled,
+  mode = "on",
+  act,
+  appId,
+  listingId,
+  deadline,
+  type,
+  dwellMs = 2000,
+}: UseAutoRecordInviteToResponseArgs) => {
+  useEffect(() => {
+    if (
+      !enabled ||
+      mode === "off" ||
+      !act ||
+      !AUTO_RECORD_ACTS.has(act) ||
+      !appId ||
+      !listingId ||
+      !type ||
+      !deadline ||
+      isDeadlinePassed(deadline)
+    ) {
+      return undefined
+    }
+
+    const key = guardKey(appId, act, mode)
+    if (isGuardSet(key)) {
+      return undefined
+    }
+
+    // Set when the fire conditions actually arm (after the render gate, while visible), so
+    // elapsedMs reflects visible dwell/interaction time and excludes hidden/background-tab time.
+    let armedAt = 0
+    let fired = false
+    let dwellTimeoutId: ReturnType<typeof setTimeout> | null = null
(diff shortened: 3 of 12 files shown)


app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,202 @@
...
+      !enabled ||
+      mode === "off" ||
+      !act ||
+      !AUTO_RECORD_ACTS.has(act) ||
+      !appId ||
+      !listingId ||
+      isDeadlinePassed(deadline)
+    ) {
+      return undefined
+    }
```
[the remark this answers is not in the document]
fwextensions: Good catch on guarding `type` — added `!!type` to the enable gate (both here in the hook and in `invite-to.tsx`), so a next-steps load missing `type` can no longer POST a payload that `params.expect` would reject.

On the throw concern: `isDeadlinePassed` delegates to `dayjs.tz(deadline, ...)`, which treats `undefined` as "now" and returns `false` rather than throwing — so this was not actually a runtime crash. Still, relying on that is fragile, so I also added `!!deadline` to the gate to make the requirement explicit. Fixed in f21c7e168.

app/javascript/pages/inviteTo/invite-to.tsx
```diff
@@ -42,10 +47,33 @@ const InviteToPage = ({
...
+      !isDeadlinePassed(deadline) &&
+      !!appId &&
+      !!listingId,
+    mode: clientRecordingMode,
+    act,
+    appId,
+    listingId,
+    deadline,
+    type,
+  })
```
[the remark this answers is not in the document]
fwextensions: Same fix applied here in `invite-to.tsx`: the enable gate now includes `!!type && !!deadline` alongside the existing `!isDeadlinePassed(deadline)` check, so the hook only arms with a complete, valid payload. (Note `isDeadlinePassed(undefined)` returns `false` via dayjs rather than throwing, but guarding `deadline` explicitly is clearer.) Fixed in f21c7e168.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -32,6 +32,34 @@ def record_response
...
+      "act=#{act}, " \
+      "trigger=#{trigger}, " \
+      "elapsedMs=#{elapsed_ms}",
+    )
+    render json: { success: true }, status: :ok
+  rescue StandardError => e
+    Rails.logger.error("Log human-verified click error: #{e.message}")
+    render json: { error: 'Log human-verified click error' },
+           status: :internal_server_error
+  end
```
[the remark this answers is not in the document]
fwextensions: Agreed. `log_human_verified` now rescues `ActionController::ParameterMissing` separately and returns `400 :bad_request` (logged at `warn`), so malformed client requests stay out of the 5xx rate; the generic `StandardError` → 500 rescue remains for genuine server faults. Fixed in f21c7e168.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -37,34 +37,76 @@
...
+          'elapsedMs=1234',
+        ),
+      )
+    end
+
+    it 'returns an error when the record param is missing' do
+      post :log_human_verified, params: { notRecord: {} }
+      expect(response).not_to be_ok
+      expect(DahliaBackend::MessageService).not_to have_received(:send_invite_to_response)
+    end
```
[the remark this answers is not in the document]
fwextensions: Updated: the spec now asserts `have_http_status(:bad_request)` explicitly (was `not_to be_ok`), so a regression back to 500 would be caught. Fixed in f21c7e168.

docs/plans/defer-invite-response-recording.md
```diff
@@ -0,0 +1,196 @@
...
+- New boolean prop `clientRecording` added to `@invite_to_props` (top level, sibling of
+  `urlParams`), `true` iff
+  `Rails.configuration.unleash.is_enabled?('temp.webapp.inviteToClientRecording')`.
+- When `clientRecording` is true, `InviteToController#index` **does not** call `record_response`
+  server-side; the React page is responsible for POSTing.
+- The client POST reuses the existing `recordResponse()` payload shape (see current usage in
+  `InviteToApplyNextSteps.tsx:71` / `InviteToInterviewNextSteps.tsx:45`):
+  `{ appId, applicationNumber: appId, listingId, deadline, action: act, response: act, type }`.
+  (`action` and `response` carry the same value; `applicationNumber` mirrors `appId` — matches
+  existing on-page button behavior and what the API controller expects.)
```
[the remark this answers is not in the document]
fwextensions: Fixed — the Contract section now describes the `clientRecordingMode` (`off`/`shadow`/`on`) prop and its semantics, consistent with the Rollout modes section and the implementation. Updated in f21c7e168.

docs/plans/defer-invite-response-recording.md
```diff
@@ -0,0 +1,196 @@
...
+- `act === "submit"` and `act === "appointment"` never arrive via email links that need recording;
+  only `yes`, `no`, `contact` should auto-record. Gate on those three explicitly.
+- Import `isDeadlinePassed` from `app/javascript/util/listingUtil`.
+
+### Wiring in `invite-to.tsx`
+
+- Accept the new `clientRecording?: boolean` prop in `HomePageProps`.
+- Derive `listingId` the same way the existing `getListing` effect does
+  (`getPathWithoutLanguagePrefix(pathname).split("/")[2]`) so the hook doesn't wait on the listing
+  fetch.
```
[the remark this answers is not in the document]
fwextensions: Fixed — the Wiring section now says `clientRecordingMode?: "off" | "shadow" | "on"` (defaulting to `"off"`) and notes it is passed to the hook as `mode`. The hook-signature snippet above it was updated to match too. Updated in f21c7e168.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -32,6 +32,38 @@ def record_response
...
+      'InviteToResponseController#log_human_verified: ' \
+      'human-verified click (shadow, not recorded) ' \
+      "type=#{type}, " \
+      "listingId=#{listing_id}, " \
+      "deadline=#{deadline}, " \
+      "appId=#{application_id}, " \
+      "act=#{act}, " \
+      "trigger=#{trigger}, " \
+      "elapsedMs=#{elapsed_ms}",
+    )
```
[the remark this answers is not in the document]
fwextensions: Done — each interpolated value in `log_human_verified` is now `.inspect`'d, which neutralizes newline/control-char log forging from this unauthenticated endpoint and makes nil/blank values legible. Added a comment explaining why. Fixed in 1f95da58a.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,210 @@
+/* eslint-disable @typescript-eslint/no-use-before-define -- the fire/cleanup/listener
+   closures below reference each other circularly; there is no acyclic definition order */
```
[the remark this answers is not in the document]
fwextensions: Done — removed the file-wide `eslint-disable` and replaced it with two targeted `eslint-disable-next-line` comments on the only two genuine forward references (`handleInteraction` in `removeInteractionListeners`, `handleVisibilityChange` in `cleanup`). The rule now stays enabled for the rest of the file; `eslint` passes clean. Fixed in 1f95da58a.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/invite_to_controller.rb
```diff
@@ -18,7 +20,21 @@ def index
...
+    #            its human-detection in parallel and log-only, so we can measure it
+    #            against real traffic before flipping to 'on'.
+    # 'off'    - legacy behavior: record server-side on GET.
+    if client_recording_mode == 'on'
+      Rails.logger.info(
+        'InviteToController#index: *NOT* recording server-side, deferring to client ' \
+        "act=#{decoded_params['act'].inspect}, " \
+        "appId=#{decoded_params['appId'].inspect}, " \
+        "deadline=#{decoded_params['deadline'].inspect}",
+      )
```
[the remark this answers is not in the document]
fwextensions: Fixed — the `on`-mode log now resolves `act`/`appId` with the same `|| response` / `|| applicationNumber` fallback that `props()` uses, so legacy links no longer log misleading nils. Fixed in ee504687f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,210 @@
...
+    ) {
+      return undefined
+    }
+
+    const key = guardKey(appId, act, mode)
+    if (isGuardSet(key)) {
+      return undefined
+    }
+
+    const armedAt = Date.now()
```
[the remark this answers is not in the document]
fwextensions: Fixed — `armedAt` is no longer set at effect start; it is now reset inside `armFireConditions` (after the render gate, while visible), so `elapsedMs` reflects only visible dwell/interaction time and excludes hidden/background-tab time. See the companion change at line ~171. Fixed in ee504687f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    const armFireConditions = () => {
+      INTERACTION_EVENTS.forEach((eventName) => {
+        window.addEventListener(eventName, handleInteraction, { passive: true, once: true })
+      })
+      dwellTimeoutId = setTimeout(() => {
+        fire("dwell")
+      }, dwellMs)
+    }
```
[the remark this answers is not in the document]
fwextensions: Done — `armFireConditions` now sets `armedAt = Date.now()` when it arms (post render-gate, visible), so hidden/background time no longer inflates the logged `elapsedMs`. Fixed in ee504687f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

fwextensions:
**Interaction with #2993 (`DAH-4053-secure-record-response-api`)**

#2993 adds `before_action :validate_token!, only: :record_response` to `Api::V1::InviteToResponseController`, requiring a validly-signed JWT in `params[:t]` (via the new `JsonWebTokenService`). This directly affects this branch:

- **Breaks client-side recording.** Every client `POST /api/v1/next-steps/record-response` here sends only `{ record }` and no token (`inviteToApiService.ts`), so after #2993 merges they will return **401 Unauthorized** and record nothing.
- **Two callers affected, not one:**
  1. The new deferred/auto-record hook (`useAutoRecordInviteToResponse`) in `mode: "on"`.
  2. The pre-existing "Submit Your Info" upload button (`InviteToApplyNextSteps.tsx` `handleSubmitClick`, and the I2I equivalent), which already POSTs `record-response` on click.
- **Unaffected:** `shadow` mode (hits `log-human-verified`, which #2993 does not guard) and legacy `off` mode (records server-side in `InviteToController#index`).

**Token availability:** the JWT `t` arrives in the email-link URL but is consumed server-side in `InviteToController#index`; only decoded fields reach React as props. The raw `t` is not currently plumbed to the client, and `submitPreviewLinkTokenParam` is a *different* token (re-encoded with `.except(:act, :response)`) intended for the preview link.

app/javascript/pages/inviteTo/invite-to.tsx
```diff
@@ -42,10 +47,35 @@ const InviteToPage = ({
...
+      !!act &&
+      act !== "submit" &&
+      act !== "appointment" &&
+      !isTestMode &&
+      !documentsPath &&
+      !!type &&
+      !!deadline &&
+      !isDeadlinePassed(deadline) &&
+      !!appId &&
+      !!listingId,
```
alulabeshue-sfgov: **question:** can we just check for clientRecordingMode, do we need this enabled param?

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
...
+            })
+          : recordResponse({
+              appId,
+              applicationNumber: appId,
+              listingId,
+              deadline,
+              action: act,
+              response: act,
+              type,
+            })
```
alulabeshue-sfgov: **note:** as discussed we can remove the calling of recordResponse since it will need to be updated after https://github.com/SFDigitalServices/sf-dahlia-web/pull/2993

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+}
+
+const setGuard = (key: string) => {
+  try {
+    sessionStorage.setItem(key, "true")
+  } catch {
+    // Ignore private-mode / storage-disabled failures; server-side dedup is the backstop.
+  }
+}
```
alulabeshue-sfgov: **note:** we discussed removing this caching strategy, we can create a follow up ticket to better handle duplicate clicks. Today it lives in the backend service.

alulabeshue-sfgov (commented):
Left some comments that we discussed. Also can you create a follow up ticket to clean up the `temp.webapp.inviteToClientRecording` toggle. When creating toggles we usually create clean up tickets to remind ourselves in the future.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -32,6 +32,40 @@ def record_response
     render json: { error: 'Submit response error' }, status: :internal_server_error
   end
 
+  # Shadow-mode endpoint: records nothing to Salesforce or the backend. It only logs
```
jimlin-sfgov: nit: Salesforce _is_ our backend, why say "or the backend."?

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
+import { useEffect } from "react"
+import { recordResponse, logHumanVerifiedClick } from "../api/inviteToApiService"
+import { isDeadlinePassed } from "../util/listingUtil"
+
+const INTERACTION_EVENTS = [
+  "pointermove",
+  "pointerdown",
+  "touchstart",
```
jimlin-sfgov: question: why use the "touchstart" event? it seems like pointer events have become the standard.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
...
+interface UseAutoRecordInviteToResponseArgs {
+  enabled: boolean
+  // "on" fires the real recordResponse; "shadow" fires a log-only call (server still records on GET).
+  mode?: ClientRecordingMode
+  act?: "yes" | "no" | "contact" | "submit" | "appointment"
+  appId?: string
+  listingId?: string
+  deadline?: string
+  type?: string
+  dwellMs?: number
```
jimlin-sfgov: suggestion: rename to `dwellTimeMs`, it's easier to look up for folks who are unfamiliar with the term.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
...
+ * Automatically records an applicant's response to an invite-to-apply/interview email link
+ * once the page has genuinely been seen by a human, gating out email-scanner-bot prefetches.
+ *
+ * In `mode: "on"` it fires the real `recordResponse`; in `mode: "shadow"` it fires a log-only
+ * `logHumanVerifiedClick` (the server still records on GET) so the detection can be measured
+ * against live traffic. Fires exactly once when all of the following hold:
+ *  - `enabled` is true (caller is responsible for computing this, including the
+ *    client-recording mode, `act`, `isTest`, `documentsPath`, and deadline checks)
+ *  - the document is visible
+ *  - two animation frames have elapsed while visible (a real paint happened)
```
jimlin-sfgov: question: why is two the magic number? can you provide a source?

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -0,0 +1,213 @@
...
+      armedAt = Date.now()
+      INTERACTION_EVENTS.forEach((eventName) => {
+        window.addEventListener(eventName, handleInteraction, { passive: true, once: true })
+      })
+      dwellTimeoutId = setTimeout(() => {
+        fire("dwell")
+      }, dwellMs)
+    }
+
+    const startRenderGate = () => {
```
jimlin-sfgov: suggestion: figure out a better function name. To me, this reads like a function is gating the render. From the code it looks like animation frames are used to trigger some actions.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: Done in 82b3af7e8. The hook only calls `logHumanVerifiedClick` now — `recordResponse` isn't imported at all, and `ClientRecordingMode` is down to `"off" | "shadow"`, so there's no code path where the client records. The controller records on GET in both modes, which means turning the flag on in production changes nothing except adding the log line.

I left a note in the hook and in the plan doc pointing at #2993 as the blocker for adding the `on` mode back.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: Done in 82b3af7e8 — the `sessionStorage` guard and its key helpers are gone. Since the endpoint writes nothing there's nothing to double-record, and dropping the guard is actually better for measurement: repeat loads and refreshes now show up in the logs instead of being silently suppressed.

Filed DAH-4276 for the follow-up on duplicate-click handling in the backend service (replacing the latest-Field-Update-Comment heuristic in `OutreachService.checkForDuplicateClick` with a durable idempotency check). It's in Sprint 273 and linked to DAH-4237.

app/javascript/pages/inviteTo/invite-to.tsx
fwextensions: You were right, and it got worse after I removed the `on` mode — the page was computing a ten-condition `enabled` expression that the hook then re-checked line for line.

Fixed in 983232806: `enabled` is gone. The hook takes `mode`, `isTest` and `documentsPath` and owns all the validity checks itself, so the call site is just the values:

```ts
useAutoRecordInviteToResponse({
  mode: clientRecordingMode,
  act, appId, listingId, deadline, type,
  isTest: isTestMode,
  documentsPath,
})
```

One thing I couldn't remove: a hook can't be called conditionally, so the decision has to arrive as an argument rather than as an `if` around the call. But it's now one prop (`mode`) plus the two page states the hook can't derive on its own, instead of a gate duplicated in two places.

app/controllers/api/v1/invite_to_response_controller.rb
fwextensions: Fair — that was redundant. Salesforce sits behind sf-dahlia-backend, so the comment now names the call we're skipping instead of listing both:

```ruby
# Shadow-mode endpoint: records nothing - it never calls sf-dahlia-backend, so no
# Salesforce state changes and no applicant email is sent.
```

Fixed in 983232806.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: No good reason — removed in 983232806. `pointerdown`/`pointermove` fire for touch input on everything in our browserslist, so `touchstart` was pure redundancy. The list is now just `pointermove`, `pointerdown`, `keydown`, `scroll`, with a comment noting that pointer events cover touch so nobody adds it back.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: Renamed to `dwellTimeMs` in 983232806, along with the matching doc comment. Agreed that "dwell" on its own isn't a term you can look up.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: Agreed, that name was backwards — it doesn't gate rendering, it waits on it. Renamed to `afterFirstPaint` in 983232806, and `clearRenderGateFrames` became `cancelPendingFrames` since all it does is cancel the outstanding rAF handles.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: There isn't a spec citation for "two frames," so let me give you the actual reasoning rather than pretend there is one.

A `requestAnimationFrame` callback runs *before* the paint of the frame it was scheduled for, so one frame proves nothing was drawn. Scheduling a second frame from inside the first means we resume after that first frame was handed off for rendering. The relevant ordering is in the HTML spec's "update the rendering" steps: https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering

What the spec does *not* promise is a commit to the screen between the two callbacks, so this is evidence of a paint, not proof of one. I've written that into the comment in 983232806 rather than implying it's guaranteed.

I'm comfortable with it being possibly a frame optimistic because it's one layer of three: the visibility check and the interaction-or-dwell gate both still have to pass before anything is logged. And if the shadow data shows scanners clearing this gate, that's exactly the signal that tells us to replace it with something stronger.

fwextensions:
## Follow-up tickets + scope change

@alulabeshue-sfgov good call on the cleanup-ticket convention — filed both, in Sprint 273 and linked to DAH-4237:

- **DAH-4275** — remove the `temp.webapp.inviteToClientRecording` toggle once we've reviewed the shadow measurement and decided whether recording moves to the client.
- **DAH-4276** — durable duplicate-click detection in the backend service, replacing the latest-Field-Update-Comment heuristic in `OutreachService.checkForDuplicateClick`. This is the follow-up from removing the `sessionStorage` guard.

### Scope has narrowed since the last review

Per our discussion, this PR no longer changes how responses are recorded. Two things came out:

1. **No client-side recording.** `recordResponse` is gone from the hook and the `on` mode is gone entirely, since recording from the client needs to be reworked after #2993 changes how the API authorizes the call. `clientRecordingMode` is now just `off | shadow`, resolved from the flag as a plain boolean — no variant is read.
2. **No `sessionStorage` guard.** The log-only endpoint writes nothing, so there's nothing to double-record, and a guard would hide the repeat loads and refreshes we want visible in the logs.

`InviteToController#index` records server-side on the GET in both modes, so **enabling the flag in production changes nothing for applicants or agents** — it only adds the human-detection log line. That's the whole point of shipping it this way: we get the measurement in production with no behavior risk, and decide about moving the recording afterwards.

The description above has been updated to match. Also picked up @jimlin-sfgov's naming and event-list feedback in 983232806.

fwextensions:
Added Opus replies to comments above.

jimlin-sfgov:
I'm satisfied with revisions made from my feedback. I'll let @alulabeshue-sfgov follow the complex review instructions and give the approval.

alulabeshue-sfgov (commented):
Code looks good to me, comments are resolved. Just need to fix some failing e2e tests and I think we can ship

alulabeshue-sfgov (approved):

PR #2988, opened 2026-07-20
chore: DAH-3138 Upgrade to React 19 on webapp

fwextensions:
## Description

Upgrades the webapp to React 19, bumping `@bloom-housing/ui-components` to 14.1.0 (now on React 19) and the related React/testing/type dependencies. `react-hook-form` intentionally stays at 6.15.5 to match ui-components' bundled copy.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-3138

## What changed

**Dependencies** (`package.json` + `resolutions`)
- `react` / `react-dom` 18 → 19.2 (single hoisted copy)
- `@bloom-housing/ui-components` ^12.7.6 → ^14.1.0
- `@types/react` / `@types/react-dom` 18 → 19
- `@testing-library/react` 14 → 16, `@testing-library/dom` 9 → 10
- `react-helmet-async` 1 → 3
- `@fortawesome/react-fontawesome` 0.2.x → 3.x (dependency + resolution)

**Source changes for React 19 behavior**
- New `app/javascript/types/jsx-global.d.ts`: restores the global `JSX` namespace (React 19 moved it to `React.JSX`) and re-adds the removed `ReactChild`/`Props` types that `markdown-to-jsx`'s vendored types still reference.
- Guard empty `imageSrc` (`imageSrc || undefined`) in the listing-detail components — React 19 warns on ``.
- `PasswordFieldset` now derives the password-strength value directly from `watch()` instead of mirroring it into state via `useEffect` (that effect lagged one keystroke under React 19).
- Type a `ReactElement.props` read that React 19 now types as `unknown`.

**Test infra (jsdom)**
- `ResizeObserver` + `MessageChannel` polyfills in `jest.setup.js` (needed by `@dnd-kit` and `react-dom/server` under React 19).
- Map `@preact/signals-core` to its CJS build (jsdom otherwise resolves the ESM `browser` export).
- Refreshed snapshots for benign markup changes: native `<title>`/`<meta>` hoisting to `<head>`, FontAwesome class formatting, and the React 19 `useId` format.

## Verification

- Full Jest suite green: 875 passed / 136 suites
- `tsc` clean, `eslint` clean

## Review instructions

Applies to all environments.

- [ ] Smoke-test key pages in the review environment: listings directory, listing detail (rental + sale), account settings, sign in / create account.
- [ ] Confirm client-side hydration works on the react-on-rails render path (no console errors).

> Not yet run in this branch and worth confirming before merge: Storybook build and the Cypress e2e suite.

## Notes

- Split from DAH-3138; the partners upgrade is tracked separately in DAH-3704.
- Listing-detail pages render noticeably slower under React 19 in jsdom (hence the raised Jest timeout). Likely just the heavier v19 + ui-components 14 tree, but flagging in case CI time needs attention.


app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
```diff
@@ -96,7 +96,10 @@ const ListingApplyStepWrapper = ({
         )}
         <Form onSubmit={formMethods.handleSubmit(onSubmit)}>
           {Children.map(children, (child) => {
-            const { schema } = (child as React.ReactElement).props
+            // React 19 types `ReactElement.props` as `unknown`; describe the shape we read.
+            const { schema } = (
+              child as React.ReactElement<{ schema?: { props?: { divider?: boolean } } }>
+            ).props
             return (
```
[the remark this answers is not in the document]
fwextensions: Good catch, fixed in da224267a. Switched to `React.isValidElement` to guard the `.props` read instead of the unsafe cast.

app/javascript/__tests__/jest.setup.js
```diff
@@ -3,3 +3,43 @@
...
+    unobserve() {}
+    disconnect() {}
+  }
+
+// React 19's react-dom/server (imported transitively by ui-components) relies on
+// MessageChannel for scheduling, which jsdom does not provide. Node's worker_threads
+// implementation keeps the event loop alive and hangs Jest on exit, so use a minimal
+// setTimeout-based polyfill that satisfies React's scheduler (port2.postMessage
+// dispatches to port1.onmessage) without holding open handles.
+if (global.MessageChannel === undefined) {
```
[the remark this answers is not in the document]
fwextensions: Verified empirically: in this repo's jsdom test environment, `typeof MessageChannel` is `undefined` before jest.setup.js runs (jest-environment-jsdom's global scope doesn't inherit Node's MessageChannel the way it does e.g. TextEncoder). So the `if (global.MessageChannel === undefined)` guard does take effect and our polyfill is the one installed for every test run. Left the guard in place (rather than always overriding) so the polyfill stays a no-op if a future jsdom/jest version does start exposing a native MessageChannel that works correctly under jsdom.

app/javascript/modules/listingDetails/ListingDetailsNeighborhood.tsx
```diff
@@ -78,7 +78,7 @@ export const ListingDetailsNeighborhood = ({
     <ListingDetailItem
       desktopClass="bg-primary-lighter"
       imageAlt={""}
-      imageSrc={imageSrc}
+      imageSrc={imageSrc || undefined}
```
jimlin-sfgov: Question: It's not obvious to me that we're guarding against empty strings, and the prop's type is `string`. An `undefined` src is also bad? Is there a better way to guard this?

app/javascript/pages/account/components/PasswordFieldset.tsx
```diff
@@ -173,15 +173,12 @@ const PasswordFieldset = ({
   labelText,
   email,
 }: PasswordFieldsetProps) => {
-  const [passwordValidationContent, setPasswordValidationContent] = React.useState("")
-  const newPassword: string = watch("password", "")
+  // Derive the validation content directly from `watch` during render. Mirroring it into
+  // state via an effect lags one keystroke behind under React 19's rendering/effect timing.
+  const passwordValidationContent: string = watch("password", "")
```
jimlin-sfgov: suggestion: Use a simpler name like `passwordField` or `passwordValue`.

app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
```diff
@@ -96,7 +96,15 @@ const ListingApplyStepWrapper = ({
         )}
         <Form onSubmit={formMethods.handleSubmit(onSubmit)}>
           {Children.map(children, (child) => {
-            const { schema } = (child as React.ReactElement).props
+            // `children` may include non-element values (e.g. `false` from a conditional
```
jimlin-sfgov: suggestion: Remove or reword this comment, we enforce children types in `formSchemas.ts` to always be React elements.

app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
```diff
@@ -96,7 +96,15 @@ const ListingApplyStepWrapper = ({
...
+            // `{cond && <X/>}`, strings, or null). Guard with isValidElement before reading
+            // `.props`, since React 19 types `ReactElement.props` as `unknown` and an unsafe
+            // cast here would throw at runtime for non-element children.
```
jimlin-sfgov: question: Wouldn't we want to throw at runtime? It reduces the chance of misconfigured schemas making it into production code.

app/views/layouts/application-react.html.slim
```diff
@@ -19,7 +19,12 @@ html lang="en"
         })(window,document,'script','dataLayer','#{ENV['GOOGLE_TAG_MANAGER_KEY']}');
     == javascript_pack_tag 'react_application'
 
-    == stylesheet_pack_tag 'react_application'
+    / Only emit the stylesheet tag when the pack actually has an extracted CSS
+    / entry. All app styles are injected at runtime via style-loader, so the
+    / extracted CSS is not always present (e.g. dev/compile builds), and calling
+    / stylesheet_pack_tag unconditionally raises Shakapacker::Manifest::MissingEntryError.
+    - if Shakapacker.manifest.lookup_pack_with_chunks('react_application', type: :css)
```
jimlin-sfgov, line 27: question: what happens when this condition is false?

config/webpack/loaders/sass.js
```diff
@@ -72,6 +72,10 @@ module.exports = {
         additionalData: tailwindVars,
         sourceMap: true,
         sassOptions: {
+          // ui-components v14 added an `exports` map that hides its deep
```
jimlin-sfgov: suggestion: link the commit / PR where this was done

app/javascript/modules/listingDetails/ListingDetailsNeighborhood.tsx
fwextensions: This was due to empty src strings in the test fixtures, which react 19 didn't like.  Reverted this change and updated the fixtures instead to have proper src's.

CC's explanation:

> Good catch — you pushed on the right thread. I traced where the empty string actually comes from, and it isn't production: imageSrc is getAssetPath("listing-neighborhood.svg"), which returns a real path (or undefined if the key is missing) — never "". The only source of "" was our own test fixtures, e.g. <ListingDetailsAside imageSrc="" />. So the || undefined guard was papering over test data.
> 
> Reverted the guard in all 8 components (they're now identical to main and out of this diff) and fixed the 3 test files to pass "listing-units.svg", matching what listing-detail.tsx really passes. Most of the listing-detail tests already did this. Snapshots regenerated.
> 
> On the typing: you're right that the prop is string, and imageSrc={imageSrc || undefined} was quietly widening it to string | undefined against ui-components' required imageSrc: string. That only compiled because strict/strictNullChecks is off in tsconfig.json — another reason to drop it. And re: an undefined src, that one's fine: React omits the attribute entirely. src="" is the harmful case, since the browser re-requests the current page URL as the image.

app/javascript/pages/account/components/PasswordFieldset.tsx
fwextensions: 👍 renamed to `passwordValue`

app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
fwextensions: CC:

> Reworded. You're right that the speculative list of non-element children was noise. The comment now states the actual reason for isValidElement (React 19 types ReactElement.props as unknown, so the guard is what narrows child enough to read .props.schema) plus the misconfiguration rationale from your other comment.
> 
> One note: the enforcement is on the schema side (children: z.array(StepComponentSchema)), but this component's own children prop is still React.ReactNode, so the type system doesn't guarantee elements at this boundary. Hence keeping the guard, just with an honest comment.

app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
fwextensions: I'm inclined to not break the UI in production in case something slips through.  Changed this to `console.error()` if it finds a bad child.  We could also throw in development mode, if that's useful.

app/views/layouts/application-react.html.slim
fwextensions: CC:

> No stylesheet <link> is emitted, which is the correct outcome rather than an unstyled page: all app styles are injected at runtime by style-loader (sass.js uses injectType: 'singletonStyleTag' in prod, 'styleTag' in dev). The extracted CSS entry doesn't exist in those builds, and calling stylesheet_pack_tag unconditionally raises Shakapacker::Manifest::MissingEntryError. So false means "no extracted CSS to link, styles come from JS." I'll expand the inline comment to say that, since it currently explains the raise but not the styling consequence.

config/webpack/loaders/sass.js
fwextensions: CC:

> Added, and the history was more specific than my comment implied. The package now lives in its own repo, and npm's gitHead pins it down: exports was introduced in PR 200 (released as 14.0.0) as a side effect of splitting ag-grid out of the main bundle — 12.7.6 and 13.0.0 have no exports field at all. A follow-up, PR 201, re-exported css-imports.scss and app-css.scss, but not the 16 deep src/global/** partials base.scss imports, so includePaths is still needed. Both links are now in the comment.

app/javascript/modules/listingDetails/ListingDetailsNeighborhood.tsx
jimlin-sfgov: > you pushed on the right thread

The llm lingo 💀 

> an undefined src, that one's fine: React omits the attribute entirely

good bit of knowledge

app/javascript/pages/form/components/ListingApplyStepWrapper.tsx
```diff
@@ -104,15 +104,16 @@ const ListingApplyStepWrapper = ({
...
-            // cast here would throw at runtime for non-element children.
-            const schema = React.isValidElement<{ schema?: { props?: { divider?: boolean } } }>(
-              child
-            )
-              ? child.props.schema
-              : undefined
+            // `isValidElement` narrows `child` enough to read `.props.schema`, which React 19
+            // types as `unknown`. Step children come from the form component registry via a
+            // validated schema, so a non-element here means the schema is misconfigured: log
+            // it rather than throwing, so a bad step doesn't blank out the applicant's form.
```
jimlin-sfgov, line 110: thought: This comment may be confusing to someone not familiar with form engine code, might be a code smell for this component in general. but addressing it is not in scope here.

app/views/layouts/application-react.html.slim
jimlin-sfgov: nit: comment expansion not detected

jimlin-sfgov (approved):
LGTM

Not completely sure if hyperlink previews and prerender.io will work the same with the updates to react-helmet-async, we'll see when this is deployed to Full

PR #2989, opened 2026-07-21
feat: DAH-4202 Create an account page - email only

cliu02:
## Description

Behind an Unleash flag, creates the first page of the account creation flow for users to enter only their email and receive a one-time passcode. 

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4202

### Version Control

### Code quality

### Code conventions

### Review instructions

### Request eng review


app/javascript/pages/account/create-account.tsx
```diff
@@ -268,6 +268,8 @@ const CreateAccount = (_props: CreateAccountProps) => {
   )
 }
 
+export { CreateAccount }
+
 export default withAppSetup(CreateAccount, {
   useFormTimeout: true,
   pageName: AppPages.CreateAccount,
```
fwextensions, line 275: Was there a resolution to how we want to export components?  Is there a reason the unwrapped component needs to be exported here, like for testing?  Seems potentially confusing to have one with the HOC and one without.

cliu02: The component is exported without `withAppSetup` so that the old and new files can be more cleanly separated for the [Unleash flag](https://github.com/SFDigitalServices/sf-dahlia-web/pull/2989/changes#diff-70d1ac49eb9b71d6aae60cde99dfc86353d5591d5b0250f05e892cf91549de4aR88). This is the same structure used for the account layout feature flag.

fwextensions: Huh, I didn't realize every page component was getting wrapped by an HOC, instead of just doing that once where the pages are mounted.  Seems like a thing we should clean up at some point...

app/javascript/pages/account/create-account.tsx
```diff
@@ -268,6 +268,8 @@ const CreateAccount = (_props: CreateAccountProps) => {
   )
 }
 
+export { CreateAccount }
```
jtcaovan, line 271: question (non-blocking): no issues, just curious on the reasoning of having the export down here vs. having it named inline with the function (export const CreateAccount...) maybe I'm just used to only seeing the default export on the bottom?

jtcaovan (approved):
The spacing between the email field and the **get a code** button doesn't visually match up to the figma designs. Also one code pattern question, but otherwise LGTM

Image

Figma below
Image

app/javascript/pages/account/create-account.tsx
cliu02: Thanks for reviewing @jtcaovan! Feel free to let me know if this doesn't answer your question: https://github.com/SFDigitalServices/sf-dahlia-web/pull/2989#discussion_r3667293984

jtcaovan: Awesome, thanks for answering the same question multiple times!

PR #2993, opened 2026-07-23
feat: DAH-4053 secure-record-response-api

cade-exygy:
## Description

Pass token to the invite_to_response_controller and decode before calling backend

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4053

### Version Control

- [ ] branch name contains the Jira ticket number
- [ ] PR name follows `type: TICKET-NUMBER Description` format, use `DAH-000` if it does not need a ticket

### Code quality

### Code conventions

### Review instructions

### Request eng review


app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -34,6 +31,12 @@ def record_response
 
   private
 
+  def validate_token!
+    JsonWebTokenService.decode_token(params[:t])
```
fwextensions: This just checks that the JWT has a valid signature, but doesn't use the signed contents to record the response, right?  In that case, I think you'd be able to use a valid token to change anyone else's response.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -1,40 +1,66 @@
...
-        "type=#{type}, " \
-        "listingId=#{listing_id}, " \
-        "deadline=#{deadline}, " \
-        "appId=#{application_id}, " \
-        "applicationNumber=#{application_number}, " \
-        "response=#{response}, " \
-        "action=#{action}",
-      )
+    params.expect(record: %i[response action])
+    response, action = params[:record].values_at(:response, :action)
```
fwextensions: Is the yes/no/later response not encoded in the token?  I forget.

fwextensions: It is:

```json
{
  "data": {
    "deadline": "2026-04-30",
    "type": "I2I",
    "appId": "123456",
    "isTest": true,
    "act": "yes"
  },
  "iat": 1782509397
}
```

This expects the application number and listing ID to be in the token, right? ` type, deadline, application_id, application_number, listing_id = token_fields`

Is rails supposed to be minting a new token and passing that to the page that's different than the one we send to the applicant?  I'd assumed it would be the same token.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -1,40 +1,66 @@
...
+
+    type, deadline, application_id, application_number, listing_id = token_fields
+    return unauthorized! if [type, deadline, application_id, application_number,
+                             listing_id].any?(&:blank?)
+
+    parsed_deadline = parse_deadline(deadline)
+    return unauthorized! if parsed_deadline.blank?
```
alulabeshue-sfgov, line 13: **suggestion:** Similar to the `ActionController::ParameterMissing` on line 29 below, missing params on line 8 or a bad deadline are closer to a 400 bad request than a 401 unauthorized

app/controllers/api/v1/invite_to_response_controller.rb
alulabeshue-sfgov: > This expects the application number and listing ID to be in the token, right?

The listingId will not be in the token so we'll still need to pass that as a param to /record-response, currently it is a POST body param. The application number will be in the token so we dont need to pass it in, we can use the tokens app number to make sure clients cant manipulate which application they are recording to.

> I'd assumed it would be the same token

The same token included in the email is the same token that is passed to this API

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -1,40 +1,66 @@
...
-        "applicationNumber=#{application_number}, " \
-        "response=#{response}, " \
-        "action=#{action}",
-      )
+    params.expect(record: %i[response action])
+    response, action = params[:record].values_at(:response, :action)
+
+    type, deadline, application_id, application_number, listing_id = token_fields
+    return unauthorized! if [type, deadline, application_id, application_number,
+                             listing_id].any?(&:blank?)
```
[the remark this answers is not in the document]
alulabeshue-sfgov: @cade-exygy we'll now need to pass the token from the [frontend](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/javascript/api/inviteToApiService.ts).

app/controllers/api/v1/invite_to_response_controller.rb
fwextensions: > The listingId will not be in the token so we'll still need to pass that as a param to /record-response,

Doesn't the API handler already look up the listing ID from the app ID when it gets the token?  Do we need to pass it down to the client and then back up to the server if rails can already get it itself?

alulabeshue-sfgov: In the webapp rails side there is still a [call](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/app/services/dahlia_backend/message_service.rb#L85-L87) to fetch the listing and application. We dont need to do this anymore from webapp since the backend messaging service will do the lookup based on the application id, but we never made this change in webapp. 

I've tested out exactly what we need, and what we can clean up, with all of this I2X code. I think we can do something similar to this [branch](https://github.com/SFDigitalServices/sf-dahlia-web/compare/DAH-4045_clean-up-I2X?expand=1) in Cade's branch here which should make the code cleaner.

cade-exygy: I tried to align some of this PR to the branch you referenced: https://github.com/SFDigitalServices/sf-dahlia-web/compare/DAH-4045_clean-up-I2X?expand=1

After taking a look, it may be easier to do the clean up first, and then apply the JWT Decode check for record-response so we are pulling the correct fields from the token and passing the intended non-token params forward.

alulabeshue-sfgov (approved):
LGTM!

PR #3031, opened 2026-08-15
feat: DAH-4291 invite-to logging improvement: add teardown beacon, env snapshot, structured events

fwextensions:
## Description

Adds diagnostic logging for the invite-to bot detection shipped in #2987: a `pagehide` beacon so a human click in an email in-app webview isn't lost when the page is torn down before the interaction/dwell gate, a passive browser-env snapshot on every signal, and one structured `invite_to.response {json}` log line per event that names *why* a response was suppressed. No applicant-facing behavior change — recording still happens server-side on GET, and the shadow endpoint still records nothing to Salesforce.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4291

## Review instructions

Applies to **any environment with the `temp.webapp.inviteToClientRecording` Unleash flag enabled** (local or a review app). Flag off = no new client behavior, server logging only.

Mint a link (rails console):

```ruby
t = JsonWebTokenService.encode_token(
  type: 'I2I', deadline: 30.days.from_now.to_date.to_s, act: 'yes', appId: '<APP_ID>'
)
```

1. Open `/listings/<LISTING_ID>/next-steps?t=<t>`, move the mouse, and confirm one log line
   `invite_to.response {"...","outcome":"recorded","source":"get","ok":true,...}` plus a second
   with `"source":"client_shadow","trigger":"interaction"` and an `env` object.
2. Reload and close the tab immediately, before touching anything. Expect
   `"trigger":"teardown"` — this is the case the PR exists to capture.
3. Load the page in a background tab and close it without ever viewing it. Expect **no**
   `client_shadow` line (a hidden unload is not evidence of a human).
4. Re-mint with `deadline: 1.day.ago.to_date.to_s` and load it. Expect
   `"outcome":"suppressed","reason":"deadline_passed"` with `deadline_date` / `today` / `late_by`.
5. Switch language on the page. Expect `"reason":"language_change"` with a `referrer` that has
   **no** `?t=` token in it.
6. Confirm the `request_id` on all lines from one page load matches.

PA testing does not apply: there is no user-visible change. Both branches of the flag are exercised by the specs.

### Version Control

- [ ] branch name contains the Jira ticket number

### Code quality

- [x] if the PR is a bugfix, there are tests and logs around the bug

### Code conventions

### Review instructions

### Request eng review

- [x] if PA testing cannot be done, changes are behind a feature flag

<details><summary>Details: what each change is for</summary>

### Client

- **Teardown beacon.** Once the page arms (paint + visible proven), a `pagehide` before interaction/dwell sends `trigger: "teardown"` via `navigator.sendBeacon`, which survives unload where `fetch` is cancelled. Confidence ranks `interaction` > `dwell` > `teardown`.
- **Disarms on `visibilitychange` → hidden**, so an unload from a hidden page does *not* report a teardown — indistinguishable from a background prefetch.
- **Env snapshot** (`webdriver`, `ua`, `coarse`, `cores`, `mem`, `screen`, `lang`, `tz`) on every signal. Evidence only, never a gate: fingerprinting is spoofable and a false positive would reject a real applicant. It exists to classify the ambiguous bucket — a `teardown` with `webdriver:false` and an `Outlook-iOS` UA is a confident webview human; one with a headless UA is not.

### Server

New `InviteToEventLogging` concern emits one line per event: `invite_to.response {json}`, a greppable prefix plus a JSON tail Papertrail can post-process into fields. Values are bounded centrally (256 chars, non-scalars coerced) since these endpoints are unauthenticated, and `env` is allow-listed to the nine keys the client sends.

| Change | Why |
|---|---|
| `reason=` names one of `no_action` / `deadline_passed` / `language_change` / `test_link` | Four causes previously collapsed into one `*NOT* recording` message |
| `deadline_date` / `today` / `late_by` | Logs what the code actually compared, in resolved local terms, instead of a UTC timestamp. `late_by` under a minute is a clock/UX issue, not a late response |
| `referrer` on `language_change` | The one suppression that can silently eat a legitimate first click; scrubbed to scheme/host/path because that referrer carries the invite JWT |
| `request_id` on every line | The three log sites correlate directly instead of by timestamp adjacency, which breaks under concurrency |
| `ok=true/false` | `send_invite_to_response` returns nil on both invalid-action and any rescued error, so "recording" previously meant *attempted* |

Deprecated JWT keys (`response`, `applicationNumber`) are intentionally omitted from the new payloads.

### Two things worth knowing

**A bug this caught in test, not review.** `window.matchMedia?.("…").matches` guards the call but still dereferences `undefined` — it would have thrown in exactly the older/webview environments this feature exists to measure. `snapshotEnv` is now defensive per-property, so a telemetry read can never throw away the signal it reports.

**`ok=` is weaker than it reads.** It's `!result.nil?`, which separates the rescued/invalid-action path from success, but on the happy path `send_invite_to_response` returns whatever `send_message` returns. For `ok=true` to mean "Salesforce accepted it", `send_message` needs to return a real success value — a behavior change rather than logging, so it's not in this PR. Follow-up: DAH-XXXX.

### Testing

Jest 39/39 and RSpec 57/57 on the touched specs, ESLint and Rubocop clean. New coverage: teardown beacon fired and suppressed-when-hidden, `sendBeacon`-unavailable fallback, a throwing env property, unparseable and blank deadlines, `test_link` suppression, each `format_duration` unit, oversized field truncation, and a referrer assertion that the JWT never reaches the log.

</details>


app/controllers/invite_to_controller.rb
```diff
@@ -91,37 +93,84 @@ def record_response(decoded_params)
...
+  end
+
+  # Compact human duration ("2m", "3h", "5d") without pulling in a gem.
+  def format_duration(seconds)
+    seconds = seconds.abs
+    return "#{seconds}s" if seconds < 60
+    return "#{seconds / 60}m" if seconds < 3600
+    return "#{seconds / 3600}h" if seconds < 86_400
+
+    "#{seconds / 86_400}d"
```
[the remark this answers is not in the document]
fwextensions: Verified this one is already the case, so no change: `private` is declared at line 41 and all of these helpers are defined below it. Confirmed by introspection rather than by eye — `InviteToController.action_methods` returns only `index` and `documents`, and `private_method_defined?` is true for `suppression_reason`, `deadline_terms`, `format_duration`, and `record_response`. Nothing new here is routable.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/invite_to_controller.rb
```diff
@@ -91,37 +93,84 @@ def record_response(decoded_params)
...
+    return {} unless reason == 'deadline_passed' && deadline.present?
+
+    deadline_time = Time.zone.parse(deadline)
+    {
+      deadline_date: deadline_time.to_date.to_s,
+      today: Time.zone.today.to_s,
+      late_by: format_duration((Time.zone.now - deadline_time).to_i),
+    }
+  rescue ArgumentError, TypeError
+    { deadline_raw: deadline } # unparseable - surface it rather than crash the log line
```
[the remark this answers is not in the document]
fwextensions: Good catch, and it was worse than described. `deadline_terms` is only reached when `reason == 'deadline_passed'`, which means `deadline_has_passed?` already ran `Time.zone.parse(deadline).to_date` — so an unparseable deadline raised `NoMethodError` there and 500'd the page before this method was ever called. Fixed at the source: `deadline_has_passed?` now handles the nil return and rescues, treating a deadline we cannot verify as passed (suppressing is the conservative choice over recording against an unverifiable deadline). `deadline_terms` keeps its own nil guard and rescue, and emits `deadline_raw` so the bad value stays visible. Covered by a spec asserting the request still returns 200.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -40,23 +43,26 @@ def record_response
...
+      reason: 'shadow_human_verified',
+      type: type,
+      listing_id: listing_id,
+      deadline: deadline,
+      app_id: application_id,
+      act: act,
+      trigger: trigger, # interaction | dwell | teardown
+      elapsed_ms: elapsed_ms,
+      env: record[:env]&.to_unsafe_h, # passive browser signals for post-hoc classification
     )
```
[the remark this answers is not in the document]
fwextensions: Agreed, fixed. `env` is now allow-listed to the nine keys `snapshotEnv()` actually sends (`webdriver, ua, touch, coarse, cores, mem, screen, lang, tz`), rejects nested Array/Hash values, truncates each value to 256 chars, and omits the key entirely when nothing recognized remains.

One caveat worth stating plainly rather than implying this fully closes the hole: Rails logs the raw request body itself in its own `Parameters:` line at INFO, which this sanitizer does not and cannot control. So the allow-list bounds *our structured event* — the thing we query, aggregate, and retain — but it does not stop unfiltered params from appearing in the request log. Properly closing that would mean adding these to `filter_parameters`, which is a broader change than this PR. Surfacing it so the guarantee isn't overstated.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/invite_to_controller.rb
```diff
@@ -91,37 +93,84 @@ def record_response(decoded_params)
...
+  end
```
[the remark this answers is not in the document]
fwextensions: Added. There is now a spec passing `deadline: 'not-a-date'` that asserts the request still returns 200, nothing is sent to the messaging service, and the structured log contains `"deadline_raw":"not-a-date"`. Also added coverage for the `test_link` suppression reason, each `format_duration` unit (s/m/h/d plus a negative delta), and the rescue paths where `Time.zone.parse` raises.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -40,23 +50,26 @@ def record_response
...
+      reason: 'shadow_human_verified',
+      type: type,
+      listing_id: listing_id,
+      deadline: deadline,
+      app_id: application_id,
+      act: act,
+      trigger: trigger, # interaction | dwell | teardown
+      elapsed_ms: elapsed_ms,
+      env: sanitized_env(record[:env]),
     )
```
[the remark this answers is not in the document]
fwextensions: Agreed and fixed, generalized rather than patched per-field. `log_invite_event` now bounds every value centrally: strings truncate to 256 chars and non-scalars coerce to a bounded string. Doing it in the concern rather than at each call site means a newly added field can't reintroduce the gap. `sanitized_env` is correspondingly simplified to key allow-listing plus non-scalar rejection, since truncation is now handled upstream. Covered by a spec sending 5000-char `type`/`act` values and asserting the emitted event stays under 3KB.

While in here I also acted on your suppressed comment about `request.referrer` — that one was more than log hygiene. `language_change?` only fires when the referrer is *another next-steps URL*, and those URLs carry the invite JWT in `?t=`, so the raw referrer would have written live tokens into the logs. It's now scrubbed to scheme/host/path (all the locale-prefix comparison needs), with a spec asserting the token value never appears in the event.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/concerns/invite_to_event_logging.rb
```diff
@@ -0,0 +1,46 @@
+# Structured logging for invite-to (I2A/I2I) response events.
+#
+# Every recording, suppression, and shadow-mode human-verification emits a single
+# line with a stable prefix and a JSON payload at the end, so Papertrail (and
+# scripts/parse-invite-to-logs.js) can group by `outcome`/`reason` and correlate
```
fwextensions: This is a local script used for offline processing of logs to analyze the bot detection results.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +82,18 @@ def log_human_verified
...
+  # centrally by log_invite_event. Returns nil when nothing usable is left, so the
+  # key drops out of the log line entirely.
+  def sanitized_env(env)
+    return nil if env.blank?
+
+    env.to_unsafe_h
+       .slice(*ENV_KEYS.map(&:to_s))
+       .reject { |_key, value| value.is_a?(Array) || value.is_a?(Hash) }
+       .presence
+  end
```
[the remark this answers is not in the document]
fwextensions: Checked this one against actual Rails behavior before changing anything, and it turns out not to apply: `to_unsafe_h` converts nested params **recursively**, so by the time the filter runs, a nested value is an `ActiveSupport::HashWithIndifferentAccess`, which *is* `is_a?(Hash)`. Verified in a console:

```ruby
env = ActionController::Parameters.new(env: { ua: "x", nested: { deep: { deeper: "v" } } })[:env]
env.to_unsafe_h["nested"].class  # => ActiveSupport::HashWithIndifferentAccess
env.to_unsafe_h["nested"].is_a?(Hash)  # => true
```

So no code change. You did surface a real gap in the *test* though: the existing nesting case nested under a key that isn't allow-listed, so `slice` dropped it and the non-scalar filter was never actually exercised. Added a spec that nests under an allow-listed key (`ua`) and asserts the inner value never reaches the log.

Belt-and-braces regardless: `log_invite_event` coerces anything non-scalar to a bounded string, so even if a value did slip through it would be truncated rather than logged as an unbounded structure.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

fwextensions:
Also acted on both **suppressed** comments from this round — both were legitimate, and one was a bug I introduced in the previous round.

**Blank deadline misclassified as expired** (`invite_to_controller.rb:149`). Correct, and it was my regression: once `deadline_has_passed?` started reporting unparseable input as passed, a `""` deadline began logging as `reason=deadline_passed` — a wrong reason in the very logs that exist to make reasons trustworthy. Now uses `deadline.present?`.

⚠️ **One judgement call I'd like a human on.** My first fix introduced a distinct `missing_deadline` suppression, which reads cleaner — but I backed it out, because it would have stopped recording responses when the deadline is nil, and the pre-existing code *records* in that case. Failing to record a real "yes" costs an applicant their spot, which is much worse than recording against a missing deadline (that at least surfaces as `ok=false`). So current behavior is unchanged from `main` for nil, and only the `""` crash is fixed.

If you'd rather a missing deadline be suppressed outright, that's a one-line change — but it's an applicant-facing tradeoff, so I didn't want to make it silently on a logging PR. Happy to follow up either way.

**Ignored `sendBeacon` return** (`useAutoRecordInviteToResponse.ts:152`). Also correct: `fired` was already set, so a refused beacon dropped the signal — precisely the failure this feature exists to prevent. It now falls back to the normal request. That may not survive an immediate unload, but `pagehide` also fires on bfcache, where it completes fine, so it's a strict improvement over dropping it. Covered both ways (fallback fires when refused; no double-send when accepted).

RSpec 54/54, Jest 22/22.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

fwextensions:
The CodeQL failures were due to some issue with GH.  Doesn't seem to be any way to re-run them other than an empty commit:

> Run github/codeql-action/init@v4
> Existing job run UUID is dc955843-90b8-42b6-96a2-f89c88c6d6b7.
> Warning: An unexpected error occurred when sending a status report: No server is currently available to service your request. Sorry about that. Please try resubmitting your request and contact us if the problem persists.
> Error: Encountered an error while trying to determine feature enablement: HttpError: No server is currently available to service your request. Sorry about that. Please try resubmitting your request and contact us if the problem persists.

jimlin-sfgov:
There appears to be quite a few changes here, and no review instructions to boot. Also, the abundance of multi-line comments with semantic fluff are a slog to get through. I don't see this getting approved and merged quickly, unless reviewers dedicate a good chunk of time to focus on this PR.

I strongly encourage you to strike a better balance of verbosity and clarity, whether that's through tuning your prompts, or editing the LLM output by hand. As a reviewer I want concise human-readable content. Amending and refining LLM output to align with that should be done before a PR is considered ready for review, not during the review process.

🧑 Written by @jimlin-sfgov

fwextensions:
@jimlin-sfgov shrank the comments by a third and streamlined the PR description.  Items 1 and 2 in the review instructions are the more important changes, but the main thing is just making sure the new logging doesn't interfere with the existing server-side response recording.  62% of the PR is tests to get to the 100% qlty threshold.

app/controllers/concerns/invite_to_event_logging.rb
```diff
@@ -0,0 +1,40 @@
+# Structured logging for invite-to (I2A/I2I) response events. Every recording,
```
jimlin-sfgov, line 1: thought: It would be nice to have structured logging throughout our Rails app, instead of building out a little module just to serve the needs of your feature. Options include:
- configure logging behavior in `Rails.application.configure do ... config.logger = ... `
- use a gem like [logstruct](https://github.com/DocSpring/logstruct)
- upgrade to [Rails 8.1 ](https://rubyonrails.org/2025/10/22/rails-8-1)which supports structured logging natively

amyc-sfds:
If I may, I want to jump in with a question. The associated Jira ticket describes the technical problem and the implementation, but not the impact or risk of leaving it unaddressed—and it is marked Critical in priority. Is this is a "drop what you're doing" situation? If it is, then I think this should be a team conversation rather than a side PR.

I ask because this PR needs a more thorough review. The code is relatively self-contained, but it extends several controllers and introduces new logging logic, so the approach you've taken here should be validated against our existing patterns and architecture. That level of review takes focused time and attention, so it would displace other things currently being worked on.

For next steps, it would be helpful to get alignment on priorities. If moving this forward is critical/time-sensitive, I recommend bringing it to the team and putting it through your standard process. The engineers own the code that goes into prod, and I want them to have the room to look at this properly.

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
 }
 
-// Shadow-mode only: records nothing, just logs that client-side detection judged this a real
-// human click. Used to measure the detection against live traffic before enabling client recording.
-export const logHumanVerifiedClick = async (record: {
+export type HumanVerifiedTrigger = "interaction" | "dwell" | "teardown"
+
+// Passive browser signals attached to every human-verified log. Spoofable, so evidence only -
+// never a gate. Used to tell an in-app-webview human from a headless scanner after the fact.
+export interface EnvSnapshot {
```
jimlin-sfgov: question: So my [earlier slack comment](https://sfdigitalservices.slack.com/archives/C030EGQAKK2/p1785363821195799?thread_ts=1784839767.115529&cid=C030EGQAKK2) about utilizing the `navigator` object was taken into consideration? Have you also considered the PII implications as well? If we combine this digital fingerprint with `listing_id` and `application_id`, does that become a privacy violation?

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
+    // only the call still throws on the undefined result.
+    coarse: safe(() => window.matchMedia?.("(pointer: coarse)")?.matches, false),
+    cores: safe(() => navigator.hardwareConcurrency, null),
+    mem: safe(() => (navigator as Navigator & { deviceMemory?: number }).deviceMemory, null),
+    screen: safe(
+      () => `${window.screen.width}x${window.screen.height}@${window.devicePixelRatio}`,
+      ""
+    ),
+    lang: safe(() => navigator.languages?.[0] ?? navigator.language, ""),
+    tz: safe(() => Intl.DateTimeFormat().resolvedOptions().timeZone, ""),
```
jimlin-sfgov: question: How was this particular set of properties decided on? There are additional ways to get clued in on headless browsers, e.g. `navigator.connection.rtt` and `eval.toString().length`

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
+  try {
+    return read() ?? fallback
+  } catch {
+    return fallback
+  }
+}
+
+// Runs inside the fire path, so every lookup is guarded individually - losing the signal to a
+// telemetry error would defeat the point.
+export const snapshotEnv = (): EnvSnapshot => {
```
jimlin-sfgov: suggestion: Remove from file, this file is for functions that make API calls.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -107,52 +115,69 @@ export const useAutoRecordInviteToResponse = ({
...
 
+    const buildRecord = (trigger: HumanVerifiedTrigger): HumanVerifiedRecord => ({
+      appId,
+      listingId,
+      deadline,
+      act,
+      type,
+      trigger,
+      elapsedMs: Date.now() - armedAt,
+      env: snapshotEnv(),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -107,52 +115,69 @@ export const useAutoRecordInviteToResponse = ({
...
-        trigger,
-        elapsedMs: Date.now() - armedAt,
-      }).catch((error) => {
+      void logHumanVerifiedClick(buildRecord(trigger)).catch((error) => {
         console.error("Error logging human-verified invite-to click:", error)
       })
     }
 
+    // Flush on unload if we armed but never reached the stronger interaction/dwell gate -
+    // the normal shape of an in-app email webview that gets torn down fast.
```
jimlin-sfgov: suggestion: Please add a human touch this comment. It sounds like a commentary on toilet-based siege warfare 🚽 🔫 🏰 

Please describe the specific scenario that this new function is supposed to handle, you mention elsewhere something about "email in-app webview"

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -73,6 +79,8 @@ export const useAutoRecordInviteToResponse = ({
     // Set when the fire conditions actually arm (after the first paint, while visible), so
     // elapsedMs reflects visible dwell/interaction time and excludes hidden/background-tab time.
     let armedAt = 0
```
jimlin-sfgov, line 81: question: Why set this to zero, or at all?

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
+  coarse: boolean // matchMedia("(pointer: coarse)") => phone/tablet
+  cores: number | null // hardwareConcurrency
+  mem: number | null // deviceMemory (Chromium only)
+  screen: string // `${w}x${h}@${dpr}`
+  lang: string
+  tz: string
+}
+
+// Reads a single env value, falling back when the property is missing or throws.
+const safe = <T>(read: () => T, fallback: T): T => {
```
jimlin-sfgov: suggestion: Remove from file, this file is for functions that make API calls.

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
   })
 }
 
-// Shadow-mode only: records nothing, just logs that client-side detection judged this a real
-// human click. Used to measure the detection against live traffic before enabling client recording.
-export const logHumanVerifiedClick = async (record: {
+export type HumanVerifiedTrigger = "interaction" | "dwell" | "teardown"
```
jimlin-sfgov: suggestion: Remove from file, this file is for functions that make API calls.

jimlin-sfgov: suggestion: use "dwellTime" insteads of "dwell"

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -1,9 +1,16 @@
 class Api::V1::InviteToResponseController < ApiController
+  include InviteToEventLogging
+
+  # What snapshotEnv() sends. Allow-listed rather than a free-form hash because this
+  # endpoint is unauthenticated.
+  ENV_KEYS = %i[webdriver ua touch coarse cores mem screen lang tz].freeze
```
jimlin-sfgov: suggestion: rename these so it's more obvious what they are, without need to to refer to the javascript function. E.g. `ua` -> `user_agent`

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -1,9 +1,16 @@
...
+
   before_action :validate_token!, only: :record_response
 
   def record_response
     record_params = params.expect(record: %i[listingId action])
-    return unauthorized! if [record_params[:listingId], record_params[:action]].any?(&:blank?)
+    return unauthorized! if [record_params[:listingId],
+                             record_params[:action]].any?(&:blank?)
```
jimlin-sfgov: question: why an array? do you plan to add more params?

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -34,29 +41,28 @@ def record_response
...
-      "type=#{type.inspect}, " \
-      "listingId=#{listing_id.inspect}, " \
-      "deadline=#{deadline.inspect}, " \
-      "appId=#{application_id.inspect}, " \
-      "act=#{act.inspect}, " \
-      "trigger=#{trigger.inspect}, " \
-      "elapsedMs=#{elapsed_ms.inspect}",
+      record.values_at(:type, :deadline, :appId, :listingId, :act, :trigger, :elapsedMs)
+
+    # Values come from an unauthenticated endpoint; log_invite_event's to_json escapes
```
jimlin-sfgov, line 50: question: How does an unauthenticated endpoint allow for log forging? What do you mean by log forging?

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
-}) => {
+  env: EnvSnapshot
+}
+
+// Shadow mode: records nothing, only logs that client-side detection judged this a human click.
+export const logHumanVerifiedClick = async (record: HumanVerifiedRecord) => {
   return post("/api/v1/next-steps/log-human-verified", { record })
 }
+
+// Teardown path: sendBeacon survives unload, where a fetch would be cancelled. Returns false if
```
jimlin-sfgov: question: This comment mentions "unload", but it's being called in a "pagehide" even listener, why?

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -107,52 +115,69 @@ export const useAutoRecordInviteToResponse = ({
...
       })
     }
 
+    // Flush on unload if we armed but never reached the stronger interaction/dwell gate -
+    // the normal shape of an in-app email webview that gets torn down fast.
+    const handlePageHide = () => {
+      if (fired || !armed) return
+      fired = true
+      cleanup()
+      const record = buildRecord("teardown")
```
jimlin-sfgov: suggestion: Use a name that better describes this scenario. "teardown" brings to mind the process of cleaning up data after running a test suite.

app/controllers/concerns/invite_to_event_logging.rb
```diff
@@ -0,0 +1,40 @@
...
+# suppression, and shadow-mode human-verification emits one line with a stable prefix and
+# a JSON payload, so Papertrail can group by `outcome`/`reason` and correlate the three log
+# sites by `request_id` rather than by timestamp adjacency on a shared dyno.
+module InviteToEventLogging
+  extend ActiveSupport::Concern
+
+  INVITE_EVENT = 'invite_to.response'.freeze
+
+  # Longest string logged for any single field. Bounded centrally rather than per call
+  # site so a newly added field can't reintroduce the gap.
```
jimlin-sfgov: suggestion: add a human touch to this comment. "Bounded centrally ... reintroduce the gap" doesn't read well

app/controllers/api/v1/invite_to_response_controller.rb
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -34,29 +41,28 @@ def record_response
...
-  # Salesforce state changes and no applicant email is sent. It only logs that the
-  # client-side human-detection judged this page view to be a real human, so we can
-  # compare against the server-side GET recording during the shadow rollout.
-  # Safe to call repeatedly.
+  # Shadow-mode endpoint: never calls sf-dahlia-backend, so no Salesforce write and no
+  # applicant email. Logs only that client-side detection judged this a human. Idempotent.
   def log_human_verified
-    params.expect(record: %i[type deadline appId listingId act trigger elapsedMs])
+    record = params.expect(record: [:type, :deadline, :appId, :listingId, :act, :trigger,
+                                    :elapsedMs, { env: ENV_KEYS }])
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -34,29 +41,28 @@ def record_response
...
+      source: 'client_shadow',
+      reason: 'shadow_human_verified',
+      type: type,
+      listing_id: listing_id,
+      deadline: deadline,
+      app_id: application_id,
+      act: act,
+      trigger: trigger, # interaction | dwell | teardown
+      elapsed_ms: elapsed_ms,
+      env: sanitized_env(record[:env]),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +77,17 @@ def log_human_verified
 
   private
 
+  # Keeps only the known env keys and drops non-scalar values; log_invite_event handles
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +77,17 @@ def log_human_verified
...
+  # truncation. nil when nothing usable is left, so the key drops out of the line.
+  def sanitized_env(env)
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +77,17 @@ def log_human_verified
...
+    return nil if env.blank?
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +77,17 @@ def log_human_verified
...
+
+    env.to_unsafe_h
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/controllers/api/v1/invite_to_response_controller.rb
```diff
@@ -71,6 +77,17 @@ def log_human_verified
...
+       .slice(*ENV_KEYS.map(&:to_s))
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -1,11 +1,28 @@
 import { post } from "../../api/apiService"
-import { recordResponse, logHumanVerifiedClick } from "../../api/inviteToApiService"
+import {
+  recordResponse,
+  logHumanVerifiedClick,
+  beaconHumanVerifiedClick,
+  snapshotEnv,
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -1,11 +1,28 @@
...
 
+const humanRecord: HumanVerifiedRecord = {
+  listingId: "a0w123",
+  appId: "a0o123",
+  deadline: "2099-01-01",
+  act: "yes",
+  type: INVITE_TO_X.APPLY,
+  trigger: "interaction",
+  elapsedMs: 1234,
+  env: snapshotEnv(),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -27,17 +44,64 @@ describe("inviteToApiService", () => {
...
+      await logHumanVerifiedClick(humanRecord)
+      expect(post).toHaveBeenCalledWith("/api/v1/next-steps/log-human-verified", {
+        record: humanRecord,
+      })
+    })
+  })
+
+  describe("snapshotEnv", () => {
+    it("captures passive browser signals used to classify ambiguous clicks", () => {
+      const env = snapshotEnv()
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -27,17 +44,64 @@ describe("inviteToApiService", () => {
...
+      expect(env).toEqual(
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -27,17 +44,64 @@ describe("inviteToApiService", () => {
...
+    })
+
+    // This runs inside the fire path, so a throwing property must never take the signal
+    // with it - some webviews raise on these reads rather than returning undefined.
+    it("falls back per-field when a property getter throws", () => {
+      const spy = jest.spyOn(navigator, "userAgent", "get").mockImplementation(() => {
+        throw new Error("blocked")
+      })
+
+      expect(() => snapshotEnv()).not.toThrow()
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -27,17 +44,64 @@ describe("inviteToApiService", () => {
...
+      expect(snapshotEnv().ua).toBe("")
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/api/inviteToApiService.test.ts
```diff
@@ -27,17 +44,64 @@ describe("inviteToApiService", () => {
...
+      // Other fields still populate normally.
+      expect(snapshotEnv().screen).toMatch(/\d+x\d+@/)
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -1,12 +1,35 @@
 import { renderHook, cleanup } from "@testing-library/react"
 import { useAutoRecordInviteToResponse } from "../../hooks/useAutoRecordInviteToResponse"
-import { recordResponse, logHumanVerifiedClick } from "../../api/inviteToApiService"
+import {
+  recordResponse,
+  logHumanVerifiedClick,
+  beaconHumanVerifiedClick,
+  snapshotEnv,
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -1,12 +1,35 @@
...
+} from "../../api/inviteToApiService"
 
 jest.mock("../../api/inviteToApiService", () => ({
   recordResponse: jest.fn(),
   logHumanVerifiedClick: jest.fn(),
+  beaconHumanVerifiedClick: jest.fn(),
+  snapshotEnv: jest.fn(),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -1,12 +1,35 @@
...
 }))
 
+const TEST_ENV = {
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -38,6 +61,9 @@ describe("useAutoRecordInviteToResponse", () => {
     jest.clearAllMocks()
     setVisibility("visible")
     ;(logHumanVerifiedClick as jest.Mock).mockResolvedValue(undefined)
+    ;(snapshotEnv as jest.Mock).mockReturnValue(TEST_ENV)
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -195,6 +221,83 @@ describe("useAutoRecordInviteToResponse", () => {
     expect(logHumanVerifiedClick).not.toHaveBeenCalled()
   })
 
+  it("attaches the env snapshot to the logged payload", () => {
```
jimlin-sfgov, line 227: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/hooks/useAutoRecordInviteToResponse.test.tsx
```diff
@@ -195,6 +221,83 @@ describe("useAutoRecordInviteToResponse", () => {
...
+    renderHook(() => useAutoRecordInviteToResponse(baseArgs))
+    flushPaintFrames()
+    jest.advanceTimersByTime(2000)
+
+    expect(logHumanVerifiedClick).toHaveBeenCalledWith(
+      expect.objectContaining({ env: expect.objectContaining({ ua: "test-ua", webdriver: false }) })
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/__tests__/pages/invite-to-apply.test.tsx
```diff
@@ -15,6 +15,8 @@ jest.mock("../../api/listingApiService")
 jest.mock("../../api/inviteToApiService", () => ({
   recordResponse: jest.fn(),
   logHumanVerifiedClick: jest.fn(),
+  beaconHumanVerifiedClick: jest.fn(),
+  snapshotEnv: jest.fn(() => ({})),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/api/inviteToApiService.ts
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
+  touch: number // navigator.maxTouchPoints
+  coarse: boolean // matchMedia("(pointer: coarse)") => phone/tablet
+  cores: number | null // hardwareConcurrency
+  mem: number | null // deviceMemory (Chromium only)
+  screen: string // `${w}x${h}@${dpr}`
+  lang: string
+  tz: string
+}
+
+// Reads a single env value, falling back when the property is missing or throws.
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/api/inviteToApiService.ts
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/api/inviteToApiService.ts
```diff
@@ -18,16 +18,72 @@ export const recordResponse = async (
...
   listingId: string
   appId: string
   deadline: string
   act: string
   type: string
-  trigger: string
+  trigger: HumanVerifiedTrigger
   elapsedMs: number
-}) => {
+  env: EnvSnapshot
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

app/javascript/hooks/useAutoRecordInviteToResponse.ts
```diff
@@ -1,5 +1,11 @@
 import { useEffect } from "react"
-import { logHumanVerifiedClick } from "../api/inviteToApiService"
+import {
+  logHumanVerifiedClick,
+  beaconHumanVerifiedClick,
+  snapshotEnv,
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -184,6 +184,7 @@
         act: 'yes',
         trigger: 'interaction',
         elapsedMs: 1234,
+        env: { webdriver: false, ua: 'Mozilla/5.0 FBAN/FBIOS', coarse: true },
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+          '"source":"client_shadow"',
+          '"reason":"shadow_human_verified"',
+          '"app_id":"a0o123"',
+          '"act":"yes"',
+          '"trigger":"interaction"',
         ),
       )
     end
 
+    # Note: ActionController::Parameters stringifies scalars, so env booleans serialize as
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+    # "false"/"true" rather than JSON booleans. Harmless for querying, but assert what we emit.
+    it 'includes the passive browser env snapshot for post-hoc classification' do
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      post :log_human_verified, params: { record: valid_record }
+
+      expect(Rails.logger).to have_received(:info).with(
+        a_string_including('"env":', '"ua":"Mozilla/5.0 FBAN/FBIOS"', '"webdriver":"false"'),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      )
+    end
+
+    it 'drops unknown env keys, nested values, and over-long values' do
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      logged = []
+      allow(Rails.logger).to receive(:info) { |msg| logged << msg.to_s }
+
+      post :log_human_verified, params: {
+        record: valid_record.merge(
+          env: {
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      # Scope to our structured event: Rails' own "Parameters:" line echoes the raw
+      # request body, which the sanitizer does not (and cannot) control.
+      event = logged.find { |msg| msg.start_with?('invite_to.response ') }
+      expect(event).to include('"webdriver":"false"')
+      expect(event).not_to include('should-not-be-logged')
+      expect(event).not_to include('"evil"')
+      expect(event).not_to include('a' * 300)
+    end
+
+    # Unauthenticated endpoint: every logged value is bounded, not just env.
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      expect(response).to be_ok
+      event = logged.find { |msg| msg.start_with?('invite_to.response ') }
+      expect(event).not_to include('T' * 300)
+      expect(event).not_to include('A' * 300)
+      expect(event.length).to be < 3000
+    end
+
+    # to_unsafe_h converts nested ActionController::Parameters to HashWithIndifferentAccess,
+    # so the non-scalar filter catches them - assert that on an allow-listed key.
+    it 'drops a nested structure sent under an allow-listed env key' do
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      logged = []
+      allow(Rails.logger).to receive(:info) { |msg| logged << msg.to_s }
+
+      post :log_human_verified, params: {
+        record: valid_record.merge(env: { ua: { nested: { deep: 'sneaky-value' } }, tz: 'UTC' }),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      }
+
+      expect(response).to be_ok
+      event = logged.find { |msg| msg.start_with?('invite_to.response ') }
+      expect(event).not_to include('sneaky-value')
+      expect(event).to include('"tz":"UTC"')
+    end
+
+    it 'omits env entirely when no recognized keys are supplied' do
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      post :log_human_verified, params: {
+        record: valid_record.merge(env: { evil: 'nope' }),
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

spec/controllers/api/v1/invite_to_response_controller_spec.rb
```diff
@@ -192,23 +193,101 @@
...
+      }
+
+      expect(response).to be_ok
+      expect(Rails.logger).not_to have_received(:info).with(a_string_including('"env":'))
```
jimlin-sfgov: suggestion: Do not use env, could be confused with environment and environment variables.

jimlin-sfgov:
There are quite a few review comments. Thoughts on adding a `claude.md` to steer your LLM work better @fwextensions ?

jimlin-sfgov:
Another thought: it would be helpful for engineers if you could illustrate what you've been seeing in the logs so far, and how those logs motivated you to make this new PR

app/controllers/api/v1/invite_to_response_controller.rb
fwextensions: That's pre-existing code.  Prettier or something just wrapped it to two lines.

app/controllers/api/v1/invite_to_response_controller.rb
fwextensions: Agreed, and it reads worse in Ruby than in the TS — the constant sat next to real `ENV` usage, and `Rails.env` / `request.env` are both live meanings.

Renamed throughout to `browser`: `BROWSER_KEYS`, `sanitized_browser`, the payload key, and on the client `EnvSnapshot`/`snapshotEnv()` → `BrowserSnapshot`/`snapshotBrowser()`. It's what the thing actually is.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

fwextensions: Agreed. Expanded to match the browser APIs they come from, so the allow-list is readable without cross-referencing the JS:

| was | now | source |
|---|---|---|
| `ua` | `userAgent` | `navigator.userAgent` |
| `touch` | `maxTouchPoints` | `navigator.maxTouchPoints` |
| `coarse` | `coarsePointer` | `matchMedia("(pointer: coarse)")` |
| `cores` | `cpuCores` | `navigator.hardwareConcurrency` |
| `mem` | `deviceMemory` | `navigator.deviceMemory` |
| `lang` | `language` | `navigator.languages[0]` |
| `tz` | `timezone` | `Intl…resolvedOptions().timeZone` |

`webdriver` and `screen` already matched their source names, so they're unchanged. Kept camelCase since these are wire keys and the allow-list has to match the payload exactly — consistent with the `appId`/`listingId`/`elapsedMs` params in the same `expect` call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/hooks/useAutoRecordInviteToResponse.ts
fwextensions: Fair — "teardown" does read like test-suite cleanup. Renamed the trigger to `pageExit`, which says what actually happened: the visitor left before we could confirm interaction or dwell time.

The handler stays `handlePageHide`, since that's literally the event it's bound to.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/api/inviteToApiService.ts
fwextensions: Agreed — renamed to `dwellTime`, which also matches the `dwellTimeMs` option that drives it.

Worth noting this one differs from the other renames in the PR: `dwell` is already being emitted in production (~53 signals since the Aug 13 rollout), so it's the only value with historical data behind it. The log parser now folds the old name into the new one rather than reporting two buckets for the same thing, with a note to drop that alias once pre-rename logs age out of the retention window.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/api/inviteToApiService.ts
fwextensions: Good catch — the comment was imprecise and I've reworded it.

We deliberately don't listen to `unload`: it's effectively deprecated, unreliable on mobile (often never fires when the OS kills a tab), and registering an `unload` handler disqualifies the page from bfcache, which would itself change behavior. `pagehide` is the supported replacement and covers both real teardown and bfcache.

I was using "unload" loosely to mean "the page going away". The point that actually matters is about `sendBeacon`: a normal `fetch` gets cancelled when the page goes away, whereas a beacon is queued by the browser and delivered afterwards. The comment now reads:

> sendBeacon is queued by the browser and survives page dismissal, where a fetch would be cancelled.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/controllers/concerns/invite_to_event_logging.rb
fwextensions: Agreed, that one was dense. Reworded to:

```ruby
# Caps how long any single logged value can be. Doing it here, rather than at each
# call site, means a field someone adds later is capped automatically without having
# to remember to do it.
```

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/api/inviteToApiService.ts
```diff
@@ -17,16 +17,77 @@ export const recordResponse = async (
...
+      null
+    ),
+    screen: safe(
+      () => `${window.screen.width}x${window.screen.height}@${window.devicePixelRatio}`,
+      ""
+    ),
+    language: safe(() => navigator.languages?.[0] ?? navigator.language, ""),
+    timezone: safe(() => Intl.DateTimeFormat().resolvedOptions().timeZone, ""),
+  }
+}
```
alulabeshue-sfgov: **suggestion:** this function does not make API calls and is used only in useAutoRecordInviteToResponse.ts. Move `snapshotBrowser` and `safe` to useAutoRecordInviteToResponse.ts

app/controllers/invite_to_controller.rb
```diff
@@ -87,29 +81,95 @@ def record_response(decoded_params)
...
+
+  # Compact human duration ("2m", "3h", "5d") without pulling in a gem.
+  def format_duration(seconds)
+    seconds = seconds.abs
+    return "#{seconds}s" if seconds < 60
+    return "#{seconds / 60}m" if seconds < 3600
+    return "#{seconds / 3600}h" if seconds < 86_400
+
+    "#{seconds / 86_400}d"
   end
```
alulabeshue-sfgov: suggestion: move all of these logging helper functions to invite_to_event_logging.rb to encapsulate all of the logging changes into one file. Move `format_duration`, `scrubbed_referrer`, and `deadline_terms`

fwextensions:
Here's a branch that will revert the logging changes after we've collected enough data: https://github.com/SFDigitalServices/sf-dahlia-web/tree/DAH-4291-revert-invite-to-logging

alulabeshue-sfgov:
> Here's a branch that will revert the logging changes after we've collected enough data: https://github.com/SFDigitalServices/sf-dahlia-web/tree/DAH-4291-revert-invite-to-logging

Its more about having a clear separation of concern where the controller files and API service files are not getting bloated with logging logic overtime. I think you already have `invite_to_event_logging.rb` and `useAutoRecordInviteToResponse.ts`, those files should be responsible for logging and human detection.

app/controllers/invite_to_controller.rb
fwextensions: Done — `format_duration`, `scrubbed_referrer`, and `deadline_terms` now live in `invite_to_event_logging.rb` under a `# --- payload shaping ---` divider.

Agreed on the rationale: all three exist only to turn raw request values into loggable fields, so they belong beside `log_invite_event` rather than in the controller that happens to call them. `scrubbed_referrer` reads `request`, which the concern already uses for `request_id`, so nothing needed threading through.

Specs needed no changes — they reach these via `controller.send(...)`, and included-module methods resolve identically. RSpec 58/58.

One side effect worth naming: `Api::V1::InviteToResponseController` also includes the concern, so it now carries three private methods it doesn't use. Harmless, and the cohesion win is worth it, but flagging it rather than leaving you to spot it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/javascript/api/inviteToApiService.ts
fwextensions: Agreed it doesn't belong here — moved out, though to a new `app/javascript/util/browserSnapshotUtil.ts` rather than into the hook, alongside the other `*Util` modules.

Two reasons I went sideways rather than into `useAutoRecordInviteToResponse.ts`:

1. **`BrowserSnapshot` has to stay reachable from the payload type.** It types `HumanVerifiedRecord.browser`, so if the interface moved into the hook, `inviteToApiService` would have to import from the hook — and the hook already imports from the service, so that's a circular import. Splitting the interface from its only constructor across two files also seemed worse than what's there today.

2. **It stays mockable.** The hook test currently does `jest.mock(...)` on the module boundary and feeds a fixed snapshot. Module-internal to the hook, that stops working, and the test has to assert against whatever jsdom happens to report instead.

The separate module gets the thing you're actually after — no browser-detection code in a file named ApiService — without either cost, and keeps the hook (already ~215 dense lines) from growing.

Tests moved with it: `snapshotBrowser`'s cases are now `__tests__/util/browserSnapshotUtil.test.ts`, the hook test mocks the new module, and `invite-to-apply.test.tsx` drops its stale mock entry. Jest 39/39, tsc and ESLint clean.

Happy to fold it into the hook instead if you'd still prefer that — just wanted the tradeoff visible first.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

alulabeshue-sfgov (approved):

PR #3049, opened 2026-08-26
fix: DAH-4328 Show SRO info card when unit summaries are in the reserved bucket

fwextensions:
## Description

The "Single Room Occupancy" info card was missing on [1303 Larkin Street](https://housing.sfgov.org/listings/a0W7y00000J1wlNEAR?preview=true), which is a 62+ SRO listing.  This seems to be because the SRO helpers only read `unitSummaries.general` and 1303 Larkin has all of its unit summaries in `unitSummaries.reserved`.

**Before**

image

<br>
<br>

**After**

image

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4328

## Review instructions

1. Open a reserved-community SRO listing, e.g. `/listings/a0W7y00000J1wlNEAR`.
2. The Single Room Occupancy info card appears below the pricing table, with the "up to 2 people" wording.
3. The eligibility section's occupancy table lists the SRO unit type.
4. Confirm no regression on `/listings/a0W7y00000ImkxREAR` (non-reserved SRO) or `/listings/a0W4U00000IgwwgUAB` (general-bucketed).

<details><summary><h3>Details</h3></summary>
<p>

## What was happening

The card is gated by `listingHasSROUnits(listing)`, which only inspected the `general` bucket:

```ts
listing.unitSummaries.general?.some((u) => u.unitType === "SRO" || u.unitType === "Room")
```

On a reserved-community listing `general` is `null`, so the optional chain short-circuits to `undefined` and the card never renders. Live API data:

| Listing | general | reserved | card |
| --- | --- | --- | --- |
| [a0W7y00000J1wlNEAR](https://housing.sfgov.org/listings/a0W7y00000J1wlNEAR?preview=true) (1303 Larkin, Senior 62+) | null | SRO, 20 units, maxOccupancy 2 | missing |
| [a0W7y00000ImkxREAR](https://housing.sfgov.org/listings/a0W7y00000ImkxREAR?preview=true) (750 Harrison, not reserved) | SRO, 9 units | empty | renders |
| [a0W4U00000IgwwgUAB](https://housing.sfgov.org/listings/a0W4U00000IgwwgUAB?preview=true) (Larkin Pine, older record) | SRO, 45 units | null | renders |

Note the third listing is the same building as the first and is also Senior 62+, but its summaries land in `general`. Its `Listing_Unit__c` junction records appear to be missing `Reserved_Type__c`, so it renders the card by accident. That is why this is fixed in the front end rather than by editing Salesforce data — correcting that record would make it lose the card too.

## Changes

- Add `getAllUnitSummaries` to `listingUtil.ts`, combining `general` and `reserved`.
- Use it in `listingHasSROUnits`, `listingHasOnlySROUnits`, and `isPluralSRO`, and extract a shared `isSROUnitType` predicate.
- Use it for `occupancyTableData` in `ListingDetailsEligibility.tsx`, which had the same `general`-only read, so the occupancy table and subtitle were also empty on reserved-only listings.
- Add a `reservedSroRentalListing` fixture modeled on a0W7y00000J1wlNEAR, plus unit tests for the helpers and a `ListingDetail` test asserting the card renders with the multiple-occupancy copy (its units are `maxOccupancy: 2`).

A union rather than the `general ?? reserved` fallback used in `for-rent.tsx` and `for-sale.tsx`, so listings with a mix of general and reserved units are handled correctly.

Two behavior notes for review:

- `listingHasOnlySROUnits` previously returned `undefined` when `general` was absent and `true` for an empty array. It now requires at least one summary, so a listing with no unit summaries returns `false` rather than a truthy-empty `true`.
- `occupancyTableData` was `UnitSummary[] | undefined` and is now always an array, so `StandardTable` gets `[]` instead of `undefined` on a listing with no summaries. Renders the same, snapshots unchanged.

</p>
</details> 

### Version Control

### Code quality

- [x] if the PR is a bugfix, there are tests and logs around the bug

### Code conventions

- [x] web pages are formatted with `.scss` stylesheets and `ui-seeds` tokens, rather than inline styles or Tailwind

### Review instructions

### Request eng review

### DIFF
--- a/app/javascript/util/listingUtil.ts
+++ b/app/javascript/util/listingUtil.ts
@@ -186,6 +186,41 @@ export const getAllUnitSummaries = (listing: RailsRentalListing | RailsSaleListi
 
 const isSROUnitType = (unitType: string) => unitType === "SRO" || unitType === "Room"
 
+const lowerBound = (a: number, b: number) => (a == null ? b : (b == null ? a : Math.min(a, b)))
+const upperBound = (a: number, b: number) => (a == null ? b : (b == null ? a : Math.max(a, b)))
+
+/**
+ * Collapse a listing's unit summaries into one occupancy range per unit type.
+ *
+ * A unit type can appear in both the general and reserved buckets (e.g. a building with one
+ * senior-reserved 1 BR alongside a general 1 BR), which would otherwise produce duplicate rows
+ * in the occupancy table. Occupancy bounds are merged so the range still covers every unit of
+ * that type.
+ *
+ * @param {RailsRentalListing | RailsSaleListing} listing
+ * @returns one entry per unit type, in the order the unit types are first encountered
+ */
+export const getOccupancyRangeByUnitType = (listing: RailsRentalListing | RailsSaleListing) => {
+  const rangesByUnitType = new Map<
+    string,
+    { unitType: string; minOccupancy: number; maxOccupancy: number }
+  >()
+
+  getAllUnitSummaries(listing).forEach(({ unitType, minOccupancy, maxOccupancy }) => {
+    const existingRange = rangesByUnitType.get(unitType)
+
+    if (!existingRange) {
+      rangesByUnitType.set(unitType, { unitType, minOccupancy, maxOccupancy })
+      return
+    }
+
+    existingRange.minOccupancy = lowerBound(existingRange.minOccupancy, minOccupancy)
+    existingRange.maxOccupancy = upperBound(existingRange.maxOccupancy, maxOccupancy)
+  })
+
+  return [...rangesByUnitType.values()]
+}
+
 /**
  * Check if a listing has only SRO units
  * @param {RailsRentalListing | RailsRentalListing} listing
--- a/app/javascript/util/listingUtil.ts
+++ b/app/javascript/util/listingUtil.ts
@@ -169,14 +169,33 @@ export const isCSLP = (listing: RailsRentalListing | RailsSaleListing) =>
   listing.Program_Type === "CSLP"
 
 /**
+ * Collect every unit summary on a listing, across both the general and reserved buckets.
+ *
+ * Salesforce splits unit summaries into `general` and `reserved`, and a listing can populate
+ * either or both. Reserved-community listings (senior, veteran, etc.) put all of their summaries
+ * in `reserved` and leave `general` null, so anything reading only `general` silently sees no
+ * units at all. Callers that care about unit types or occupancy should use this instead.
+ *
+ * @param {RailsRentalListing | RailsSaleListing} listing
+ * @returns the combined general and reserved unit summaries, empty when the listing has neither
+ */
+export const getAllUnitSummaries = (listing: RailsRentalListing | RailsSaleListing) => [
+  ...(listing.unitSummaries?.general ?? []),
+  ...(listing.unitSummaries?.reserved ?? []),
+]
+
+const isSROUnitType = (unitType: string) => unitType === "SRO" || unitType === "Room"
+
+/**
  * Check if a listing has only SRO units
  * @param {RailsRentalListing | RailsRentalListing} listing
  * @returns {boolean} returns true if the listing has all SRO unit types, false otherwise
  */
-export const listingHasOnlySROUnits = (listing: RailsRentalListing | RailsSaleListing) =>
-  listing.unitSummaries.general?.every(
-    (unit) => unit.unitType === "SRO" || unit.unitType === "Room"
-  )
+export const listingHasOnlySROUnits = (listing: RailsRentalListing | RailsSaleListing) => {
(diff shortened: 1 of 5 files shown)


fwextensions:
This old 62+ SRO listing still correctly shows the card in the PR build: https://dahlia-webap-fix-dah-43-5b4tf5.herokuapp.com/listings/a0W4U00000IgwwgUAB

image

fwextensions:
## Impact audit across all listings

To answer "which listings would actually look different," I scripted a replay of the display logic against live data: fetch every listing DAHLIA exposes, run both the pre-PR rules (general bucket only) and the post-PR rules, and diff the rendered outcomes (SRO card shown, which SRO copy, occupancy subtitle key, occupancy table rows).

Corpus: the 117 listings from `/api/v1/listings.json`, plus 2 more added explicitly because the index only returns currently-listed records and omits the waitlist/complete listings in the ticket. 119 total.

Bucket distribution:

| shape | count |
| --- | --- |
| general only | 117 |
| reserved only | 1 |
| both populated | 1 |
| neither | 0 |

### The audit caught a regression in my first commit

`Carlin Valencia Apartments` (a0W7y000004QCMbEAO) is the one listing with both buckets populated: a general 1 BR and a separate senior-reserved 1 BR, with identical occupancy bounds. Naively unioning the buckets gave the eligibility occupancy table a duplicate row:

```
[Studio/1-2, 1 BR/1-3]  ->  [Studio/1-2, 1 BR/1-3, 1 BR/1-3]
```

Fixed in 648c1774 by adding `getOccupancyRangeByUnitType`, which collapses the combined summaries to one entry per unit type and merges the occupancy bounds so the range still covers every unit of that type. The boolean SRO helpers keep using the raw union, where duplicates are harmless.

### Result after the fix

Exactly one listing changes, and it is the listing this ticket is about:

```
- a0W7y00000J1wlNEAR  Larkin Pine Senior Housing
    tenure=Re-rental status=New reserved=Senior general:0 reserved:1
    * SRO card: hidden -> shown
    * SRO card copy: n/a -> multiple occupancy
    * occupancy subtitle: occupancyDescriptionNoSro -> occupancyDescriptionAllSro(2)
    * occupancy table rows: [] -> [SRO/1-2]
```

Every other listing renders identically, which matches the unchanged `ListingDetail` and `ListingDetailsEligibility` snapshots.

### Caveats

- Production data only. A listing whose junction records change later could enter the reserved-only shape and start behaving differently, which is the intended behavior.
- The audit replays the display logic rather than rendering components, so it verifies the helper inputs and derived text keys, not the DOM.
- The script is a throwaway diagnostic and is not part of this PR. It lives at `tmp/sro-audit/audit-sro-buckets.mjs` (gitignored); happy to commit it under `scripts/` if it seems worth keeping.

fwextensions:
Apparently, the logic this PR adds also used to be in the Angular version of the page:

> The decisive finding is that this app already had my fix, in the AngularJS code the React pages replaced.
> 
> `ListingUnitService.js.coffee`:
> 
> ```coffee
> Service.combineUnitSummaries = (listing) ->
>   combined = _.concat(listing.unitSummaries.reserved, listing.unitSummaries.general)
>   combined = _.omitBy(_.uniqBy(combined, 'unitType'), _.isNil)
>   ...
> 
> Service.listingHasSROUnits = (listing) ->
>   combined = Service.combineUnitSummaries(listing)
>   _.some(combined, { Unit_Type: 'SRO' })
> 
> Service.listingHasOnlySROUnits = (listing) ->
>   combined = Service.combineUnitSummaries(listing)
>   return false unless combined.length
>   _.every(combined, { Unit_Type: 'SRO' })
> ```
> 
> Angular combined both buckets, deduped by unitType, and required a non-empty result. The React port kept the SRO checks but dropped the combine. My PR independently reconstructs all three behaviors, including the dedupe I only added after the audit flagged Carlin, and the length > 0 guard I flagged as a behavior change. That is strong evidence this is a port regression rather than a Salesforce data problem.

app/javascript/util/listingUtil.ts
```diff
@@ -168,23 +168,88 @@ export const isBMR = (listing: RailsRentalListing | RailsSaleListing) =>
...
+ * @returns the combined general and reserved unit summaries, empty when the listing has neither
+ */
+export const getAllUnitSummaries = (listing: RailsRentalListing | RailsSaleListing) => [
+  ...(listing.unitSummaries?.general ?? []),
+  ...(listing.unitSummaries?.reserved ?? []),
+]
+
+const isSROUnitType = (unitType: string) => unitType === "SRO" || unitType === "Room"
+
+const definedBounds = (a: number, b: number) => [a, b].filter((bound) => bound != null)
```
cliu02, line 189: non-blocking suggestion: simplify the helpers `definedBounds`, `lowerBound`,  `upperBound` to be more understandable.

cliu02 (approved):
🔥

PR #3055, opened 2026-09-03
feat: DAH-4337 Account Setting: Change Password Flow

jtcaovan:
## Description

Implements new Change your password flow from account settings page - includes flow of Adding a Password if the user does not yet have a password

The most important changes are in `settings.tsx` `change-password.tsx` `add-password.tsx`. the rest of the changes are tests and routing

new strings need to be pushed

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4337

### Version Control

### Code quality

### Code conventions

- [x] web pages are formatted with `.scss` stylesheets and `ui-seeds` tokens, rather than inline styles or Tailwind

### Review instructions

### Request eng review


cypress/e2e/accounts/accountSettings.e2e.ts
```diff
@@ -86,21 +86,21 @@ describe("Account Settings", () => {
     )
     cy.wait("@emailChange").its("response.statusCode").should("eq", 200)
 
-    // Submit password change
```
jtcaovan, line 89: note: most of these e2e tests will need to be updated as we continue to build out more of the new accounts page

app/javascript/pages/account/change-password.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return
+    }
+    if (!isLoaded) return
+    if (!isSignedIn) {
+      void navigate(getSignInPath())
+      return
+    }
+  }, [flagsReady, clerkEnabled, isLoaded, isSignedIn, navigate])
+
+  const ready = flagsReady && clerkEnabled && isLoaded && isSignedIn
```
kramduckner, line 120: Blocker: I think we'll need a guard somewhere to ensure that you can't get here is you don't have a password already

app/javascript/pages/account/settings.tsx
```diff
@@ -586,18 +574,20 @@ const AccountSettings = ({ profile }: { profile: User }) => {
...
         className="mt-8"
         message={t("accountSettings.accountChangesSaved")}
+        fullWidth
         onClose={() => setNameSavedBanner(false)}
       />
       <NameSection user={user} setUser={setUser} handleBanners={handleBanners} />
       <DateOfBirthSection user={user} setUser={setUser} />
       <EmailSection user={user} setUser={setUser} />
-      <PasswordSection user={user} setUser={setUser} />
+      {clerkEnabled && <PasswordSection />}
```
kramduckner: Non Blocking: if i understand correctly, deleting the Devise `PasswordSection` leaves `newAccountLayout=on` +`auth.clerk=off` with no password UI at all, but that might not be a problem.

app/javascript/pages/account/add-password.tsx
```diff
@@ -97,7 +99,11 @@ const AddPasswordPage = ({ flow }: AddPasswordPageProps) => {
     try {
       if (!user) return
       await user.updatePassword({ newPassword })
-      void navigate(getAddProfilePath())
+      if (isAccountSettingsFlow) {
```
kramduckner, line 102: Non Blocking: This might not be related to this pr - but looking at the re-verification in change-password got me wondering if we need to do that here? What happens if a user isn't 'verified' (however long it's set in clerk) by doing something like: signing in then exploring some rentals for a while then deciding that they want to change password? If I understand it, they get a generic error message when they try to change password.

kramduckner (changes requested):
Looks/works great! Just caught one thing

app/javascript/pages/account/settings.tsx
jtcaovan: that's a good call, devise uses a different accounts component `my-account.tsx` so adding clerkEnabled doesn't actually do anything I believe.

app/javascript/pages/account/add-password.tsx
jtcaovan: great catch! [i created a separate ticket ](https://sfgovdt.jira.com/browse/DAH-4372) because updating the login email may require this re-verification too

kramduckner (approved):

PR #3060, opened 2026-09-08
feat: DAH-4344 Upgrade clerk react sdk

jimlin-sfgov:
## Description

Run `yarn` to update the Clerk React package.

Upgrade Clerk React SDK. Auth behavior with Clerk enabled should look the same for users.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4344

### Version Control

### Code quality

  - not small, but the scope is small-ish 
  - not adding more code coverage, since there will be a lot of refactoring for the abstraction layer ticket 

### Code conventions

### Review instructions

### Request eng review


app/javascript/authentication/ForgotPasswordFlow.tsx
```diff
@@ -18,31 +18,32 @@ const ForgotPasswordFlow = () => {
     handleSubmit,
     formState: { errors },
   } = useForm()
-  const { isLoaded, signIn } = useSignIn()
+  const { signIn, fetchStatus: signInFetchStatus } = useSignIn()
+
   const navigate = useNavigate()
   const prefilledEmailParam = new URLSearchParams(window.location.search).get("email") ?? ""
 
+  // TODO: show error messages
```
jtcaovan: question (non-blocking): which error messages do we need to show? are we showing these errors in the UI (banner, field level, etc)?

app/javascript/pages/account/verification-code.tsx
```diff
@@ -345,7 +357,10 @@ const EnterVerificationCode = (_props: { assetPaths: unknown }) => {
    * If the user is signed in with a profile, redirect to my account.
    * If the user is signed in without a profile, redirect to the add profile page.
    */
+  const hasRun = useRef(false) // only redirect when first visiting this page, otherwise it overrides navigate() calls from code submission
```
jtcaovan: nit (non-blocking):  the addition of `hasRun` was confusing at first glance but makes sense now after thinking through the redirect override. It could be helpful if we added a more descriptive variable name here. Off the top of my head that may or may not be better: `hasPageRun` `hasCheckedPageEntry` `formShown` `userEnteredPage` etc

jtcaovan (approved):
LGTM! Awesome job with this upgrade! a couple of non-blocking comments but otherwise no issues found

app/javascript/authentication/ForgotPasswordFlow.tsx
jimlin-sfgov: Sorry for the brevity of the comment. I'm actually at a loss for what needs to be done. Originally, we would always navigate to the forgot password page, [even if there was an error](https://github.com/SFDigitalServices/sf-dahlia-web/blob/8ea2797cfa93d237152f78070d6841cd687f49d6/app/javascript/authentication/ForgotPasswordFlow.tsx#L42). That doesn't seem right.

But looking at the associated [ticket](https://sfgovdt.jira.com/browse/DAH-4199), I don't see any error messages in the designs either.

app/javascript/pages/account/verification-code.tsx
jimlin-sfgov: Thanks for the feedback. I can update the name to be more meaningful. This is intended to be a temporary workaround, this coding pattern of using `useEffect` to redirect is really confusing and should be removed soon imo.

app/javascript/authentication/ForgotPasswordFlow.tsx
jtcaovan: No worries! I had the same question too when working on the ticket. Yindi and team decided the user will always pass through to the verification code page, regardless of whether the user's email exists in Clerk or not. This may prevent bad actors from going through and inputting various emails to confirm if they are active users. For the user, we added the text 'If the account exists, you should get a code...'

On the forgot password screen, I believe the only error the user will get is if they enter an invalid email format pattern checked by regex

app/javascript/pages/account/verification-code.tsx
jtcaovan: Sure! If it's temporary, no need to update the name

app/javascript/authentication/ForgotPasswordFlow.tsx
jimlin-sfgov: Enumeration protection would only involve "hiding" the email-does-not-exist type of error. There may be other errors, like if our Clerk service was down.

I'll add the enumeration protection scenario to my other [ticket](https://sfgovdt.jira.com/browse/DAH-4244)

app/javascript/authentication/SignInFlow.tsx
```diff
@@ -97,34 +99,36 @@ const SignInFlow = () => {
...
+      console.error("Sign in not complete:", signIn.status)
+      setShowError(true)
+      return
+    }
+    clearHeaders() // Clear headers in case of existing Devise session (while testing)
+
+    const housingCounselorToken = getHousingCounselorToken()
+    if (housingCounselorToken) {
+      // housingCounselorChecked.current = true // not needed because we assign it in useEffect, it also violates linter rules
+      const housingCounselorAccess = await checkHousingCounselorAccess()
```
kramduckner, line 128: Blocking: If I understand signIn.finalize, we won't have a session token for housingCounselorAccess to use until finalize() runs below.

app/javascript/pages/account/add-password.tsx
```diff
@@ -48,27 +48,53 @@ const AddPasswordPage = ({ flow }: AddPasswordPageProps) => {
     shouldFocusError: false,
   })
 
-  if (isForgotPasswordFlow && signInLoaded && !isResettingPassword && !signIn?.status) {
+  if (
+    isForgotPasswordFlow &&
+    signInFetchStatus !== "fetching" &&
```
kramduckner: Quick question on fetchStatus - is it filling in for the old signInLoaded here?

The type is 'idle' | 'fetching', which looks like it means "request in flight" rather than "Clerk has booted". If that's right, wouldn't it be "idle" on first render, with signIn still null, firing the redirect before Clerk restores the reset attempt? Worried that bounces people back to forgot-password right after they enter a valid code.

 Could easily be misreading how it initializes though@

Same swap is in a few other spots. Mostly disabled-button props, but SignInFlow.tsx:55 gates the "default to email code" effect, so maybe worth a look there too.

app/javascript/pages/account/verification-code.tsx
```diff
@@ -161,58 +158,72 @@ const EnterVerificationCodePage = ({
...
-      await signIn.prepareFirstFactor({
-        strategy: "email_code",
-        emailAddressId: emailCodeFactor.emailAddressId,
-      })
-      return true
-    } catch (error) {
-      console.error("Sign in code resend error", error)
+    if (signInFetchStatus === "fetching" || !signIn) return false
+
+    if (!signIn.emailAddress) {
```
kramduckner: Blocking: I think signIn.emailAddress might not exist on the resource - it's only on the create params, not SignInFutureResource. If so it'd always be undefined, so this guard always returns false and the resend never actually fires. Affects sign-in and forgot-password; sign-up is fine since signUp.emailAddress is real.

app/javascript/pages/account/verification-code.tsx
```diff
@@ -82,76 +85,70 @@ const EnterVerificationCodePage = ({
...
+  }
+
+  const verifySignUpCode = async (code: string) => {
+    if (signUpFetchStatus === "fetching" || !signUp) return
+
+    await signUp.verifications.verifyEmailCode({ code })
+    if (signUp.status !== "complete") {
+      console.error("Code verification not complete:", signUp.status)
       setError("code", { message: "invalid" })
     }
```
kramduckner, line 135: Is this missing a return? As written it sets the "invalid" error and then still falls through to finalize() on line 130, so a bad code would show the error and navigate to add-password anyway, on an unverified signup.

Also noticed verifyEmailCode on line 124 returns an { error } that isn't checked - the other two verify functions in this file destructure it. Could be intentional though!

kramduckner (changes requested):
Looks good! Upgrades are slogs thanks for doing it. I had a couple questions and a couple blockers if I understand things correctly

app/javascript/pages/account/verification-code.tsx
jimlin-sfgov: Good catch. I could check for `signIn.identifier` instead, but I'm fairly sure that the Clerk SDK will check for that when `signIn.resetPasswordEmailCode.sendCode()` is called, and give us the appropriate error.

I will remove this conditional block, and the two other similar blocks.

app/javascript/pages/account/verification-code.tsx
jimlin-sfgov: Your suspicions are correct, I didn't implement the standard pattern here, which is to check for errors and status, and return early accordingly.

app/javascript/authentication/SignInFlow.tsx
jimlin-sfgov: That's correct, and now I have to figure out how to do the housingCounselor stuff after `finalize` 😞

kramduckner: Looks good! While checking, I did catch that this might also be happening in verification-code.tsx:102-114. verifySignInCode still calls getToken() and authorizeHousingCounselor before signIn.finalize()

kramduckner (approved):
Approved! I think I caught another finalize() thing and noted it in a comment. The updates look great, I'm gonna approve

jimlin-sfgov:
Still awaiting PA testing, but at this point I'm gonna merge to unblock folks. We can follow up with fixes if PA testing reveals more issues.

PR #3068, opened 2026-09-11
refactor: DAH-4345 route Clerk components through the auth session layer

kramduckner:
## Description

Routes the Clerk sign-in, sign-out, and session-status reads in Layout, AccountNav, and the account pages through the useAuthSession hook instead of calling @clerk/react's useAuth and clearHeaders directly. Several account and authentication entry points still read session status and tokens straight from Clerk, and the sign-in, sign-up, and password flows aren't modeled by the layer yet - that'll be done in the next pr.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-4345

### Version Control

### Code quality

### Code conventions

### Review instructions

### Request eng review


app/javascript/authentication/session/adapters/clerk/useClerkAuthSession.ts
```diff
@@ -21,11 +22,20 @@ export const useClerkAuthSession = (): AuthSession => {
     }
   }, [getToken])
 
+  const signOut = useCallback(async (): Promise<void> => {
+    // TODO(DAH-4366): CLERK MIGRATION - DEVISE TECH DEBT TO REMOVE
+    // apiService attaches stored Devise headers to every request, so a user who
+    // last signed in with Devise would keep sending them.
+    clearHeaders()
```
jimlin-sfgov, line 29: thought: "adapters" shouldn't be aware of other auth strategies. It might be better to do this kind of clean up in a higher layer.

jimlin-sfgov (approved):
lgtm

app/javascript/authentication/session/adapters/clerk/useClerkAuthSession.ts
kramduckner: I think you're right, that's a good thing to catch. If this was going to be in there long term I would rewrite it, but I'm thinking it's ok since we'll remove the devise coupling in the tech debt ticket? I intend this will only be one auth strategy after this devise + clerk double support state. LMK if you still think we should move it though and I'll do it in a follow up pr

jimlin-sfgov: > we'll remove the devise coupling in the tech debt ticket?

The question is when that will happen? I can easily see the timeline getting significantly delayed, as they tend to do around here.

